Issue #19, opened 2020-02-27
Use SDK for building Go binaries


jahkeup:
**Issue or Feature Request:**

The build process for the update operator should use the same Go toolchain used throughout Bottlerocket. 

This can be accomplished by subbing in the SDK container where the `golang:N.N` image was being used before. The license scanner (#6) should also use this image when collecting the modules into the to-be-scanned `vendor/` directory.

**Suggested Solution**

The "golang" image can be provided as an `ARG` to the build commands (as is proposed in #6 for other uses - see `Dockerfile.licenses`). I was able to make a set of changes to do this in short order (and still support `docker build -t neio:latest .` at the root). Here's a [gist with the set of changes](https://gist.github.com/jahkeup/f8024e941d36815cec551dbb7e9cd15c).

PR #21, opened 2020-03-02
Use SDK's Go toolchain

jahkeup:
**Issue number:**

#19 

**Description of changes:**

This uses the [`bottlerocket/sdk-x86_64`](https://github.com/bottlerocket-os/bottlerocket-sdk) container image to build the update operator's binaries.

**Testing done:**

Built a container image and ran unit tests.

**Terms of contribution:**


Makefile
```diff
@@ -85,11 +86,12 @@ container: licenses
...
 	docker build \
 		--network=host \
-		--build-arg GO_LDFLAGS='$(GO_LDFLAGS)' \
-		--build-arg GOARCH='$(GOARCH)' \
-		--build-arg SHORT_SHA='$(SHORT_SHA)' \
-		--build-arg NOCACHE='$(shell date +"%s")' \
-		--build-arg LICENSES_IMAGE=$(LICENSES_IMAGE) \
+		--build-arg GO_LDFLAGS \
+		--build-arg GOARCH \
+		--build-arg "SHORT_SHA=$(SHORT_SHA)" \
```
etungsten: nit: Can we do the same (setting `build-arg` without a value and just use local environment variable) for `SHORT_SHA` ( and `LICENSES_IMAGE` below)?
```suggestion
		--build-arg GOARCH \
		--build-arg SHORT_SHA \
```

jahkeup: We can't as is, the variable [`GOARCH` is exported](https://github.com/bottlerocket-os/bottlerocket-update-operator/pull/21/files/e4d254471a6ec2d920a471bb29aa75025154a6ee#diff-b67911656ef5d18c4ae36cb6741b7965R52) whereas [`SHORT_SHA` is not](https://github.com/bottlerocket-os/bottlerocket-update-operator/pull/21/files/e4d254471a6ec2d920a471bb29aa75025154a6ee#diff-b67911656ef5d18c4ae36cb6741b7965R20) - to use the nicer `--buildarg MY_ENV_VAR` syntax the variable must be present. There's only a couple variables that are `export`ed that are related to build-wide influencing variables (like `GOARCH`). I would prefer to not have `SHORT_SHA` also exported too because in my mind it doesn't quite fit in with that sort of influential build/compile environment variable.

Do you think it would be less tempting if we instead made all the `--build-arg` flags use the fuller syntax? (ie: `--build-arg MY_ENV_VAR=my-value`)

etungsten: I would prefer the syntax to be consistent. But it is not a hard blocker for me.

jahkeup: I agree, I'll update them to be consistent so we aren't tempted to get them any more out of whack with each other :+1:

etungsten (approved):

webern (approved):

jahkeup:
I want to bump this to use the latest SDK image before we merge this in (it got bumped to [`0.10.1`](https://github.com/bottlerocket-os/bottlerocket/blob/a07d60c6e540e96392a73a311394b90cb391f39c/Makefile.toml#L31) in the OS repo).

jahkeup:
Bumped to `v0.10.1` and rebased. It builds :popcorn:

zmrow (approved):
☺️

PR #30, opened 2020-03-19
change `platform-version` to `updater-interface-version`

### LINKED ISSUE #4: Change `platform-version` key name
**Issue or Feature Request:**

`platform-version` used in the Label `bottlerocket.aws/platform-version` for picking compatible hosts to schedule an Operator's Agent Pod stands to be better named. The current name does not convey the Label's meaning and is overloaded in use (ex: used by [EKS](https://docs.aws.amazon.com/eks/latest/userguide/platform-versions.html) and [Fargate](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/platform_versions.html)). The name `platform-version` is replaced with should make it clear to Cluster operators that it's a host integration compatibility version.

Some suggested keys:

| Name                              | Full Key                                           |
|:----------------------------------|:---------------------------------------------------|
| `updater-interface-version`       | `bottlerocket.aws/updater-interface-version`       |
| `host-update-interface-version`   | `bottlerocket.aws/host-update-interface-version`   |
| `host-update-integration-version` | `bottlerocket.aws/host-update-integration-version` |

**My preference:** `updater-interface-version`

With `updater-interface-version`: `bottlerocket.aws/updater-interface-version=1.0.0`. 

This naming would also nicely relate other keys, if needed. For example, if this were the selector for a chosen backing "updater" implementation: `bottlerocket.aws/updater-implementation=noop`. I don't think we'll need anything of the sort soon, but it's a nice thought and reads as a very straightforward relationship to me.

etungsten:
**Issue number:**
Addresses https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/4

**Description of changes:**
This commit renames references of `platform-version` to
`platform-interface-version` to better convey the label's meaning and
avoid conflicts with other keywords.

**Testing done:**
Built my own Bottlerocket update operator container with the changes, modified `update-operator.yaml` to pull that container for the agent and controller.
Labelled my nodes in my cluster with the updated key name (`bottlerocket.aws/updater-interface-version`) and I see the pods being scheduled to run immediately after.
```
$ kubectl label node $(kubectl get nodes -o jsonpath='{.items[*].metadata.name}') bottlerocket.aws/updater-interface-version=1.0.0
node/ip-192-168-17-182.us-west-2.compute.internal labeled
node/ip-192-168-25-184.us-west-2.compute.internal labeled
node/ip-192-168-40-60.us-west-2.compute.internal labeled
node/ip-192-168-60-95.us-west-2.compute.internal labeled
node/ip-192-168-70-84.us-west-2.compute.internal labeled
node/ip-192-168-9-191.us-west-2.compute.internal labeled 
$ kubectl get pods -A
NAMESPACE      NAME                                         READY   STATUS    RESTARTS   AGE
bottlerocket   update-operator-agent-5zm4x                  1/1     Running   0          14s
bottlerocket   update-operator-agent-6wlwq                  1/1     Running   0          14s
bottlerocket   update-operator-agent-85xfj                  1/1     Running   0          14s
bottlerocket   update-operator-agent-j99mq                  1/1     Running   0          14s
bottlerocket   update-operator-agent-r27ng                  1/1     Running   0          14s
bottlerocket   update-operator-agent-tjwjs                  1/1     Running   0          14s
bottlerocket   update-operator-controller-f749847d7-2mvhs   1/1     Running   0          2m52s
....
```

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,27 +48,29 @@ kubectl apply -f ./update-operator.yaml
 Once the deployment's resources are in place, there is one more step needed to schedule and place the required pods on Bottlerocket nodes.
 By default - in the suggested deployment, each Workload resource constrains scheduling of the update operator by limiting pods to Bottlerocket nodes based on their labels.
 These labels are not applied on nodes automatically and will need to be set on each using `kubectl`.
-Deployments match on `bottlerocket.aws/platform-version` (a host compatibility label) to determine which pod to schedule.
+The agent relies on each node's updater components and schedules its pods based on their interface supported.
+The node indicates its updater interface version in a label called `bottlerocket.aws/updater-interface-version`.
+Agent deployments, respective to the interface version, are scheduled using this label and target only a single version in each.
 
-For the `1.0.0` `platform-version`, this label looks like:
+For the `1.0.0` `updater-interface-version`, this label looks like:
 
 ``` text
-bottlerocket.aws/platform-version=1.0.0
+bottlerocket.aws/updater-interface-version=1.0.0
 ```
 
 `kubectl` can be used to set this label on a node in the cluster:
 
 ``` sh
-kubectl label node $NODE_NAME bottlerocket.aws/platform-version=1.0.0
+kubectl label node $NODE_NAME bottlerocket.aws/updater-interface-version=1.0.0
 ```
 
-If all nodes in the cluster are running Bottlerocket and have the same `platform-version`, all can be labeled at the same time:
+If all nodes in the cluster are running Bottlerocket and have the same `updater-interface-version`, all can be labeled at the same time:
 
 ``` sh
-kubectl label node $(kubectl get nodes -o jsonpath='{.items[*].metadata.name}') bottlerocket.aws/platform-version=1.0.0
+kubectl label node $(kubectl get nodes -o jsonpath='{.items[*].metadata.name}') bottlerocket.aws/updater-interface-version=1.0.0
 ```
 
-Each workload resource may have additional constraints or scheduling affinities based on each node's labels in addition to the `bottlerocket.aws/platform-version` label scheduling constraint.
+Each workload resource may have additional constraints or scheduling affinities based on each node's labels in addition to the `bottlerocket.aws/updater-interface-version` label scheduling constraint.
 
 Customized deployments may use the [suggested deployment](./update-operator.yaml) or the [example development deployment](./dev/deployment.yaml) as a starting point, with customized container images specified if needed.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -48,27 +48,27 @@ kubectl apply -f ./update-operator.yaml
 Once the deployment's resources are in place, there is one more step needed to schedule and place the required pods on Bottlerocket nodes.
 By default - in the suggested deployment, each Workload resource constrains scheduling of the update operator by limiting pods to Bottlerocket nodes based on their labels.
 These labels are not applied on nodes automatically and will need to be set on each using `kubectl`.
-Deployments match on `bottlerocket.aws/platform-version` (a host compatibility label) to determine which pod to schedule.
+Deployments match on `bottlerocket.aws/updater-interface-version` (a host compatibility label) to determine which pod to schedule.
```
jahkeup: Let's update the wording a bit to briefly explain why the new name is what it is :+1: 

Maybe something like this?

```suggestion
The agent relies on each node's updater components and schedules its pods based on their interface supported.
The node indicates its updater interface version in a label called `bottlerocket.aws/updater-interface-version`.
Agent deployments, respective to the interface version, are scheduled using this label and target only a single version in each.
```

dev/deployment.yaml
```diff
@@ -94,7 +94,7 @@ spec:
           requiredDuringSchedulingIgnoredDuringExecution:
             nodeSelectorTerms:
               - matchExpressions:
-                  - key: bottlerocket.aws/platform-version
+                  - key: bottlerocket.aws/updater-interface-version
                     operator: Exists
```
jahkeup, line 98: Quick thought (and not for this particular PR to resolve):

I think we may want to update this deployment prior to making any changes to the updater components host side. The current deployments will land on hosts that end up having a bumped version without this being scoped down to `updater-interface-version==1.0.0`. That way we can begin to vend a newer version without impacting existing deployments (which are scheduled if the key exists at all). 

I had originally aimed to have the host set its interface version by providing a compiled in default setting. If we do this then we reduce the chance that the wrong version be set for a host *and* eliminate the need for cluster operators to label their nodes.

```toml
[settings.kubernetes.node-labels]
"bottlerocket.aws/updater-interface-version" = "1.0.0"
```

jahkeup (approved):
Looks good!

Unless others object, I'm in favor of moving on with the `updater-interface-version` name.

etungsten:
Addresses @jahkeup 's comments - updated README with suggestion.

jahkeup (approved):

jahkeup:
We should make sure to include instructions for the next update to change the labels to the new name.

samuelkarp (approved):
Once we release this, we'll need to call out that this is a breaking change in the release notes.  I think this is a good change, but anyone who is already using the update operator is going to have to make manual changes to the annotations in their Kubernetes cluster.

Issue #24, opened 2020-03-03
Log no-update-available result from updater interface as non-error


jahkeup:
**Image I'm using:**

`328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-update-operator:v0.1.2`

**Issue or Feature Request:**

When the update operator agent checks for an update, the logged messages makes the result look like an error - though its an expected behavior/response from the call to the updater.

Instead of making this logged as an error, this should instead be logged as a "no update available" result based on the returned status code and the updater's message.

jahkeup:
This isn't the default updater with `1.0` being released. I think we can punt for now.

PR #32, opened 2020-04-06
platform-updog: treat no update available as a non error

etungsten:
**Issue number:**
Addresses https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/24

**Description of changes:**
Determines if `updog` has returned `no update available` when querying
for updates and have the agent treat it as a non-error even if `updog`
returns a non-zero exit code.

**Testing done:**
Launched latest Bottlerocket version instance, have it join my cluster, label it for the update operator.
<details>
<summary>Observed the following</summary>

```
DEBU[2020-04-06T21:54:53Z] starting                                      component=agent worker=informer
DEBU[2020-04-06T21:54:53Z] resource add event                            component=agent worker=informer
DEBU[2020-04-06T21:54:53Z] active intent received                        component=agent intent="stabilize,unknown,unknown update:unknown" node=ip-192-168-7-212.us-west-2.compute.internal
DEBU[2020-04-06T21:54:53Z] handling intent                               component=agent intent="stabilize,unknown,unknown update:unknown" worker=handler
DEBU[2020-04-06T21:54:53Z] posted intent                                 component=agent intent="stabilize,stabilize,busy update:unknown" node=ip-192-168-7-212.us-west-2.compute.internal
DEBU[2020-04-06T21:54:53Z] sitrep                                        component=agent intent="stabilize,unknown,unknown update:unknown" worker=handler
DEBU[2020-04-06T21:54:53Z] querying status                               component=platform
DEBU[2020-04-06T21:54:53Z] fetching list of available updates            component=platform
DEBU[2020-04-06T21:54:53Z] running command                               cmd="/usr/bin/updog check-update" component=updog subcomponent=host-bin
INFO[2020-04-06T21:54:54Z] no update available                           component=agent intent="stabilize,unknown,unknown update:unknown" update-available=false worker=handler 
...
``` 

</details>

<details>
<summary>Whereas before ... (without these changes):</summary>

```
DEBU[2020-04-06T22:09:11Z] running command                               cmd="/usr/bin/updog check-update" component=updog subcomponent=host-bin
ERRO[2020-04-06T22:09:11Z] error during command run                      cmd="/usr/bin/updog check-update" component=updog error="exit status 1" subcomponent=host-bin
ERRO[2020-04-06T22:09:11Z] unable to query available updates             component=agent error="unable to check for updates: exit status 1" intent="stabilize,stabilize,busy update:unknown" worker=handler
ERRO[2020-04-06T22:09:11Z] sitrep update check errored                   component=agent error="unable to check for updates: exit status 1" intent="stabilize,stabilize,busy update:unknown" worker=handler 
```

</details>

The updater interface no longer treats no available update as an error.

**Terms of contribution:**

### DIFF
--- a/pkg/platform/updog/updog.go
+++ b/pkg/platform/updog/updog.go
@@ -41,6 +41,8 @@ type executable struct {
 	log logging.SubLogger
 }
 
+// Boolean currently only used by CheckUpdate (and consequently ListUpdate). Returns true if the
+// command yielded output, which indicates an update is available.
 func (e *executable) runOk(cmd *exec.Cmd) (bool, error) {
 	cmd.SysProcAttr = bottlerocket.ProcessAttrs()
 
@@ -60,6 +62,11 @@ func (e *executable) runOk(cmd *exec.Cmd) (bool, error) {
 	}
 	err := cmd.Wait()
 	if err != nil {
+		// Updog returns an non zero exit code if there are no available updates.
+		// Don't treat it as an error, instead indicate that there's no available update.
+		if len(cmd.Args) > 1 && cmd.Args[1] == "check-update" && buf.String() == "No update available\n" {
+			return false, nil
+		}
 		log.WithError(err).Error("error during command run")
 		if logging.Debuggable {
 			log.WithField("output", buf.String()).Debug("command output")
@@ -70,9 +77,6 @@ func (e *executable) runOk(cmd *exec.Cmd) (bool, error) {
 	if logging.Debuggable {
 		log.WithField("output", buf.String()).Debug("command output")
 	}
-	// Boolean currently only used by ListUpdate. Returns true if the
-	// command yielded output, which indicates an update is available.
-	//
 	// TODO: Use richer interface data once provided by host integration.
 	updateEmitted := len(buf.String()) > 0
 	return updateEmitted, err
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -159,6 +159,8 @@ func (a *Agent) checkUpdate(log logging.Logger) (bool, error) {
 	log = log.WithField("update-available", hasUpdate)
 	if hasUpdate {
 		log.Info("an update is available")
+	} else {
+		log.Info("no update available")
 	}
 	return hasUpdate, nil
 }


pkg/platform/updog/updog.go
```diff
@@ -60,6 +62,11 @@ func (e *executable) runOk(cmd *exec.Cmd) (bool, error) {
 	}
 	err := cmd.Wait()
 	if err != nil {
+		// Updog returns an non zero exit code if there are no available updates.
+		// Don't treat it as an error, instead indicate that there's no available update.
+		if len(cmd.Args) > 1 && cmd.Args[1] == "check-update" && buf.String() == "No update available\n" {
```
tjkirch, line 67: How does it match with the difference in case?

tjkirch (commented):
Is there something we can do to make this not a permanent solution?  Can we fix shutdown ordering / messaging so that we're not working around updog's exit code being different because of shutdown signals, or inform brupop earlier that an update is being performed so it doesn't have to guess status based on output / return code?

pkg/platform/updog/updog.go
jahkeup, line 67: I think that we might want to handle this with `$1 == "check-update" && exitCode == 1`, or something similar instead of reading into the (human oriented) error message.

`updog`'s ["no update available"](https://github.com/bottlerocket-os/bottlerocket/blob/a07d60c6e540e96392a73a311394b90cb391f39c/sources/updater/updog/src/main.rs#L501) and ["no-update in node's wave"](https://github.com/bottlerocket-os/bottlerocket/blob/a07d60c6e540e96392a73a311394b90cb391f39c/sources/updater/updog/src/main.rs#L506) "errors" are only distinguished by the [printed error string](https://github.com/bottlerocket-os/bottlerocket/blob/a07d60c6e540e96392a73a311394b90cb391f39c/sources/updater/updog/src/error.rs#L206-L213). The update operator, for what its concerned with here anyway, doesn't make a distinction between these two - and I don't think it'll need to (not before the interface is formalized anyway). So, if the agent ran `updog check-update` and it gets a `exitCode == 1` back, then we know that `updog` effectively is reporting "no updates [yet]!".

pkg/platform/updog/updog.go
```diff
@@ -41,6 +41,8 @@ type executable struct {
 	log logging.SubLogger
 }
 
+// Boolean currently only used by CheckUpdate (and consequently ListUpdate). Returns true if the
```
jahkeup, line 44: nit: conventionally, this is worded with `runOk` being the subject (at the start of the wording)

```go
// runOk executes the provided command, checking for printed output and error
// from the process. This command is used by CheckUpdate (and ListUpdate by
// extension) where printed output indicates that there is an update available.
func (e *executable) runOk(...) ... {
```

jahkeup:
> Is there something we can do to make this not a permanent solution? 

We'll want to have a machine oriented interface here - we might be able to take advantage of the produced JSON (which wasn't there when this was originally written). The issue is that `updog` doesn't provide data about the check result other than the printed error message, at least the way we're calling and checking right now.

> Can we fix shutdown ordering / messaging so that we're not working around updog's exit code being different because of shutdown signals, or inform brupop earlier that an update is being performed so it doesn't have to guess status based on output / return code?

This isn't related to the shutdown ordering problem, this is removing "error" log messages from the printed messages when the agent checks for an update.

pkg/platform/updog/updog.go
tjkirch: I don't follow how we know that.  How do we know that updog didn't fail to check, in which case brupop should fail as well?

jahkeup: I think with the context used as-is that we'll see the same error printed: [`errors::UpdateNotAvailable`](https://github.com/bottlerocket-os/bottlerocket/blob/a07d60c6e540e96392a73a311394b90cb391f39c/sources/updater/updog/src/main.rs#L501). There could be inner error handling I missed or misread that makes this not true! Though in that case, I wonder what we'd see if errors *don't* bubble up. Does it show up by have the error result massaged into `None` (in the check `fn`) from the downstream errors?

jahkeup: I'm glossing over the *earlier* errors. My bad.

jahkeup: I would really like to see different error codes, but I know this isn't the API we want to formalize so its unlikely we'll get any additional distinguishers there. Without some method to distinguish between operational error and check-result "errors", we're having to resort to checking strings..and that's no bueno.

etungsten:
I'm holding off on this PR for now in favour of waiting for a better updater interface so string matching on `updog`'s error output wouldn't be necessary.

PR #39, opened 2020-08-20
platform api: list update when non-nil

jahkeup:
**Issue number:**

n/a

**Description of changes:**

When using the `update-api` update platform, the node lists itself as having updates available when it does not. This is because the response includes a `nil` update in its list. The update itself isn't inspected and instead the response is checked for `len(availableUpdatesResponse.Updates()) > 0`. The `nil` element makes this evaluate to `true` and then the node will cycle through update attempts.

**Testing done:**

Nodes in a cluster running Bottlerocket `0.5.0` upgrades successfully to an artificial `0.5.1` then stabilize.

**Terms of contribution:**

### DIFF
--- a/pkg/platform/api/platform.go
+++ b/pkg/platform/api/platform.go
@@ -53,6 +53,10 @@ type listAvailableResponse struct {
 }
 
 func (lar *listAvailableResponse) Updates() []platform.Update {
+	if lar.chosenUpdate == nil {
+		return nil
+	}
+
 	return []platform.Update{lar.chosenUpdate}
 }
 


etungsten (approved):

samuelkarp:
> When using the update-api update platform, the node lists itself as having updates available when it does not. This is because the response includes a `nil` update in its list.

Do we have an issue to track this bug in the update API?

samuelkarp (approved):

pkg/platform/api/platform.go
```diff
@@ -53,7 +53,11 @@ type listAvailableResponse struct {
 }
 
 func (lar *listAvailableResponse) Updates() []platform.Update {
-	return []platform.Update{lar.chosenUpdate}
+	if lar.chosenUpdate != nil {
+		return []platform.Update{lar.chosenUpdate}
+	}
+
+	return []platform.Update{}
```
samuelkarp:
```suggestion
	if lar.chosenUpdate == nil {
		return nil
	}

	return []platform.Update{lar.chosenUpdate}
```

This would also work (unless there's some reason not to return `nil` from this function.

But please do not block on this suggestion!

etungsten:
> > When using the update-api update platform, the node lists itself as having updates available when it does not. This is because the response includes a `nil` update in its list.
> 
> Do we have an issue to track this bug in the update API?

This is an issue specific to the update-api platform in brupop.
The update API return structure looks like:
```
{
   "update_state":"Idle",
   "available_updates":[
      "0.5.0"
   ],
   "chosen_update":null,
   "active_partition":{
      "image":{
         "arch":"aarch64",
         "version":"0.5.0",
         "variant":"aws-k8s-1.17"
      },
      "next_to_boot":true
   },
   "staging_partition":null,
   "most_recent_command":{
      "cmd_type":"refresh",
      "cmd_status":"Success",
      "timestamp":"2020-08-20T18:44:43.574977236Z",
      "exit_status":0,
      "stderr":""
   }
}
```

pkg/platform/api/platform.go
etungsten: This is what I originally had. I removed it during my PR by mistake at some point. Sorry for the trouble :(
https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/ccf3431090e036828bf511162d9e0b7cd927df60/pkg/platform/api/platform.go#L51-L54

jahkeup: Yeah, there's no call sites that need the array to be hydrated for use. Updated to use `return nil`.

PR #51, opened 2020-11-20
Bring Go to 1.15 for build

gliptak:
**Terms of contribution:**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Setup Go version ${{ matrix.go }}
         uses: actions/setup-go@v1
         with:
-          go-version: 1.14.1
+          go-version: 1.15.5
       - name: Checkout
         uses: actions/checkout@v2
       - name: Build native


gliptak:
@jahkeup @webern please review

.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: Setup Go version ${{ matrix.go }}
         uses: actions/setup-go@v1
         with:
-          go-version: 1.14.1
+          go-version: 1.15.5
```
jahkeup, line 17: The version of Go used to build here should match [Go in the Bottlerocket SDK](https://github.com/bottlerocket-os/bottlerocket-sdk/blob/9455f4b192be0399b49aec97c29cf0c331f217e0/Dockerfile#L375). The SDK's distributed Go toolchain was bumped recently to `1.15.4`, but the update operator's release image builds aren't using that version of the SDK yet.

Until we've caught up to that version, the build should continue to use the same version of Go the SDK provides: [`1.14.1`](https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/369a4c7ba2680916522b60a98f77768fa987888d/Dockerfile#L8-L10).

In the future, this could be expanded into a matrix with at least the SDK's Go version with a 'latest' and the SDK version.

jahkeup:
At this time, we are unable to accept this PR. This bump is expected for the work to update the SDK - see #52 for the "next" bump. 

Please open an issue to discuss follow up changes here - like the mentioned matrix build. We can't guarantee that we'll be able to accept a change and it is painful to have to turn away contributions.

If you are able to, please create issues first - simple as they might be. Issues give us a place discuss, to reach agreement, and to accept the associated change before its even made real in a PR.

Issue #54, opened 2021-02-12
Clarify updater-interface-version


bgdnlp:
This is more of a documentation request. Would it be possible to clarify a few things regarding `updater-interface-version` in the README? It took some digging for me to find out. Like:
1. 1.0.0 vs 2.0.0 (updog vs API). It's unclear what is version 1 and what is v2 and why they are versioned. Simply using "updog" vs "api" instead would help, I see there was a comment about it in the pull request.
2. When should one use Updog instead of API, and is there any reason to use "1.0.0" on Bottlerocket v1.0+?
3. In cases where API si fine, should the `update-operator-agent-updog` DaemonSet be removed from the K8s manifest? 600 MiB is a lot of memory for a container that runs on every node. And the manifest has two of those. That's 1.2 GiB, which is huge. See #47.

etungsten:
Hi @bgdnlp 

1. The platform version lets us differentiate deployments that might be specific to certain host versions. The only use right now is to differentiate between hosts that support the update api and hosts that do not. If the host API surface changes in a incompatible way, we would need to introduce new platform versions.
2. The updog deployment should only be used with bottlerocket hosts that do not support the update API (any Bottlerocket version < 0.4.1). This is definitely a documentation issue we need to address. The default recommended updater-interface-version node label to apply should be 2.0.0
3. The updog agent pod should not run if the updater-interface-version is set to 2.0.0. Conversely the update api agent won’t run if the updater-interface-version isn’t set to 2.0.0. We’ll adjust the memory reserved for the update-api agent DaemonSet to something more befitting.

Issue #47, opened 2020-09-30
Can we reduce the memory of 600Mi in the update-operator daemon sets configuration?


shthalla:
**Feature Request:**

The bottlerocket-update-operator is configured to request 600 MB of memory by default:

https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/3c91edb2423557445b37f2407c9ba1daf805914e/update-operator.yaml#L185

https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/3c91edb2423557445b37f2407c9ba1daf805914e/update-operator.yaml#L262

Can we reduce the memory less than 600Mi for both updog and also the API DaemonSet in the file and deploy?

I am sure it will work initially with out any pods and the load. but not sure in future do we face any issue because of this modifications. 

Can we safely run it with less memory?  Also want to know why it is recommended 600 MB as the request for this both DaemonSets?

jahkeup:
Thanks for cutting this issue @shthalla :+1: 

> Can we safely run it with less memory?

For the `api` driven agent Pod: I strongly suspect this is much higher than necessary. That said, we'll need to confirm before reducing the value. For the `updog` updater, we can't reduce the value - `updog` might exhaust the Pod's memory during an update (I'll touch on this below).

> Also want to know why it is recommended 600 MB as the request for this both DaemonSets?

Originally, the agent in a given Pod would have selected which updater to use - either `updog` or `api` (that is there'd only be a single DaemonSet). The `updog` updater would load disk images within the context of the container (through a chroot). This update method required a fair amount of memory as result - experience in operation informed the decision to set the value at 600 MB. We're not going to be able to change that as that concerns the operation of `updog` itself on existing hosts. In the case of the `api` method, the updater does not run up against the Pod's memory constraints. Instead, the API calls made to the host drive the updater outside in the host's namespace. 

There was a last minute change made to use a per-updater-implementation (chosen based on the nodes' labels) and the required memory value was not updated alongside. We'll need to validate a lowered setting and then update the deployment for these Pods.

bgdnlp:
If I understand correctly, using the recommended deployment manifest reserves 1.2 GiB of RAM on each node for a couple of pods that most of the time do nothing. Which means that any machine with 2 GiB of RAM or less is effectively unusable as a node. This is a big deal for something that is supposed to be lean and minimalistic.

[...]

PR #55, opened 2021-02-15
Update docs, update cpu/memory request, limits for update-api agent DaemonSet

etungsten:
**Issue number:**
https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/54
https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/47

**Description of changes:**
```
Author: Erikson Tung <etung@amazon.com>
Date:   Mon Feb 15 13:17:29 2021 -0800

    deployment: change cpu and memory request, limits for api agent pods
    
    Significantly lower cpu, memory request/limits for the update-api agent
    DaemonSet


```

```
commit 96771ced4149e95ebc92ae65aeb63a436e573d74
Author: Erikson Tung <etung@amazon.com>
Date:   Mon Feb 15 13:05:04 2021 -0800

    README: change recommended update interface version to 2.0.0, add desc
    
    Update documentation to include short descriptions of the update
    interface versions. Include compatibility information

```

**Testing done:**

Monitored the memory usage of the agent using the update API. I never saw it peak above 25MB in physical memory usage.
Tried different CPU request values, the agent still works fine in updating nodes when requesting 10 millicores.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,27 +52,31 @@ The agent relies on each node's updater components and schedules its pods based 
 The node indicates its updater interface version in a label called `bottlerocket.aws/updater-interface-version`.
 Agent deployments, respective to the interface version, are scheduled using this label and target only a single version in each.
 
-For the `1.0.0` `updater-interface-version`, this label looks like:
+- For Bottlerocket OS versions >= v0.4.1, we recommend using `update-interface-version` 2.0.0 to leverage Bottlerocket's API to dispatch updates.
+- Bottlerocket OS versions < v0.4.1 are only compatible with `update-interface-version` 1.0.0.
+  - With this version, the agent needs to run in a priviledged container with access to the root filesystem.
+
+For the `2.0.0` `updater-interface-version`, this label looks like:
 
 ``` text
-bottlerocket.aws/updater-interface-version=1.0.0
+bottlerocket.aws/updater-interface-version=2.0.0
 ```
 
 `kubectl` can be used to set this label on a node in the cluster:
 
 ``` sh
-kubectl label node $NODE_NAME bottlerocket.aws/updater-interface-version=1.0.0
+kubectl label node $NODE_NAME bottlerocket.aws/updater-interface-version=2.0.0
 ```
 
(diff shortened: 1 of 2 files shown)


bgdnlp:
Should the [example development deployment](https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/develop/dev/deployment.yaml) that is linked to just below in the README also be adjusted or maybe removed?

To me it seems to be a bit out of date and the suggested deployment looks simple enough to serve as a starting point, I would completely remove that line.

jahkeup (approved):

srgothi92 (approved):

etungsten:
Push above removes mention of the the development deployment as a customization starting point since that might be outdated.

Fixing it up should be addressed separately

README.md
```diff
@@ -76,7 +76,7 @@ kubectl label node $(kubectl get nodes -o jsonpath='{.items[*].metadata.name}')
 
 Each workload resource may have additional constraints or scheduling affinities based on each node's labels in addition to the `bottlerocket.aws/updater-interface-version` label scheduling constraint.
 
-Customized deployments may use the [suggested deployment](./update-operator.yaml) or the [example development deployment](./dev/deployment.yaml) as a starting point, with customized container images specified if needed.
+Customized deployments may use the [suggested deployment](./update-operator.yaml) as a starting point, with customized container images specified if needed.
```
srgothi92, line 79: was this change planned to be part of another commit where you are updating the read me?

srgothi92 (approved):
Looks good! except for one minor confusion of commits I think. Multiple commit are pain to update especially if change is required in last commit in the chain. 

> Fixing it up should be addressed separately

Do you know where we use `./dev/deployment.yaml` ? Should we delete it instead of fixing it ? I feel we should delete `kind-cluster.yaml ` as well, we cannot use that as direct alternative to our actual bottlerocket cluster. It adds more confusion then to help. However, all of these suggestion are not for this PR, we can create an issue for cleanup.

README.md
etungsten: Ah, missed that. I'll fix it.

Fixed below

PR #84, opened 2021-10-06
Add docker build for brupop 0.2.0

cbgbt:
**Issue number:**
N/A

**Description of changes:**
This creates a singular container image which contains three Rust binaries which will be needed for brupop. The Dockerfile includes a few changes to the SDK to allow the `openssl-sys` crate to link against an openssl built with musl.

**Testing done:**
Running `make` results in a container image with the three distinct binaries.

**Terms of contribution:**


cbgbt:
I've added github workflows to check image and build creation here so that subsequent PRs will have the checks enabled, so there will be a bit of churn here while the workflows are being configured correctly. If they pass, then the PR should be ready.

.github/workflows/image.yaml
```diff
@@ -0,0 +1,15 @@
+name: Images
+on:
+  pull_request:
+    paths-ignore:
+      - '**.md'
```
jpculp, line 5: It might be worth ignoring
```
      - 'COPYRIGHT'
      - 'LICENSE**'
```

as well since it shouldn't impact the build artifacts.

.github/workflows/image.yaml
```diff
@@ -0,0 +1,15 @@
...
+  push:
```
jpculp: It might not be necessary to rebuild the artifacts on push. I would consider a separate workflow for push later down the line.

Makefile
```diff
@@ -0,0 +1,16 @@
+.PHONY: build sdk-openssl brupop-image
```
jpculp, line 1: Since this is being written in Rust, it would be a good idea to add a license check using `cargo-deny`. That could be added as a separate PR later down the road though.

cbgbt: Great tip! I wasn't aware of that tool.

cbgbt: I've created issue #87  for this!

cbgbt:
Building `openssl` here isn't ideal, but we would need support for statically linking against it in the Bottlerocket SDK to do away with that. I've opened #88 to stop building Openssl during our build, with a few options listed.

jpculp (approved):

arnaldo2792 (approved):
:boot:

PR #85, opened 2021-10-06
Add kubernetes models for API server, and yaml generation

### LINKED ISSUE #77: brupop 0.2.0 apiserver: Create k8s deployment targets for development
We'll use [`kube-rs`](https://github.com/kube-rs/kube-rs) to autogenerate kubernetes objects for deployments. This task tracks creating targets (or at least documentation) to create a cluster using the autogenerated objects.

cbgbt:
**Issue number:** #77 

**Description of changes:**
This change adds a build step which converts our Rust models into YAML files which can be deployed to kubernetes.

Included is a client which can be used to create and update the `BottlerocketNode` custom resources

**Testing done:**
The YAML files were applied to an EKS cluster, and then a `BottlerocketNodeClient` was used to create a `BottlerocketNode` custom resource, and update its status:

```
$ kubectl describe brn
Name:         brn-ip-192-168-66-198.us-west-2.compute.internal
Labels:       <none>
Annotations:  <none>
Metadata:
  Creation Timestamp:  2021-10-05T19:41:03Z
  Generation:          1
  Managed Fields:
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:ownerReferences:
          .:
          k:{"uid":"623ca9b3-bfc6-4c0f-b5d8-d8bd320a0a2f"}:
            .:
            f:apiVersion:
            f:kind:
            f:name:
            f:uid:
      f:spec:
        .:
        f:state:
        f:version:
      f:status:
        .:
        f:available_versions:
        f:current_state:
        f:current_version:
    Manager:    unknown
    Operation:  Update
    Time:       2021-10-05T19:41:03Z
  Owner References:
    API Version:     v1
    Name:            ip-192-168-66-198.us-west-2.compute.internal
    UID:             623ca9b3-bfc6-4c0f-b5d8-d8bd320a0a2f
  Resource Version:  36761
  UID:               0d69e179-dbda-4058-bf4f-8e099b517a52
Spec:
  State:    WaitingForUpdate
  Version:  <nil>
Status:
  available_versions:
    1.3.0
  current_version:  1.2.1
Events:             <none>
```

**Terms of contribution:**

### DIFF
--- a/models/Cargo.toml
+++ b/models/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "models"
+version = "0.1.0"
+edition = "2018"
+publish = false
+license = "Apache-2.0 OR MIT"
+
+[dependencies]
+async-trait = "0.1"
+# k8s-openapi must match the version required by kube and enable a k8s version feature
+k8s-openapi = { version = "0.13.0", default-features = false, features = ["v1_20"] }
+kube = { version = "0.59.0", default-features = true, features = [ "derive"] }
+kube-runtime = "0.59.0"
+
+maplit = "1.0"
+schemars = "0.8"
+serde = { version = "1", features = [ "derive" ] }
+serde_json = "1"
+snafu = "0.6"
--- a/models/src/apiserver.rs
+++ b/models/src/apiserver.rs
@@ -1,0 +1,257 @@
+use crate::constants::{
+    APISERVER, APISERVER_HEALTH_CHECK_ROUTE, APISERVER_INTERNAL_PORT, APISERVER_MAX_UNAVAILABLE,
+    APISERVER_SERVICE_NAME, APISERVER_SERVICE_PORT, APP_COMPONENT, APP_MANAGED_BY, APP_PART_OF,
+    BRUPOP, BRUPOP_DOMAIN_LIKE_NAME, LABEL_COMPONENT, NAMESPACE,
+};
+use crate::node::{K8S_NODE_PLURAL, K8S_NODE_STATUS};
+use k8s_openapi::api::apps::v1::{
+    Deployment, DeploymentSpec, DeploymentStrategy, RollingUpdateDeployment,
+};
+use k8s_openapi::api::core::v1::{
+    Affinity, Container, ContainerPort, HTTPGetAction, LocalObjectReference, NodeAffinity,
+    NodeSelector, NodeSelectorRequirement, NodeSelectorTerm, PodSpec, PodTemplateSpec, Probe,
+    Service, ServiceAccount, ServicePort, ServiceSpec,
+};
+use k8s_openapi::api::rbac::v1::{ClusterRole, ClusterRoleBinding, PolicyRule, RoleRef, Subject};
+use k8s_openapi::apimachinery::pkg::apis::meta::v1::LabelSelector;
+use k8s_openapi::apimachinery::pkg::util::intstr::IntOrString;
+use kube::api::ObjectMeta;
+use maplit::btreemap;
+
+const BRUPOP_APISERVER_SERVICE_ACCOUNT: &str = "brupop-controller-service-account";
+const BRUPOP_APISERVER_CLUSTER_ROLE: &str = "brupop-controller-role";
+
+/// Defines the brupop-apiserver service account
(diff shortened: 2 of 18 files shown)


models/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
```
jpculp, line 1: It's probably a good idea to add authors and license information to some of these Cargo tomls.

models/src/namespace.rs
```diff
@@ -0,0 +1,19 @@
...
+use k8s_openapi::api::core::v1::Namespace;
+use kube::api::ObjectMeta;
+use maplit::btreemap;
+
+/// Defines the brupop namespace
+pub fn brupop_namespace() -> Namespace {
+    Namespace {
+        metadata: ObjectMeta {
+            labels: Some(btreemap! {
+                "name".to_string() => "brupop".to_string()
```
arnaldo2792, line 11: Nit: shouldn't this `brupop` be a constant somewhere?

models/src/apiserver.rs
```diff
@@ -0,0 +1,252 @@
...
+            name: Some(BRUPOP_APISERVER_CLUSTER_ROLE.to_string()),
+            namespace: Some(NAMESPACE.to_string()),
+            ..Default::default()
+        },
+        rules: Some(vec![
+            PolicyRule {
+                api_groups: Some(vec![BRUPOP.to_string()]),
+                resources: Some(vec![
+                    "bottlerocketnodes".to_string(),
+                    "bottlerocketnodes/status".to_string(),
```
arnaldo2792: Nit: should these two be their own constants?

models/src/node.rs
```diff
@@ -0,0 +1,213 @@
...
+        selector.node_uid,
+        source
+    ))]
+    CreateBottlerocketNode {
+        source: Box<dyn std::error::Error>,
+        selector: BottlerocketNodeSelector,
+    },
+
+    #[snafu(display(
+        "Unable to update BottlerocketNode status ({}, {}): {}",
```
arnaldo2792: Nit: I remember folks usually ask to put error messages between `''`:

```suggestion
        "Unable to update BottlerocketNode status ({}, {}): '{}'",
```

models/src/node.rs
```diff
@@ -0,0 +1,213 @@
...
+    async fn update_node_status(
+        &self,
+        selector: &BottlerocketNodeSelector,
+        status: &BottlerocketNodeStatus,
+    ) -> Result<(), BottlerocketNodeError> {
+        let br_node_patch = json!({
+            "apiVersion": constants::API_VERSION,
+            "kind": "BottlerocketNode",
+            "status": status
+        });
```
arnaldo2792: Will this be at any point its own serializable struct? Probably with `serde_json`?

arnaldo2792 (approved):
:boot:

gthao313 (approved):

models/src/node.rs
```diff
@@ -0,0 +1,213 @@
...
+    state: BottlerocketNodeState,
+    version: Option<String>,
+}
+
+/// `BottlerocketNodeStatus` surfaces the current state of a bottlerocket node. The status is updated by the host agent,
+/// while the spec is updated by the brupop controller.
+#[derive(Clone, Debug, Default, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
+pub struct BottlerocketNodeStatus {
+    current_version: String,
+    available_versions: Vec<String>,
```
gthao313, line 90: `available_versions` might be an empty Vec which means current version is the latest version, I think it's better to use `Option<Vec<String>>`

models/Cargo.toml
cbgbt: These fell out of a rebase. I'll re-add them!

models/src/node.rs
cbgbt: I think it's only best to wrap a `Vec` in `Option` if there is a meaningful distinction between an empty vector, and a null one. In this case, both represent the absence of updates, so it's probably best to only allow a single representation of it if possible. I think not using `Option` is more clear here.

jpculp (approved):

Issue #76, opened 2021-09-27
brupop 0.2.0 apiserver: k8s token-based authorization


cbgbt:
brupop agents will utilize [token volume projection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection) to receive a pod-unique token. The brupop apiserver should accept this data in a header for requests to modify any `BottlerocketNode` custom resources. It should use this token to determine the identity of the requester using the k8s Token APIs, and use that information to authorize the requests to modify said custom resources.

PR #86, opened 2021-10-07
0.2.0 Initial apiserver implementation

### LINKED ISSUE #75: brupop 0.2.0 apiserver: Create/update node endpoints
Create the web endpoints that agents use to modify the custom resources associated with their nodes. The APIServer should accept JSON-encoded requests containing the `BottlerocketNode` object state and update the correct custom resource so that its `status` reflects that state.

cbgbt:
**Issue number:** closes #75 

**Description of changes:** This change creates initial API endpoints to allow brupop agents to create and modify their BottlerocketNode resources. It doesn't yet handle token-based authorization (which is the purpose for creating this apiserver). That will be tracked in #76.

**Testing done:** The provided tests pass. Additionally, I spun up an EKS cluster containing the service, then made an HTTP request to one of the nodes creating and updating a `brn` resource, which succeeded.

```
Sending POST request with body {"node_selector":{"node_name":"<redacted>","node_uid":"623ca9b3-bfc6-4c0f-b5d8-d8bd320a0a2f"}}
Sending PUT request with body {"node_selector":{"node_name":"<redacted>","node_uid":"623ca9b3-bfc6-4c0f-b5d8-d8bd320a0a2f"},"node_status":{"current_version":"1.2.1","available_versions":["1.3.0"],"current_state":"WaitingForUpdate"}}
{"current_version":"1.2.1","available_versions":["1.3.0"],"current_state":"WaitingForUpdate"}
[ec2-user@ip-10-0-0-27 bottlerocket-update-operator]$ kubectl describe brn
Name:         brn-<redacted>
Labels:       <none>
Annotations:  <none>
Metadata:
  Creation Timestamp:  2021-10-07T00:27:59Z
  Generation:          1
  Managed Fields:
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:ownerReferences:
          .:
          k:{"uid":"623ca9b3-bfc6-4c0f-b5d8-d8bd320a0a2f"}:
            .:
            f:apiVersion:
            f:kind:
            f:name:
            f:uid:
      f:spec:
        .:
        f:state:
        f:version:
      f:status:
        .:
        f:available_versions:
        f:current_state:
        f:current_version:
    Manager:    unknown
    Operation:  Update
    Time:       2021-10-07T00:27:59Z
  Owner References:
    API Version:     v1
    Name:            <redacted>
    UID:             623ca9b3-bfc6-4c0f-b5d8-d8bd320a0a2f
  Resource Version:  283524
  UID:               b23911e1-f589-4141-a409-fefeb991f0ef
Spec:
  State:    WaitingForUpdate
  Version:  <nil>
Status:
  available_versions:
    1.3.0
  current_version:  1.2.1
Events:             <none>
```

**Terms of contribution:**

### DIFF
--- a/apiserver/src/api.rs
+++ b/apiserver/src/api.rs
@@ -1,0 +1,232 @@
+use crate::error::{self, Result};
+use models::constants::APISERVER_HEALTH_CHECK_ROUTE;
+use models::node::{BottlerocketNodeClient, BottlerocketNodeSelector, BottlerocketNodeStatus};
+
+use actix_web::{
+    middleware,
+    web::{self, Data},
+    App, HttpRequest, HttpResponse, HttpServer, Responder,
+};
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+use snafu::ResultExt;
+
+const NODE_RESOURCE_ENDPOINT: &'static str = "/bottlerocket-node-resource";
+
+/// HTTP endpoint that implements a shallow health check for the HTTP service.
+async fn health_check() -> impl Responder {
+    HttpResponse::Ok().body("pong")
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+/// Describes a node for which a BottlerocketNode custom resource should be constructed.
+pub struct CreateBottlerocketNodeRequest {
+    pub node_selector: BottlerocketNodeSelector,
+}
+
+/// HTTP endpoint which creates BottlerocketNode custom resources on behalf of the caller.
+async fn create_bottlerocket_node_resource<T: BottlerocketNodeClient>(
+    _req: HttpRequest,
+    settings: web::Data<APIServerSettings<T>>,
+    create_request: web::Json<CreateBottlerocketNodeRequest>,
+) -> Result<impl Responder> {
+    let br_node = settings
+        .node_client
+        .create_node(&create_request.node_selector)
+        .await
+        .context(error::BottlerocketNodeCreate)?;
+
+    Ok(HttpResponse::Ok().body(format!("{}", json!(&br_node))))
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+/// Describes updates to a BottlerocketNode object's `status`.
+pub struct UpdateBottlerocketNodeRequest {
+    pub node_selector: BottlerocketNodeSelector,
+    pub node_status: BottlerocketNodeStatus,
+}
+
+/// HTTP endpoint which updates the `status` of a BottlerocketNode custom resource on behalf of the caller.
+async fn update_bottlerocket_node_resource<T: BottlerocketNodeClient>(
+    _req: HttpRequest,
+    settings: web::Data<APIServerSettings<T>>,
+    update_request: web::Json<UpdateBottlerocketNodeRequest>,
+) -> Result<impl Responder> {
+    settings
+        .node_client
+        .update_node_status(&update_request.node_selector, &update_request.node_status)
+        .await
+        .context(error::BottlerocketNodeUpdate)?;
+
+    Ok(HttpResponse::Ok().body(format!("{}", json!(&update_request.node_status))))
+}
+
+#[derive(Clone)]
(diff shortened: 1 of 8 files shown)


apiserver/src/main.rs
```diff
@@ -1,3 +1,33 @@
...
+
+use snafu::ResultExt;
+
+use std::fs;
+
+#[actix_web::main]
+async fn main() {
+    match run_server().await {
+        Err(error) => {
+            fs::write("/dev/termination-log", format!("{}", error))
```
arnaldo2792: Nit: Should `/dev/termination-log` be a constant?

Also, I'm confused, why is this error written to `/dev/termination-log`, but the server errors to whatever the environment variables have?

apiserver/src/main.rs
```diff
@@ -1,3 +1,33 @@
...
+async fn run_server() -> Result<()> {
+    env_logger::init();
+
+    let k8s_client = kube::client::Client::try_default()
+        .await
+        .context(error::ClientCreate)?;
+
+    let settings = APIServerSettings {
+        node_client: K8SBottlerocketNodeClient::new(k8s_client),
+        server_port: 8080,
```
arnaldo2792, line 37: Maybe use an env variable or mount configuration file to fetch the port in which the server has to run?

apiserver/src/api.rs
```diff
@@ -0,0 +1,230 @@
...
+    pub node_selector: BottlerocketNodeSelector,
+}
+
+/// HTTP endpoint which creates BottlerocketNode custom resources on behalf of the caller.
+async fn create_bottlerocket_node_resource<T: BottlerocketNodeClient>(
+    _req: HttpRequest,
+    settings: web::Data<APIServerSettings<T>>,
+    create_request: web::Json<CreateBottlerocketNodeRequest>,
+) -> Result<impl Responder> {
+    let br_node = settings
```
arnaldo2792, line 33: Maybe add some `trace!` logging here, and log some information from the http request? Or are we getting logs from actix?

apiserver/src/error.rs
```diff
@@ -0,0 +1,26 @@
...
+use snafu::Snafu;
+
+/// The crate-wide result type.
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// The crate-wide error type.
+#[derive(Debug, Snafu)]
+#[snafu(visibility = "pub")]
+pub enum Error {
+    #[snafu(display("Unable to create client: {}", source))]
```
arnaldo2792: Nit: same as in other PRs, people usually ask me to put the error messages between `''`

apiserver/src/main.rs
cbgbt: In this case, whatever is bubbled up as a server-terminating error is written to the termination log, which is a location that kubernetes uses to bubble up error messages to the cluster administrator.

I agree that it should be a constant, or decided by an environment variable.

apiserver/src/api.rs
cbgbt: Actix provides middleware which logs the request and information about the response, but agreed that we could add additional logging about the actions taking place during the request.

apiserver/src/main.rs
cbgbt: In this case, this is the port used inside the server container itself, which needs to be used in the kubernetes deployment files as well. It should probably be a constant so that it's easy to cross reference through `brupop's` repo, but I don't think there's a use-case which would require changing it: Folks can use port mapping at the cluster level to decide which port is actually used for communication.

gthao313 (approved):

apiserver/src/api.rs
cbgbt: I've been investigating adding support for tracing via [tracing-actix-web](https://github.com/LukeMathWalker/tracing-actix-web); however, the changes are large enough that I'd like to add them in a subsequent PR. I've opened #94 to track that. Is it okay to proceed here without that feature for now?

arnaldo2792 (approved):
:motor_boat:

Issue #80, opened 2021-09-27
0.2.0 agent: Implement token-based authorization


cbgbt:
The brupop agent must use a token provided by [Token Volume Projection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection) and include it in requests to the `apiserver` to update `BottlerocketNode` custom resources.

cbgbt:
This should be resolved by #116

Issue #79, opened 2021-09-27
0.2.0 agent: Implement `BottlerocketNode` Custom Resource Updates


cbgbt:
The brupop agent must gather the current state of the system and then compare it to that stored in k8s as a custom resource. This task involves gathering the system state and comparing it to the existing resource, then making requests to the `apiserver` if any differences are detected.

Issue #78, opened 2021-09-27
0.2.0 agent: Implement Fetching node-associated custom resource


cbgbt:
This task is straightforward, but also implies having k8s configured on the client. The `apiserver` is only needed for write operations to the `BottlerocketNode` custom resources. The node will need to implement fetching its the `status` of its current resource in order to compare it to the state gathered on the node.

PR #92, opened 2021-10-14
0.2.0: Add initial implementation of agent

gthao313:
**Issue number:**
#80 #78  #79

**Description of changes:**
Implementing agent component which have two responsibilities
- Gather metadata about the system and update the custom resource associated with this node
- Determine if the spec on the system's custom resource demands the node take action. If so, begin taking that action.   

**Testing done:**
Creating agent for each node by apply deamonset
Test lists
- [x] If agent can correctly create bottlerocketnode (custom resource)
- [x] If agent can correctly update bottlerocketnode status (custom resource)
- [x] If agent can interact with Bottlerocket Update API via apiclient

```
[tianhg@ip-172-31-39-243 bottlerocket-update-operator]$ kubectl -n brupop-bottlerocket-aws describe brn brn-ip-192-168-66-248.us-west-2.compute.internal
Name:         brn-ip-192-168-66-248.us-west-2.compute.internal
Labels:       <none>
Annotations:  <none>
Metadata:
  Creation Timestamp:  2021-10-26T06:06:04Z
  Generation:          1
  Managed Fields:
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:ownerReferences:
          .:
          k:{"uid":"3905a2e5-e2df-4c58-ab86-9ecc33d60997"}:
            .:
            f:apiVersion:
            f:kind:
            f:name:
            f:uid:
      f:spec:
        .:
        f:state:
        f:version:
      f:status:
        .:
        f:available_versions:
        f:current_state:
        f:current_version:
    Manager:    unknown
    Operation:  Update
    Time:       2021-10-26T06:06:04Z
  Owner References:
    API Version:     v1
    Name:            ip-192-168-66-248.us-west-2.compute.internal
    UID:             3905a2e5-e2df-4c58-ab86-9ecc33d60997
  Resource Version:  47784043
  Self Link:         /apis/brupop.bottlerocket.aws/v1/namespaces/brupop-bottlerocket-aws/bottlerocketnodes/brn-ip-192-168-66-248.us-west-2.compute.internal
  UID:               f90796e1-c6e3-4963-bfaa-2c6b45d103a9
Spec:
  State:    WaitingForUpdate
  Version:  <nil>
Status:
  available_versions:
    1.3.0
    1.2.1
    1.2.0
    1.1.4
    1.1.3
    1.1.2
    1.1.1
    1.1.0
    1.0.8
    1.0.7
    1.0.6
    1.0.5
    1.0.4
    1.0.3
    1.0.2
  current_version:  1.1.1
```
**Terms of contribution:**

### DIFF
--- a/agent/Cargo.toml
+++ b/agent/Cargo.toml
@@ -6,3 +6,22 @@ edition = "2018"
 publish = false
 
 [dependencies]
+models = { path = "../models", version = "0.1.0" }
+apiserver = { path = "../apiserver", version = "0.1.0" }
+
+dotenv = "0.15"
+env_logger = "0.9"
+futures = "0.3"
+
+# k8s-openapi must match the version required by kube and enable a k8s version feature
+k8s-openapi = { version = "0.13.0", default-features = false, features = ["v1_20"] }
+kube = { version = "0.59.0", default-features = true, features = [ "derive"] }
+kube-runtime = "0.59.0"
+
+log = "0.4"
+serde = { version = "1", features = [ "derive" ] }
+serde_json = "1"
+snafu = "0.6"
+tokio = { version = "1", features = ["macros", "rt-multi-thread", "time"] }
+reqwest = { version = "0.11", features =  [ "json" ] }
+chrono = { version = "0.4.11", features = [ "serde" ] }
--- a/agent/src/agentclient.rs
+++ b/agent/src/agentclient.rs
@@ -1,0 +1,310 @@
+use chrono::naive::NaiveDateTime;
+use k8s_openapi::api::core::v1::Node;
+use kube::Api;
+use snafu::OptionExt;
+use snafu::ResultExt;
+use std::env;
+use std::process::Command;
+use tokio::time::{sleep, Duration};
+
+use crate::apiclient::{boot_update, get_os_info, list_available, prepare, update};
+use crate::error::{self, Result};
+use apiserver::api::{CreateBottlerocketNodeRequest, UpdateBottlerocketNodeRequest};
+use models::constants::{APISERVER_SERVICE_NAME, APISERVER_SERVICE_PORT, NAMESPACE};
+use models::node::{
+    node_resource_name, BottlerocketNode, BottlerocketNodeSelector, BottlerocketNodeState,
+    BottlerocketNodeStatus,
+};
+
+const AGENT_SLEEP_DURATION: Duration = Duration::from_secs(5);
+const NODE_RESOURCE_ENDPOINT: &'static str = "/bottlerocket-node-resource";
+
+#[derive(Clone)]
+pub struct BrupopAgent {
+    k8s_client: kube::client::Client,
+    node_selector: Option<BottlerocketNodeSelector>,
+}
+
+impl BrupopAgent {
+    pub fn new(k8s_client: kube::client::Client) -> Self {
+        BrupopAgent {
+            k8s_client,
+            node_selector: None,
+        }
+    }
+
+    pub async fn check_node_custom_resource_exists(&mut self) -> Result<bool> {
+        let associated_bottlerocketnode_name = node_resource_name(&self.get_node_selector().await?);
+        let bottlerocket_nodes: Api<BottlerocketNode> =
+            Api::namespaced(self.k8s_client.clone(), NAMESPACE);
+
+        Ok(bottlerocket_nodes
+            .get(&associated_bottlerocketnode_name)
+            .await
+            .context(error::BottlerocketNodeNotExist {
+                node_name: associated_bottlerocketnode_name,
+            })
+            .is_ok())
+    }
+
+    pub async fn check_custom_resource_status_exists(&mut self) -> Result<bool> {
+        Ok(self.fetch_custom_resource().await?.status.is_some())
+    }
+
+    async fn get_node_selector(&mut self) -> Result<BottlerocketNodeSelector> {
+        if self.node_selector.is_none() {
+            let associated_node_name = env::var("MY_NODE_NAME").context(error::GetNodeName)?;
+
+            let nodes: Api<Node> = Api::all(self.k8s_client.clone());
+            let associated_node: Node = nodes
+                .get(&associated_node_name)
+                .await
+                .context(error::FetchNode)?;
+
+            let associated_node_uid = associated_node
+                .metadata
+                .uid
+                .context(error::MissingNodeUid)?;
+            self.node_selector = Some(BottlerocketNodeSelector {
+                node_name: associated_node_name.clone(),
+                node_uid: associated_node_uid.clone(),
+            });
+        }
+
+        Ok(self
+            .node_selector
+            .clone()
+            .context(error::NodeSelectorIsNone)?)
+    }
+
+    // fetch associated bottlerocketnode (custom resource) and help to get node `metadata`, `spec`, and  `status`.
+    async fn fetch_custom_resource(&mut self) -> Result<BottlerocketNode> {
+        // get associated bottlerocketnode (custom resource) name, and we can use this
+        // bottlerocketnode name to fetch targeted bottlerocketnode (custom resource) and get node `metadata`, `spec`, and  `status`.
+
+        //TODO: use a reflector to watch event instead of polling etcd
+        let associated_bottlerocketnode_name =
+            format!("brn-{}", self.get_node_selector().await?.node_name);
+
+        let bottlerocket_nodes: Api<BottlerocketNode> =
+            Api::namespaced(self.k8s_client.clone(), NAMESPACE);
+        let associated_bottlerocketnode: BottlerocketNode = bottlerocket_nodes
+            .get(&associated_bottlerocketnode_name)
+            .await
+            .context(error::FetchCustomResource)?;
+
+        Ok(associated_bottlerocketnode)
+    }
+
+    /// Gather metadata about the system, node status provided by spec.
+    async fn gather_system_metadata(
+        &mut self,
+        state: BottlerocketNodeState,
+    ) -> Result<BottlerocketNodeStatus> {
+        let current_version = get_os_info()
+            .await
+            .context(error::BottlerocketNodeStatusVersion)?;
+        let update_versions = list_available()
+            .await
+            .context(error::BottlerocketNodeStatusAvailableVersions)?;
+
+        Ok(BottlerocketNodeStatus {
+            current_version: current_version.version_id.clone(),
+            available_versions: update_versions,
+            current_state: state,
+        })
+    }
+
+    /// create the custom resource associated with this node
+    pub async fn create_metadata_custom_resource(&mut self) -> Result<()> {
+        let client = reqwest::Client::new();
+        let current_node_selector = self.get_node_selector().await?;
+        let node_req = CreateBottlerocketNodeRequest {
+            node_selector: current_node_selector.clone(),
+        };
+        let server_domain = get_server_domain();
+
+        let response = client
+            .post(format!(
+                "http://{}{}",
+                &server_domain, NODE_RESOURCE_ENDPOINT
+            ))
+            .json(&node_req)
+            .send()
+            .await
+            .context(error::CreateBottlerocketNodeResource {
+                node_name: current_node_selector.node_name,
+            })?;
+
+        log::info!(
+            "{}",
+            response
+                .text()
+                .await
+                .context(error::ConvertResponseToText)?
+        );
+
+        Ok(())
+    }
+
+    /// update the custom resource associated with this node
+    async fn update_metadata_custom_resource(
+        &mut self,
+        _current_metadata: BottlerocketNodeStatus,
+    ) -> Result<()> {
+        let client = reqwest::Client::new();
+        let current_node_selector = self.get_node_selector().await?;
+        let server_domain = get_server_domain();
+
+        let node_req = UpdateBottlerocketNodeRequest {
+            node_status: _current_metadata,
+            node_selector: current_node_selector.clone(),
+        };
+
+        let response = client
+            .put(format!(
+                "http://{}{}",
+                &server_domain, NODE_RESOURCE_ENDPOINT
+            ))
+            .json(&node_req)
+            .send()
+            .await
+            .context(error::UpdateBottlerocketNodeResource {
+                node_name: current_node_selector.node_name,
+            })?;
+
+        log::info!(
+            "{}",
+            response
+                .text()
+                .await
+                .context(error::ConvertResponseToText)?
+        );
+
+        Ok(())
+    }
+
+    /// initialize bottlerocketnode (custom resource) `status` when create new bottlerocketnode
+    pub async fn initialize_metadata_custom_resource(&mut self) -> Result<()> {
+        let update_node_status = self
+            .gather_system_metadata(BottlerocketNodeState::WaitingForUpdate)
+            .await?;
+
+        self.update_metadata_custom_resource(update_node_status)
+            .await?;
+        Ok(())
+    }
+
+    pub async fn run(&mut self) -> Result<()> {
+        // A running agent has two responsibilities:
+        // - Gather metadata about the system and update the custom resource associated with this node
+        // - Determine if the spec on the system's custom resource demands the node take action. If so, begin taking that action.
+
+        loop {
+            // Requests metadata 'status' and 'spec' for the current BottlerocketNode
+            let bottlerocket_node = self.fetch_custom_resource().await?;
+
+            let bottlerocket_node_status = bottlerocket_node
+                .status
+                .context(error::MissingBottlerocketNodeStatus)?;
+            let bottlerocket_node_spec = bottlerocket_node.spec;
+
+            // Determine if the spec on the system's custom resource demands the node take action. If so, begin taking that action.
+            if bottlerocket_node_spec.state != bottlerocket_node_status.current_state {
+                log::info!("Detected drift between spec state and current state. Requesting node to take action: {:?}.", &bottlerocket_node_spec.state);
+
+                match bottlerocket_node_spec.state {
+                    BottlerocketNodeState::WaitingForUpdate => {
+                        log::info!("Ready to finish monitoring and start update process")
+                    }
+                    BottlerocketNodeState::PreparingToUpdate => {
+                        log::info!("Preparing update");
+                        prepare().await.context(error::UpdateActions {
+                            action: "Prepare".to_string(),
+                        })?;
+
+                        // TODO: This function needs to use the k8s drain API to remove any pods from the host,
+                        // and then we need to wait until the host is successfully drained before transitioning
+                        // to the next state.
+                    }
+                    BottlerocketNodeState::PerformingUpdate => {
+                        log::info!("Performing update");
+                        update().await.context(error::UpdateActions {
+                            action: "Perform".to_string(),
+                        })?;
+                    }
+                    BottlerocketNodeState::RebootingToUpdate => {
+                        log::info!("Rebooting node to complete update");
+
+                        if ensure_reboot_happened(&bottlerocket_node_status)? {
+                            // previous execution `reboot` exited loop and did not update the custom resource
+                            // associated with this node. When re-enter loop and finished reboot,
+                            // try to update the custom resource.
+                            let update_node_status = self
+                                .gather_system_metadata(bottlerocket_node_spec.state.clone())
+                                .await?;
+                            self.update_metadata_custom_resource(update_node_status)
+                                .await?;
+                        } else {
+                            boot_update().await.context(error::UpdateActions {
+                                action: "Reboot".to_string(),
+                            })?;
+                        }
+                    }
+                    BottlerocketNodeState::MonitoringUpdate => {
+                        log::info!("Monitoring node's healthy condtion");
+                        // TODO: we need add some critierias here by which we decide to transition
+                        // from MonitoringUpdate to WaitingForUpdate.
+                    }
+                }
+            } else {
+                log::info!("Did not detect action demand.");
+            }
+
+            // update the custom resource associated with this node
+            let update_node_status = self
+                .gather_system_metadata(bottlerocket_node_spec.state.clone())
+                .await?;
+            self.update_metadata_custom_resource(update_node_status)
+                .await?;
+
+            log::debug!(
+                "Agent loop completed. Sleeping for {:?}.",
+                AGENT_SLEEP_DURATION
+            );
+            sleep(AGENT_SLEEP_DURATION).await;
+        }
+    }
+}
+
+// ensure if the nodes just rebooted and re-enter the loop.
+fn ensure_reboot_happened(status: &BottlerocketNodeStatus) -> Result<bool> {
+    // criteria that make sure node just rebooted
+    // 1. Check if system last reboot time is same as or around the timestamp that agent tried to reboot the system
+    // 2. Check current system version if it's same as the version we wanted to update to. (latest version)
+
+    let output = Command::new("uptime")
+        .args(["-s"])
+        .output()
+        .context(error::GetUptime)?;
+
+    // convert output string to naivedatetime
+    let uptime = String::from_utf8_lossy(&output.stdout).to_string();
+    let reboot_time = NaiveDateTime::parse_from_str(&uptime.trim_end(), "%Y-%m-%d %H:%M:%S")
+        .context(error::ConvertStringToDatetime { uptime })?;
+
+    // TODO: compare uptime to the timestamp that we're issuing at custom resource
+    // currently we compare it to current time, if the reboot happened in the past 24 hours, we
+    // assume the system just rebooted.
+    let now = chrono::offset::Utc::now().naive_local();
+    let duration_since_reboot = now.signed_duration_since(reboot_time).num_hours();
+
+    Ok(duration_since_reboot <= 1 && status.available_versions[0] == status.current_version)
+}
+
+fn get_server_domain() -> String {
+    format!(
+        "{}.{}.svc.cluster.local:{}",
+        APISERVER_SERVICE_NAME, NAMESPACE, APISERVER_SERVICE_PORT
+    )
+}
--- a/agent/src/apiclient.rs
+++ b/agent/src/apiclient.rs
@@ -1,0 +1,351 @@
+/*!
+apiclient is a client for interacting with the Bottlerocket Update API.
+Brupop volume mounts apiclient binary into the agent container.
+
+Bottlerocket Update API: https://github.com/bottlerocket-os/bottlerocket/tree/develop/sources/updater
+Bottlerocket apiclient: https://github.com/bottlerocket-os/bottlerocket/tree/develop/sources/api/apiclient
+*/
+
+use serde::Deserialize;
+use snafu::{ensure, ResultExt};
+use std::process::{Command, Output};
+use tokio::time::{sleep, Duration};
+
+const API_CLIENT_BIN: &str = "apiclient";
+const UPDATES_STATUS_URI: &str = "/updates/status";
+const OS_URI: &str = "/os";
+const REFRESH_UPDATES_URI: &str = "/actions/refresh-updates";
+const PREPARE_UPDATES_URI: &str = "/actions/prepare-update";
+const ACTIVATE_UPDATES_URI: &str = "/actions/activate-update";
+const REBOOT_URI: &str = "/actions/reboot";
+const MAX_ATTEMPTS: i8 = 5;
+const UPDATE_API_SLEEP_DURATION: Duration = Duration::from_millis(10000);
+const UPDATE_API_BUSY_STATUSCODE: &str = "423";
+
+/// The module-wide result type.
+pub type Result<T> = std::result::Result<T, apiclient_error::Error>;
+
+/// UpdateState represents four states during system update process
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize)]
+pub enum UpdateState {
+    // Idle: Unknow
+    Idle,
+    // Available: available versions system can update to
+    Available,
+    // Staged: processing system update (refresh, prepare, activate)
+    Staged,
+    // Ready: successfully complete update commands, ready to reboot
+    Ready,
+}
+
+/// UpdateCommand represents three commands to update system
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum UpdateCommand {
+    // Refresh: refresh the list of available updates
+    Refresh,
+    // Prepare: request that the update be downloaded and applied to disk
+    Prepare,
+    // Activate: proceed to "activate" the update
+    Activate,
+}
+
+/// CommandStatus represents three status after running update command
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize)]
+enum CommandStatus {
+    Success,
+    Failed,
+    Unknown,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct StagedImage {
+    image: Option<UpdateImage>,
+    next_to_boot: bool,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct CommandResult {
+    cmd_type: UpdateCommand,
+    cmd_status: CommandStatus,
+    timestamp: String,
+    exit_status: u32,
+    stderr: String,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct UpdateImage {
+    arch: String,
+    version: String,
+    variant: String,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct OsInfo {
+    pub version_id: String,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct UpdateStatus {
+    update_state: UpdateState,
+    available_updates: Vec<String>,
+    chosen_update: Option<UpdateImage>,
+    active_partition: Option<StagedImage>,
+    staging_partition: Option<StagedImage>,
+    most_recent_command: CommandResult,
+}
+
+fn get_raw_args(mut args: Vec<String>) -> Vec<String> {
+    let mut subcommand_args = vec!["raw".to_string(), "-u".to_string()];
+    subcommand_args.append(&mut args);
+
+    return subcommand_args;
+}
+
+/// Extract error statuscode from stderr string
+/// Error Example:
+/// "Failed POST request to '/actions/refresh-updates': Status 423 when POSTing /actions/refresh-updates: Update lock held\n"
+fn extract_status_code_from_error(error: &str) -> &str {
+    let error_content_split_by_status: Vec<&str> = error.split("Status").collect();
+    let error_content_split_by_whitespace: Vec<&str> = error_content_split_by_status[1]
+        .split_whitespace()
+        .collect();
+    let error_statuscode = error_content_split_by_whitespace[0];
+
+    error_statuscode
+}
+
+/// Apiclient binary has been volume mounted into the agent container, so agent is able to
+/// invoke `/bin apiclient` to interact with the Bottlerocket Update API.
+/// This function helps to invoke apiclient raw command.
+async fn invoke_apiclient(args: Vec<String>) -> Result<Output> {
+    let mut attempts: i8 = 0;
+    // Retry up to 5 times in case the Update API is busy; Waiting 10 seconds between each attempt.
+    while attempts < MAX_ATTEMPTS {
+        let output = Command::new(API_CLIENT_BIN)
+            .args(&args)
+            .output()
+            .context(apiclient_error::ApiClientRawCommand { args: args.clone() })?;
+
+        if output.status.success() {
+            return Ok(output);
+        }
+        let error_content = String::from_utf8_lossy(&output.stderr).to_string();
+        let error_statuscode = extract_status_code_from_error(&error_content);
+
+        match error_statuscode {
+            UPDATE_API_BUSY_STATUSCODE => {
+                log::info!(
+                    "API server busy, retrying in {:?} seconds ...",
+                    UPDATE_API_SLEEP_DURATION
+                );
+                // Retry after ten seconds if we get a 423 Locked response (update API busy)
+                sleep(UPDATE_API_SLEEP_DURATION).await;
+                attempts += 1;
+            }
+            _ => {
+                // API response was a non-transient error, bail out
+                return apiclient_error::BadHttpResponse {
+                    statuscode: error_statuscode,
+                }
+                .fail();
+            }
+        };
+    }
+    // Update API is currently unavailable, bail out
+    Err(apiclient_error::Error::UpdateApiUnavailable { args })
+}
+
+pub async fn get_update_status() -> Result<UpdateStatus> {
+    // Refresh list of updates and check if there are any available
+    refresh_updates().await?;
+
+    let raw_args = vec![UPDATES_STATUS_URI.to_string()];
+
+    let update_status_output = invoke_apiclient(get_raw_args(raw_args)).await?;
+
+    let update_status_string = String::from_utf8_lossy(&update_status_output.stdout).to_string();
+    let update_status: UpdateStatus = serde_json::from_str(&update_status_string)
+        .context(apiclient_error::UpdateStatusContent)?;
+
+    Ok(update_status)
+}
+
+/// extract OS info from the local system. For now this is just the version id.
+pub async fn get_os_info() -> Result<OsInfo> {
+    let raw_args = vec![OS_URI.to_string()];
+
+    let os_info_output = invoke_apiclient(get_raw_args(raw_args)).await?;
+
+    let os_info_content_string = String::from_utf8_lossy(&os_info_output.stdout).to_string();
+    let os_info: OsInfo =
+        serde_json::from_str(&os_info_content_string).context(apiclient_error::OsContent)?;
+
+    Ok(os_info)
+}
+
+pub async fn refresh_updates() -> Result<Output> {
+    let raw_args = vec![
+        REFRESH_UPDATES_URI.to_string(),
+        "-m".to_string(),
+        "POST".to_string(),
+    ];
+
+    Ok(invoke_apiclient(get_raw_args(raw_args)).await?)
+}
+
+pub async fn prepare_update() -> Result<()> {
+    let raw_args = vec![
+        PREPARE_UPDATES_URI.to_string(),
+        "-m".to_string(),
+        "POST".to_string(),
+    ];
+
+    invoke_apiclient(get_raw_args(raw_args)).await?;
+
+    Ok(())
+}
+
+pub async fn activate_update() -> Result<()> {
+    let raw_args = vec![
+        ACTIVATE_UPDATES_URI.to_string(),
+        "-m".to_string(),
+        "POST".to_string(),
+    ];
+
+    invoke_apiclient(get_raw_args(raw_args)).await?;
+
+    Ok(())
+}
+
+pub async fn reboot() -> Result<Output> {
+    let raw_args = vec![REBOOT_URI.to_string(), "-m".to_string(), "POST".to_string()];
+
+    Ok(invoke_apiclient(get_raw_args(raw_args)).await?)
+}
+
+// List all available versions which current Bottlerocket OS can update to.
+pub async fn list_available() -> Result<Vec<String>> {
+    let update_status = get_update_status().await?;
+
+    // Raise error if failed to refresh update or update acton performed out of band
+    ensure!(
+        update_status.most_recent_command.cmd_type == UpdateCommand::Refresh
+            && update_status.most_recent_command.cmd_status == CommandStatus::Success,
+        apiclient_error::RefreshUpdate
+    );
+
+    Ok(update_status.available_updates)
+}
+
+pub async fn prepare() -> Result<()> {
+    let update_status = get_update_status().await?;
+
+    ensure!(
+        update_status.update_state == UpdateState::Available
+            || update_status.update_state == UpdateState::Staged,
+        apiclient_error::UpdateStage {
+            expect_state: "Available or Staged".to_string(),
+            update_state: update_status.update_state,
+        },
+    );
+
+    // Download the update and apply it to the inactive partition
+    prepare_update().await?;
+
+    // Raise error if failed to prepare update or update action performed out of band
+    let recent_command = get_update_status().await?.most_recent_command;
+    ensure!(
+        recent_command.cmd_type == UpdateCommand::Prepare
+            || recent_command.cmd_status == CommandStatus::Success,
+        apiclient_error::PrepareUpdate
+    );
+
+    Ok(())
+}
+
+pub async fn update() -> Result<()> {
+    let update_status = get_update_status().await?;
+
+    ensure!(
+        update_status.update_state == UpdateState::Staged,
+        apiclient_error::UpdateStage {
+            expect_state: "Staged".to_string(),
+            update_state: update_status.update_state,
+        }
+    );
+
+    // Activate the prepared update
+    activate_update().await?;
+
+    // Raise error if failed to activate update or update action performed out of band
+    let recent_command = get_update_status().await?.most_recent_command;
+
+    ensure!(
+        recent_command.cmd_type == UpdateCommand::Activate
+            || recent_command.cmd_status == CommandStatus::Success,
+        apiclient_error::Update
+    );
+
+    Ok(())
+}
+
+// Reboot the host into the activated update
+pub async fn boot_update() -> Result<Output> {
+    let update_status = get_update_status().await?;
+
+    ensure!(
+        update_status.update_state == UpdateState::Ready,
+        apiclient_error::UpdateStage {
+            expect_state: "Ready".to_string(),
+            update_state: update_status.update_state,
+        }
+    );
+
+    Ok(reboot().await?)
+}
+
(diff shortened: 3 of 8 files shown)


agent/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "agent"
+version = "0.1.0"
+edition = "2018"
+publish = false
+
```
cbgbt, line 7: It would be best to add a license line here.
```suggestion
license = "Apache-2.0 OR MIT"

```

agent/src/agentclient.rs
```diff
@@ -0,0 +1,176 @@
...
+    pub fn new(k8s_client: kube::client::Client) -> Self {
+        BrupopAgent { k8s_client }
+    }
+
+    pub async fn ensure_node_custom_resource_exists(&self) {
+        // Todo (tianhg)
+    }
+
+    async fn fetch_custom_resource_helper(&self) -> Result<Node> {
+        // get associated node name
```
cbgbt: Nit: This comment doesn't say much more than the line of code following it. It might be good to add more context about what the node name is, or what it's used for.

agent/src/agentclient.rs
```diff
@@ -0,0 +1,176 @@
...
+        let associated_node_name = format!("brn-{}", self.get_node_selector().await.node_name);
```
cbgbt: I wonder if we should add a function to `models` to compute the `brn` name from a `node` name, that way we can ensure we use the same method in the `apiserver` and `agent`?

agent/src/agentclient.rs
```diff
@@ -0,0 +1,176 @@
...
+
+        let bottlerocket_nodes: Api<Node> = Api::namespaced(self.k8s_client.clone(), NAMESPACE);
```
cbgbt: In kubernetes, it's best to avoid polling etcd if we can. One possible solution is to use a [reflector](https://github.com/kube-rs/kube-rs#reflectors) which uses the k8s `watch` APIs and caches changes locally. If we can get a reflector to specifically watch this node, that might be ideal.

I'm not 100% sure if it's better to have a WATCH per-node or to do polling, since it's possible WATCH scales poorly on number-of-WATCHes. It might be best for now to open an issue to investigate this further, and do our best to minimize polling in this PR.

agent/src/apiclient.rs
```diff
@@ -0,0 +1,290 @@
+use serde::{Deserialize, Serialize};
+use serde_json;
+use snafu::{ResultExt, Snafu};
+use std::process::{Command, Output};
+
+/// The crate-wide result type.
```
cbgbt: Nit:
```suggestion
/// The module-wide result type.
```

agent/src/apiclient.rs
```diff
@@ -0,0 +1,290 @@
+use serde::{Deserialize, Serialize};
```
cbgbt: It might be worth adding a module-level doc comment clarifying that this is a client for the Bottlerocket API, rather than the Brupop one. In retrospect, maybe naming Brupop's API the "apiserver" is a recipe for ambiguity problems. We could consider a change in name.

agent/src/agentclient.rs
cbgbt: Separately, I believe you want `Api<BottlerocketNode>` here, which will return our custom resource. `Api<Node>` will return the k8s node object.

agent/src/agentclient.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    pub async fn run(&mut self) -> Result<()> {
+        // A running agent has two responsibilities:
+        // - Gather metadata about the system and update the custom resource associated with this node
+        // - Determine if the spec on the system's custom resource demands the node take action. If so, begin taking that action.
+
+        loop {
+            // Gather metadata about the system
+            let node_custom_resource_spec = self.fetch_metadata_custom_resource_spec().await;
+            let node_custom_resource_status = self.fetch_metadata_custom_resource_status().await;
```
cbgbt: I think if you can fix `fetch_custom_resource_helper` to return a `BottlerocketNode`, you can probably do away with these two functions and just refer to the `.status` and `.spec` directly.

agent/src/agentclient.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    /// Requests the metadata 'spec' for the current BottlerocketNode from the Kubernetes custom resource associated
+    /// with this node.
+    async fn fetch_metadata_custom_resource_spec(&self) -> Result<BottlerocketNodeSpec> {
+        let associated_node_metadata = self.fetch_custom_resource_helper().await?;
+
+        Ok(BottlerocketNodeSpec {
+            state: associated_node_metadata.state,
+            version: associated_node_metadata.status.version,
```
cbgbt: You'll want to make sure this version comes from the `spec` rather than the status.

agent/src/agentclient.rs
```diff
@@ -0,0 +1,176 @@
...
+                    BottlerocketNodeState::PerformingUpdate => {
+                        log::info!("Performing update");
+                        update().context(error::PerformUpdate)?;
+                    }
+                    BottlerocketNodeState::RebootingToUpdate => {
+                        log::info!("Rebooting node to complete update");
+                        boot_update().context(error::BootUpdate)?;
+                    }
+                    BottlerocketNodeState::MonitoringUpdate => {
+                        log::info!("Monitoring node's healthy condtion");
```
cbgbt, line 255: A `TODO` for here is to add some criteria by which we decide to transition from `MonitoringUpdate` to `WaitingForUpdate`. I'd open an issue for this.

agent/src/agentclient.rs
```diff
@@ -0,0 +1,176 @@
...
+                    }
+                    BottlerocketNodeState::PreparingToUpdate => {
+                        log::info!("Preparing update");
+                        prepare().context(error::PrepareUpdate)?;
+                    }
+                    BottlerocketNodeState::PerformingUpdate => {
+                        log::info!("Performing update");
+                        update().context(error::PerformUpdate)?;
+                    }
+                    BottlerocketNodeState::RebootingToUpdate => {
```
cbgbt, line 236: This might be answered later in the PR, but I'm wondering if we have a way of knowing here if we have just rebooted and are re-entering the loop after a previous execution, or if we still need to reboot?

agent/src/apiclient.rs
```diff
@@ -0,0 +1,290 @@
...
+    #[snafu(display("Failed to deserialize Os info: '{}'", source))]
+    OsContent { source: serde_json::Error },
+
+    #[snafu(display("Failed to deserialize update status: '{}'", source))]
+    UpdateStatusContent { source: serde_json::Error },
+
+    #[snafu(display("Failed to deserialize Os info: '{}'", source))]
+    GetMostRecentCommand { source: serde_json::Error },
+
+    #[snafu(display("failed to refresh updates or update action performed out of band"))]
```
cbgbt: Is it possible to store a `source` error with these so that we can add them to the `display`? If there are multiple possible sources of error, you could make the type `Box<dyn std::error::Error>`, e.g. [defining the error](https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/0.2.0-apiserver/models/src/node.rs#L25) and then you have to use [this `map` invocation](https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/0.2.0-apiserver/models/src/node.rs#L247) when you call `context()`

agent/src/apiclient.rs
```diff
@@ -0,0 +1,290 @@
...
+         update_state, expect_state
+    ))]
+    UpdateStage {
+        expect_state: String,
+        update_state: UpdateState,
+    },
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize)]
+pub enum UpdateState {
```
cbgbt, line 30: Some docstrings on these enums would be really useful.

agent/src/apiclient.rs
```diff
@@ -0,0 +1,290 @@
...
+pub fn parse_raw_args(mut args: Vec<String>) -> Vec<String> {
+    let mut subcommand_args = Vec::new();
+    subcommand_args.push("raw".to_string());
+    subcommand_args.push("-u".to_string());
+    subcommand_args.append(&mut args);
+
+    return subcommand_args;
+}
+
+/// a helper function to invoke apiclient command
```
cbgbt: Nit: I'd add some details to this docstring about the fact that the API client is volume mounted into the agent container.

agent/src/apiclient.rs
```diff
@@ -0,0 +1,290 @@
...
+    let update_status_output = invoke_apiclient(parse_raw_args(raw_args))?;
+
+    let update_status_string = String::from_utf8_lossy(&update_status_output.stdout).to_string();
+    let update_status: UpdateStatus =
+        serde_json::from_str(&update_status_string).context(UpdateStatusContent)?;
+
+    Ok(update_status)
+}
+
+/// extract OS info like version id.
```
cbgbt: nit

```suggestion
/// extract OS info from the local system. For now this is just the version id.
```

agent/src/apiclient.rs
```diff
@@ -0,0 +1,290 @@
...
+pub struct UpdateStatus {
+    update_state: UpdateState,
+    available_updates: Vec<String>,
+    chosen_update: Option<UpdateImage>,
+    active_partition: Option<StagedImage>,
+    staging_partition: Option<StagedImage>,
+    most_recent_command: CommandResult,
+}
+
+pub fn parse_raw_args(mut args: Vec<String>) -> Vec<String> {
```
cbgbt: nit: Since this function isn't doing any parsing itself, it might be best to change the name.

agent/src/apiclient.rs
```diff
@@ -0,0 +1,290 @@
...
+    let os_info_content_string = String::from_utf8_lossy(&os_info_output.stdout).to_string();
+    let os_info_content_json: serde_json::Value =
+        serde_json::from_str(&os_info_content_string).context(OsContent)?;
+
+    Ok(OsInfo {
+        version_id: os_info_content_json["version_id"].to_string(),
+    })
+}
+
+pub fn get_chosen_update() -> Result<Option<UpdateImage>> {
```
cbgbt: I think it might be better design to add these as `getters` to the `UpdateStatus`, or alternatively change the privacy of `UpdateStatus` attributes to `pub` and expect callers to call `get_update_status()` themselves.

agent/src/apiclient.rs
```diff
@@ -0,0 +1,290 @@
...
+
+    Ok(())
+}
+
+// List all available versions which current Bottlerocket OS can update to.
+pub fn list_available() -> Result<Vec<String>> {
+    // Refresh list of updates and check if there are any available
+    refresh_updates()?;
+
+    let update_statues = get_update_status()?;
```
cbgbt: nit: typo

agent/src/apiclient.rs
```diff
@@ -0,0 +1,290 @@
...
+    Ok(())
+}
+
+pub fn reboot() -> Result<()> {
+    let mut raw_args = Vec::new();
+    raw_args.push("/actions/reboot".to_string());
+    raw_args.push("-m".to_string());
+    raw_args.push("POST".to_string());
+
+    invoke_apiclient(raw_args)?;
```
cbgbt: It's probably best to exit this program here, after issuing the reboot.

agent/src/apiclient.rs
```diff
@@ -0,0 +1,290 @@
...
+    if update_statues.most_recent_command.cmd_type != UpdateCommand::Refresh
+        && update_statues.most_recent_command.cmd_status != CommandStatus::Success
+    {
+        return RefreshUpdate{}.fail();
+    }
+
+    Ok(update_statues.available_updates)
+}
+pub fn prepare() -> Result<()> {
+    let update_statues = get_update_status()?;
```
cbgbt: nit: typo. `update_statues` -> `update_status`

agent/src/apiclient.rs
```diff
@@ -0,0 +1,290 @@
...
+
```
cbgbt, line 240: This function needs to use the k8s `drain` API to remove any pods from the host, and then we need to wait until the host is successfully drained before transitioning to the next state.

cbgbt: I'm realizing that this might need to be a feature that we add to `apiserver`, since I'm not sure if it's possible for us to provide a role to the `agent` which allows it to only drain itself...

For now let's do it here, but I'll create an issue to move this functionality.

agent/src/agentclient.rs
gthao313: That sounds a good idea which can keep consistent. I can implement here and move to models in future.

agent/src/agentclient.rs
gthao313: As discussed offline, I'll add a logic to determine if system just get update

agent/src/agentclient.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    pub async fn ensure_node_custom_resource_exists(&self) {
+        // Todo (tianhg)
+    }
+
+    async fn fetch_custom_resource(&self) -> Result<BottlerocketNode> {
+        // TODO Nit: This comment doesn't say much more than the line of code following it.
+        // It might be good to add more context about what the node name is, or what it's used for.
+        let associated_bottlerocketnode_name =
```
arnaldo2792, line 86: Nit: why not `associated_bottlerocket_node_name` to be consistent with `associated_node_name` in lines below?

agent/src/agentclient.rs
```diff
@@ -0,0 +1,173 @@
...
+        let associated_bottlerocketnode: BottlerocketNode = bottlerocket_nodes
+            .get(&associated_bottlerocketnode_name)
+            .context(error::FetchCustomResource)
+            .await?;
+
+        Ok(associated_bottlerocketnode)
+    }
+
+    async fn get_node_selector(&self) -> Result<BottlerocketNodeSelector> {
+        let associated_node_name = env::var("MY_NODE_NAME").context(error::GetNodeName)?;
```
arnaldo2792: Is this going to be the actual environment variable name that customers will set up? Or is that env variable coming from the kubernetes cluster in a weird way?

agent/src/agentclient.rs
```diff
@@ -0,0 +1,173 @@
...
+        })
+    }
+
+    /// update the custom resource associated with this node
+    async fn update_metadata_custom_resource(
+        &self,
+        _current_metadata: BottlerocketNodeStatus,
+    ) -> Result<()> {
+        let client = reqwest::Client::new();
+        let currernt_node_selector = self.get_node_selector().await?;
```
arnaldo2792: Nit: typo
```suggestion
        let current_node_selector = self.get_node_selector().await?;
```

agent/src/agentclient.rs
```diff
@@ -0,0 +1,173 @@
...
+        Ok(BottlerocketNodeSelector {
+            node_name: associated_node_name,
+            node_uid: associated_node_uid,
+        })
+    }
+
+    async fn get_server_domain(&self) -> &str {
+        // Todo (Tianhg)
+
+        "52.24.238.246:30000"
```
arnaldo2792: Might be worth not having this hard coded value

agent/src/agentclient.rs
```diff
@@ -0,0 +1,173 @@
...
+        // A running agent has two responsibilities:
+        // - Gather metadata about the system and update the custom resource associated with this node
+        // - Determine if the spec on the system's custom resource demands the node take action. If so, begin taking that action.
+
+        loop {
+            // Requests metadata 'status' and 'spec' for the current BottlerocketNode
+            let bottlerocket_node_status = match self.fetch_custom_resource().await?.status {
+                Some(status) => status,
+                None => return error::MissBottlerocketNodeStatus {}.fail(),
+            };
```
arnaldo2792: If you include `snafu::OptionExt`, you can do:

```suggestion
            let bottlerocket_node_status = self
              .fetch_customer_resource()
              .await?
              .status
              .context(error::MissBottlerocketNodeStatus)?;
```

agent/src/apiclient.rs
```diff
@@ -0,0 +1,296 @@
...
+    most_recent_command: CommandResult,
+}
+
+pub fn get_raw_args(mut args: Vec<String>) -> Vec<String> {
+    let mut subcommand_args = Vec::new();
+    subcommand_args.push("raw".to_string());
+    subcommand_args.push("-u".to_string());
+    subcommand_args.append(&mut args);
+
+    return subcommand_args;
```
arnaldo2792:
```suggestion
    let mut subcommand_args = vec!["raw", "-u"];
    subcommand_args.append(&mut args);

    return subcommand_args;
```

agent/src/apiclient.rs
```diff
@@ -0,0 +1,296 @@
...
+}
+
+/// TODO
+/// a helper function to invoke apiclient command
+/// Nit: I'd add some details to this docstring about the fact that the API client is volume mounted into the agent container.
+pub fn invoke_apiclient(args: Vec<String>) -> Result<Output> {
+    let output = Command::new("/bin/apiclient")
```
arnaldo2792: Nit: `"/bin/apiclient"` should be a constant

agent/src/apiclient.rs
```diff
@@ -0,0 +1,296 @@
...
+        .args(&args)
+        .output()
+        .context(ApiClientCommand { args })?;
+
+    Ok(output)
+}
+
+pub fn get_update_status() -> Result<UpdateStatus> {
+    let mut raw_args = Vec::new();
+    raw_args.push("/updates/status".to_string());
```
arnaldo2792: Nit: `"/updates/status"`, should be a constant, and then:

```suggestion
    let mut raw_args = vec![UPDATES_STATUS_URI.to_string()];
```

agent/src/apiclient.rs
```diff
@@ -0,0 +1,296 @@
...
+    let update_status: UpdateStatus =
+        serde_json::from_str(&update_status_string).context(UpdateStatusContent)?;
+
+    Ok(update_status)
+}
+
+/// extract OS info from the local system. For now this is just the version id.
+pub fn get_os_info() -> Result<OsInfo> {
+    let mut raw_args = Vec::new();
+    raw_args.push("/os".to_string());
```
arnaldo2792: Same as my comment above

agent/src/apiclient.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    let os_info_output = invoke_apiclient(get_raw_args(raw_args))?;
+
+    let os_info_content_string = String::from_utf8_lossy(&os_info_output.stdout).to_string();
+    let os_info_content_json: serde_json::Value =
+        serde_json::from_str(&os_info_content_string).context(OsContent)?;
+
+    Ok(OsInfo {
+        version_id: os_info_content_json["version_id"].to_string(),
+    })
```
arnaldo2792: There should be a way to deserialize the values that you want from the Json string, into the `OsInfo` struct, with serde. Maybe worth looking at how to deserialize with serde:

https://serde.rs/derive.html

agent/src/apiclient.rs
```diff
@@ -0,0 +1,296 @@
...
+pub fn get_chosen_update() -> Result<Option<UpdateImage>> {
+    let update_status = get_update_status()?;
+
+    Ok(update_status.chosen_update)
+}
+
+pub fn get_most_recent_command() -> Result<CommandResult> {
+    let update_status = get_update_status()?;
+
+    Ok(update_status.most_recent_command)
```
arnaldo2792: Maybe:

```suggestion
    Ok(get_update_status()?.most_recent_command)
```

agent/src/apiclient.rs
```diff
@@ -0,0 +1,296 @@
...
+}
+
+pub fn refresh_updates() -> Result<()> {
+    let mut raw_args = Vec::new();
+    raw_args.push("/actions/refresh-updates".to_string());
+    raw_args.push("-m".to_string());
+    raw_args.push("POST".to_string());
```
arnaldo2792: Same as my previous comments related to vecs and constants

agent/src/apiclient.rs
```diff
@@ -0,0 +1,296 @@
...
+    refresh_updates()?;
+
+    let update_status = get_update_status()?;
+
+    // Raise error if failed to refresh update or update acton performed out of band
+    if update_status.most_recent_command.cmd_type != UpdateCommand::Refresh
+        && update_status.most_recent_command.cmd_status != CommandStatus::Success
+    {
+        return RefreshUpdate {}.fail();
+    }
```
arnaldo2792: You can include `snafu::ensure` and do

```suggestion
   ensure!(
      update_status.most_recent_command.cmd_type != UpdateCommand::Refresh 
      && update_status.most_recent_command.cmd_status != CommandStatus::Success, 
      RefreshUpdate
   );
  
```

agent/src/apiclient.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    if update_status.update_state != UpdateState::Available
+        && update_status.update_state != UpdateState::Staged
+    {
+        return UpdateStage {
+            expect_state: "Available or Staged".to_string(),
+            update_state: update_status.update_state,
+        }
+        .fail();
+    }
```
arnaldo2792: Same as my snafu comment with `ensure!`

agent/src/apiclient.rs
```diff
@@ -0,0 +1,296 @@
...
+    // Download the update and apply it to the inactive partition
+    prepare_update()?;
+
+    // Raise error if failed to prepare update or update action performed out of band
+    let recent_command = get_most_recent_command()?;
+    if recent_command.cmd_type != UpdateCommand::Prepare
+        || recent_command.cmd_status != CommandStatus::Success
+    {
+        return PrepareUpdate {}.fail();
+    }
```
arnaldo2792: Same as `ensure!` with snafu, and for other cases down this line

agent/src/error.rs
```diff
@@ -0,0 +1,55 @@
...
+    UpdateBottlerocketNodeResource { source: reqwest::Error },
+
+    #[snafu(display("Unable to take action `prepare`: {}", source))]
+    PrepareUpdate { source: ApiClientError },
+
+    #[snafu(display("Unable to take action `update`: {}", source))]
+    PerformUpdate { source: ApiClientError },
+
+    #[snafu(display("Unable to take action `reboot`: {}", source))]
+    BootUpdate { source: ApiClientError },
```
arnaldo2792: Maybe this could be the same error, and just receive `what` as a parameter:

```suggestion
    #[snafu(display("Unable to take action '{}': {}", action, source))]
    # Maybe a better name, this is just a suggestion
    TakeAction { action: String, source: ApiClientError },
```

agent/src/error.rs
```diff
@@ -0,0 +1,55 @@
...
+    FetchNode { source: kube::Error },
+
+    #[snafu(display("Uable to get Node uid because of missing"))]
+    MissNodeUid {},
+
+    #[snafu(display("Uable to get Bottlerocket node 'status' because of missing"))]
+    MissBottlerocketNodeStatus,
+
+    #[snafu(display("Uable to get Bottlerocket node 'spec' because of missing"))]
+    MissBottlerocketNodeSpec,
```
arnaldo2792: Nit: typos, and you didn't specify what is actually missing?

arnaldo2792 (changes requested):

agent/src/agentclient.rs
gthao313: There are two kinds, one is `BottlerocketNode` which is created by Burpop (custom resource), and another one is the [Node](https://docs.rs/k8s-openapi/0.13.1/k8s_openapi/api/core/v1/struct.Node.html) which we are familiar with. Therefore, I name them `associated_bottlerocket_node` and `associated_node`

agent/src/agentclient.rs
gthao313: Yeah. It's a actual env variable name that set up by ourselves on `burpop-agent.yaml` file
```
    env:
        - name: MY_NODE_NAME
          valueFrom:
            fieldRef:
              fieldPath: spec.nodeName
```

gthao313:
> I think it might be better design to add these as getters to the UpdateStatus, or alternatively change the privacy of UpdateStatus attributes to pub and expect callers to call get_update_status() themselves.

I was thinking about designing to add getters, but I do not think we should do this because `get_update_status` gonna be called so many time during update process. If we add this as a getter if we need initialize the object so many times. Please correct me if I understand and think wrong. thanks @cbgbt  : )
```
struct UpdateStatus {
}

impl UpdateStatus {
   pub new() {
      get_update_status()
}
   pub get_update_status() {}
}
```

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+
+use crate::apiclient::{boot_update, get_os_info, list_available, prepare, update};
+use crate::error::{self, Result};
+use apiserver::api::{CreateBottlerocketNodeRequest, UpdateBottlerocketNodeRequest};
+use models::constants::{APISERVER_SERVICE_NAME, APISERVER_SERVICE_PORT, NAMESPACE};
+use models::node::{
+    BottlerocketNode, BottlerocketNodeSelector, BottlerocketNodeState, BottlerocketNodeStatus,
+};
+
+const AGENT_SLEEP_DURATION: Duration = Duration::from_millis(5000);
```
cbgbt: nit: I think there's a `Duration::from_secs`.

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+}
+
+impl BrupopAgent {
+    pub fn new(k8s_client: kube::client::Client) -> Self {
+        BrupopAgent { k8s_client }
+    }
+
+    pub async fn ensure_node_custom_resource_exists(&self) -> Result<bool> {
+        let associated_bottlerocketnode_name =
+            format!("brn-{}", self.get_node_selector().await?.node_name);
```
cbgbt: Nit: Might be best to move the name computation to the `models` crate.

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+        let bottlerocket_nodes: Api<BottlerocketNode> =
+            Api::namespaced(self.k8s_client.clone(), NAMESPACE);
+
+        let bottlerocket_nodes_objectlist = bottlerocket_nodes
+            .list(&ListParams::default())
+            .await
+            .context(error::ConvertToObjectList {})?;
```
cbgbt: I believe that the [`Api.get`](https://docs.rs/kube/0.63.1/kube/api/struct.Api.html#method.get) operation should work here, and is probably cheaper. I'm not sure what the error type is if the object doesn't exist though, possibly one of [these](https://docs.rs/kube/0.63.1/kube/error/enum.DiscoveryError.html)?

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+pub struct BrupopAgent {
+    k8s_client: kube::client::Client,
+}
+
+impl BrupopAgent {
+    pub fn new(k8s_client: kube::client::Client) -> Self {
+        BrupopAgent { k8s_client }
+    }
+
+    pub async fn ensure_node_custom_resource_exists(&self) -> Result<bool> {
```
cbgbt: Nit: I would change `ensure_` to `check_`. Usually "ensure" implies that this function will create the resource if it doesn't exist.

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+                    == &associated_bottlerocketnode_name
+            })
+        {
+            return Ok(true);
+        };
+
+        Ok(false)
+    }
+
+    pub async fn ensure_custom_resource_status_exists(&self) -> Result<bool> {
```
cbgbt: Nit: Same as above. I'd change "ensure" to "check"

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+
+        Ok(BottlerocketNodeStatus {
+            current_version: current_version.version_id.clone(),
+            available_versions: update_versions,
+            current_state: state,
+        })
+    }
+
+    /// create the custom resource associated with this node
+    pub async fn create_metadata_custom_resource(&self) -> Result<()> {
```
cbgbt: I think it would be a good improvement to add retries with exponential backoff to this. Feel free to cut an issue and followup with this to avoid making this PR larger.

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+                node_name: current_node_selector.node_name,
+            })?;
+
+        log::info!("{}", response.text().await.unwrap());
+
+        Ok(())
+    }
+
+    /// update the custom resource associated with this node
+    async fn update_metadata_custom_resource(
```
cbgbt, line 151: Same feedback here in regards to retries.

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+            Api::namespaced(self.k8s_client.clone(), NAMESPACE);
+        let associated_bottlerocketnode: BottlerocketNode = bottlerocket_nodes
+            .get(&associated_bottlerocketnode_name)
+            .await
+            .context(error::FetchCustomResource)?;
+
+        Ok(associated_bottlerocketnode)
+    }
+
+    async fn get_node_selector(&self) -> Result<BottlerocketNodeSelector> {
```
cbgbt: Maybe take an `&mut` reference to `self` and cache the result to this call? We call `get_node_selector()` a fair bit later.

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+    async fn update_metadata_custom_resource(
+        &self,
+        _current_metadata: BottlerocketNodeStatus,
+    ) -> Result<()> {
+        let client = reqwest::Client::new();
+        let current_node_selector = self.get_node_selector().await?;
+        let server_domain = format!(
+            "{}.{}.svc.cluster.local:{}",
+            APISERVER_SERVICE_NAME, NAMESPACE, APISERVER_SERVICE_PORT
+        );
```
cbgbt: Might be good to refactor this out into a function.

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+                node_name: current_node_selector.node_name,
+            })?;
+
+        log::info!("{}", response.text().await.unwrap());
+
+        Ok(())
+    }
+
+    /// initialize bottlerocketnode (custom resource) `status` when create new bottlerocketnode
+    pub async fn initialize_metadata_custom_reource(&self) -> Result<()> {
```
cbgbt: Nit: typo in name (`reource`)

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    pub async fn run(&mut self) -> Result<()> {
+        // A running agent has two responsibilities:
+        // - Gather metadata about the system and update the custom resource associated with this node
+        // - Determine if the spec on the system's custom resource demands the node take action. If so, begin taking that action.
+
+        loop {
+            // Requests metadata 'status' and 'spec' for the current BottlerocketNode
+            let bottlerocket_node_status = self
+                .fetch_custom_resource()
```
cbgbt: Since `fetch_custom_resource()` is calling the kubernetes API, we should probably call it only just once to get both the status and spec.

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+            //Handle the situation that system just rebooted and re-enter the loop after a previous execution
+
+            // Determine if the spec on the system's custom resource demands the node take action. If so, begin taking that action.
+            if bottlerocket_node_spec.state != bottlerocket_node_status.current_state {
+                log::info!("Detected drift between spec state and current state. Requesting node to take action: {:?}.", &bottlerocket_node_spec.state);
+
+                match bottlerocket_node_spec.state {
+                    BottlerocketNodeState::WaitingForUpdate => {
+                        log::info!("Waiting for signal to update")
+                    }
```
cbgbt: If our current state is in `MonitoringForUpdate`, then this could be a signal that it's time to finish monitoring.

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+        .context(error::ConvertStringToDatetime { uptime })?;
+
+    // TODO: compare uptime to the timestamp that we're issuing at custom resource
+    // currently we compare it to current time, if the reboot happened in the past 24 hours, we
+    // assume the system just rebooted.
+    let now = chrono::offset::Utc::now().naive_local();
+    let duration_since_reboot = now.signed_duration_since(reboot_time).num_days();
+
+    if duration_since_reboot == 0 && status.available_versions[0] == status.current_version {
+        return Ok(true);
```
cbgbt: I think this is a clever first cut. I would make the duration check a bit tighter (e.g. within the last hour), but otherwise this looks good.

cbgbt (commented):
Some initial comments, I want to take one more closer look at `apiclient.rs` as well.

agent/src/apiclient.rs
```diff
@@ -0,0 +1,353 @@
...
+/// Idle: Unknow
+/// Available: available versions system can update to
+/// Staged: processing system update (refresh, prepare, activate)
+/// Ready: successfully complete update commands, ready to reboot.
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize)]
+pub enum UpdateState {
+    Idle,
+    Available,
+    Staged,
+    Ready,
```
cbgbt: Nit: As a tip, in Rust you can add docstrings directly to enum variants, like so:

```
pub enum UpdateState {
    /// No action required for updates
    Idle,
    /// Available versions system can update to
    Available,
    ...
}
```

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+            .iter()
+            .any(|bottlerocket_node_name| {
+                bottlerocket_node_name.metadata.name.as_ref().unwrap()
+                    == &associated_bottlerocketnode_name
+            })
+        {
+            return Ok(true);
+        };
+
+        Ok(false)
```
arnaldo2792: The `unwrap` call will panic if `name` is an `Err<>`, maybe your can try this, but you need to import `ResultExt` from snafu:

```suggestion
        let exists = bottlerocket_nodes_objectlist
            .iter()
            .any(|bottlerocket_node_name| {
                bottlerocket_node_name.metadata.name.as_ref().context(error::MissingName)?
                    == &associated_bottlerocketnode_name
            })

        Ok(exists)
```

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    pub async fn ensure_custom_resource_status_exists(&self) -> Result<bool> {
+        match self.fetch_custom_resource().await?.status {
+            Some(_status) => return Ok(true),
+            None => return Ok(false),
+        };
```
arnaldo2792: You can change this to:

```suggestion
        self.fetch_custom_resource().await?.status.is_some()
```

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+        let nodes: Api<Node> = Api::all(self.k8s_client.clone());
+        let associated_node: Node = nodes
+            .get(&associated_node_name)
+            .await
+            .context(error::FetchNode)?;
+
+        let associated_node_uid = match associated_node.metadata.uid {
+            Some(uid) => uid,
+            None => return error::MissNodeUid {}.fail(),
+        };
```
arnaldo2792:
```suggestion
        # Also, I renamed you error :D
        let associated_node_uid = match associated_node.metadata.uid.context(error::MissingNodeUid)?;
```

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+                &server_domain
+            ))
+            .json(&node_req)
+            .send()
+            .await
+            .context(error::UpdateBottlerocketNodeResource {
+                node_name: current_node_selector.node_name,
+            })?;
+
+        log::info!("{}", response.text().await.unwrap());
```
arnaldo2792: Please watch out all the unwraps that you used and replace them with `method.context(error::MyCoolError)?;`

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+                .gather_system_metadata(bottlerocket_node_spec.state.clone())
+                .await?;
+            self.update_metadata_custom_resource(update_node_status)
+                .await?;
+
+            log::debug!(
+                "Agent loop completed. Sleeping for {:?}.",
+                AGENT_SLEEP_DURATION
+            );
+            sleep(AGENT_SLEEP_DURATION).await;
```
arnaldo2792, line 275: Don't you have to add `?` at the end? I'm not sure, I'm just asking :smile:

agent/src/agentclient.rs
```diff
@@ -0,0 +1,306 @@
...
+    // currently we compare it to current time, if the reboot happened in the past 24 hours, we
+    // assume the system just rebooted.
+    let now = chrono::offset::Utc::now().naive_local();
+    let duration_since_reboot = now.signed_duration_since(reboot_time).num_days();
+
+    if duration_since_reboot == 0 && status.available_versions[0] == status.current_version {
+        return Ok(true);
+    }
+
+    Ok(false)
```
arnaldo2792:
```suggestion
    Ok(duration_since_reboot == 0 && status.available_versions[0] == status.current_version)
```

agent/src/apiclient.rs
```diff
@@ -0,0 +1,353 @@
...
+const UPDATE_API_SLEEP_DURATION: Duration = Duration::from_millis(10000);
+const UPDATE_API_BUSY_STATUSCODE: &str = "423";
+
+/// The module-wide result type.
+pub type Result<T> = std::result::Result<T, ApiClientError>;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility = "pub")]
+
+pub enum ApiClientError {
```
arnaldo2792: Maybe a painful refactor, but consider moving errors at the end, and follow the "news paper" style, where the most important parts of your PR are at the top. Also consider prefixing the errors with `error`:

```suggestion
mod error {
    pub(super) enum Error {
```

And then you can use the errors as: `error::Update`

But before all that, why aren't these errors part of the `error.rs` file?

arnaldo2792 (changes requested):

agent/src/agentclient.rs
cbgbt: The name should always exist, assuming the object is being returned by the k8s API. Interestingly, `kube-rs` provides the [`ResourceExt` trait](https://docs.rs/kube/0.60.0/kube/trait.ResourceExt.html#tymethod.name) that you can import to get a method which gets the name, and panics if it doesn't exist, which I think is pretty indicative that the name particularly is rarely missing.

That said, I do like to hedge our bets -- best to avoid panicing if we can!

agent/src/agentclient.rs
gthao313: We do not need it 😄

agent/src/agentclient.rs
gthao313: I agree with we should avoid panic if we can, so I would try with arnaldo's suggestion. thanks!

agent/src/agentclient.rs
gthao313: Already created [Issue](https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/99) to follow up this comment

agent/src/agentclient.rs
gthao313: Already created [Issue](https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/99) to follow up this comment

agent/src/agentclient.rs
gthao313: I update this to following format because of error `cannot use the ? operator in a closure that returns bool`
```
        for bottlerocket_node in bottlerocket_nodes_objectlist.iter() {
            let bottlerocket_node_name = bottlerocket_node
                .metadata
                .name
                .as_ref()
                .context(error::MissingNodeName)?;
            if bottlerocket_node_name == &associated_bottlerocketnode_name {
                return Ok(true);
            }
        }
```

agent/src/apiclient.rs
```diff
@@ -0,0 +1,359 @@
...
+        apiclient_error::UpdateStage {
+            expect_state: "Ready".to_string(),
+            update_state: update_status.update_state,
+        }
+    );
+
+    Ok(reboot().await?)
+}
+
+pub mod apiclient_error {
```
gthao313, line 308: I move all errors to a mod but not move all errors to a same file. The reason isI think  apiclient is a bottlerocket package and we mount the binary to burpop, so I want to demonstrate those errors are belong to apiclient not burpop.

gthao313:
Push above address the changes to Sean and Arnaldo 's comments.

agent/src/apiclient.rs
```diff
@@ -0,0 +1,359 @@
...
+    let mut subcommand_args = vec!["raw".to_string(), "-u".to_string()];
+    subcommand_args.append(&mut args);
+
+    return subcommand_args;
+}
+
+/// Apiclient binary has been volume mounted into the agent container, so agent is able to
+/// invoke `/bin apiclient` to interact with the Bottlerocket Update API.
+/// This function helps to invoke apiclient command.
+async fn invoke_apiclient(args: Vec<String>) -> Result<Output> {
```
cbgbt, line 121: Nit: I think this function may only work for `raw` commands to the API, as the error detection doesn't work otherwise. Might be best to add that to the docstring.

cbgbt (approved):
Thanks for addressing so many comments! I'm pretty sure anything that we haven't followed up on here has been documented as an issue. Nice work. :ship:

cbgbt:
> > I think it might be better design to add these as getters to the UpdateStatus, or alternatively change the privacy of UpdateStatus attributes to pub and expect callers to call get_update_status() themselves.
> 
> I was thinking about designing to add getters, but I do not think we should do this because `get_update_status` gonna be called so many time during update process. If we add this as a getter if we need initialize the object so many times. Please correct me if I understand and think wrong. thanks @cbgbt : )

I don't think I said this well, but I think `get_chosen_update()` and `get_most_recent_command()` should be removed, and users who want to get this information should instead just call ` get_update_status()`. As written, `get_chosen_update()` and `get_most_recent_command()` are essentially just getter functions written as top-level functions.

agent/src/apiclient.rs
gthao313: Sure! will update it. Thanks

agent/src/agentclient.rs
```diff
@@ -0,0 +1,318 @@
...
+            .context(error::BottlerocketNodeNotExist {
+                node_name: associated_bottlerocketnode_name,
+            });
+
+        let bottlerocket_nodes_exits = match bottlerocket_node {
+            Ok(_) => true,
+            Err(_) => false,
+        };
+
+        Ok(bottlerocket_nodes_exits)
```
arnaldo2792:
```suggestion
        Ok(bottlerocket_nodes
            .get(&associated_bottlerocketnode_name)
            .await
            .context(error::BottlerocketNodeNotExist {
                node_name: associated_bottlerocketnode_name,
            }).is_ok())
```

agent/src/agentclient.rs
```diff
@@ -0,0 +1,318 @@
...
+                    node_name: associated_node_name.clone(),
+                    node_uid: associated_node_uid.clone(),
+                });
+
+                Ok(BottlerocketNodeSelector {
+                    node_name: associated_node_name,
+                    node_uid: associated_node_uid,
+                })
+            }
+        }
```
arnaldo2792: Nit:

```suggestion
        if self.node_selector.is_none() {
              let associated_node_name = env::var("MY_NODE_NAME").context(error::GetNodeName)?;

                let nodes: Api<Node> = Api::all(self.k8s_client.clone());
                let associated_node: Node = nodes
                    .get(&associated_node_name)
                    .await
                    .context(error::FetchNode)?;

                let associated_node_uid = associated_node
                    .metadata
                    .uid
                    .context(error::MissingNodeUid)?;
                self.node_selector = Some(BottlerocketNodeSelector {
                    node_name: associated_node_name.clone(),
                    node_uid: associated_node_uid.clone(),
                });
        }
        
        Ok(self.node_selector.clone())
```

agent/src/agentclient.rs
```diff
@@ -0,0 +1,318 @@
...
+        let server_domain = get_server_domain();
+
+        let node_req = UpdateBottlerocketNodeRequest {
+            node_status: _current_metadata,
+            node_selector: current_node_selector.clone(),
+        };
+
+        let response = client
+            .put(format!(
+                "http://{}/bottlerocket-node-resource",
```
arnaldo2792: Nit:

Shouldn't this be a constant?

agent/src/apiclient.rs
```diff
@@ -0,0 +1,359 @@
...
+    let mut attempts: i8 = 0;
+    // Retry up to 5 times in case the Update API is busy; Waiting 10 seconds between each attempt.
+    while attempts < MAX_ATTEMPTS {
+        let output = Command::new(API_CLIENT_BIN)
+            .args(&args.clone())
+            .output()
+            .unwrap();
+        // .context(ApiClientCommand { args.clone() })?;
```
arnaldo2792: I think you forgot to remove the comment here? Remember not to call `unwrap` because it could cause a panic :fearful:

agent/src/apiclient.rs
```diff
@@ -0,0 +1,359 @@
...
+
+        if output.status.success() {
+            return Ok(output);
+        }
+
+        // Extract error statuscode from stderr string
+        // Error Example:
+        // "Failed POST request to '/actions/refresh-updates': Status 423 when POSTing /actions/refresh-updates: Update lock held\n"
+        let error_content = String::from_utf8_lossy(&output.stderr).to_string();
+        println!("{}: {:?} {:?}", error_content, args, output);
```
arnaldo2792: `log::info!` instead of `println!`?

agent/src/apiclient.rs
```diff
@@ -0,0 +1,359 @@
...
+        let error_content_split_by_status: Vec<&str> = error_content.split("Status").collect();
+        let error_content_split_by_whitespace: Vec<&str> = error_content_split_by_status[1]
+            .split_whitespace()
+            .collect();
+        let error_statuscode = error_content_split_by_whitespace[0];
```
arnaldo2792: Nit: this could be a function `parse_error` or `status_code_from_error`

agent/src/apiclient.rs
```diff
@@ -0,0 +1,359 @@
...
+        apiclient_error::UpdateStage {
+            expect_state: "Staged".to_string(),
+            update_state: update_status.update_state,
+        }
+    );
+
+    // Activate the prepared update
+    activate_update().await?;
+
+    // Raise error if failed to activate update or update action performed out of band"
```
arnaldo2792: Nit: extra `"` at the end

agent/src/error.rs
```diff
@@ -0,0 +1,94 @@
...
+    #[snafu(display("Unable to get associated node name: {}", source))]
+    GetNodeName { source: std::env::VarError },
+
+    #[snafu(display(
+        "Unable to fetch the Kubernetes node associated with this pod: {}",
+        source
+    ))]
+    FetchNode { source: kube::Error },
+
+    #[snafu(display("Unable to get Node uid because of missing"))]
```
arnaldo2792: Nit: your forgot to say what is missing?

agent/src/error.rs
```diff
@@ -0,0 +1,94 @@
...
+        node_name: String,
+        source: kube::Error,
+    },
+
+    #[snafu(display(
+        "Unable to get Bottlerocket node 'status' because of missing 'status' value"
+    ))]
+    MissingBottlerocketNodeStatus,
+
+    #[snafu(display("Uable to get Bottlerocket node 'spec' because of missing 'spec' value"))]
```
arnaldo2792: Nit: typo

```suggestion
    #[snafu(display("Unable to get Bottlerocket node 'spec' because of missing 'spec' value"))]
```

agent/src/error.rs
```diff
@@ -0,0 +1,94 @@
...
+    ))]
+    GetUptime { source: std::io::Error },
+
+    #[snafu(display("Unable to convert `{} `to datetime: {}", uptime, source))]
+    ConvertStringToDatetime {
+        uptime: String,
+        source: chrono::ParseError,
+    },
+    #[snafu(display(
+        "Unable to convert response of interacting with ctcustom resource to readable text: {}",
```
arnaldo2792: Nit: typo? or is `ctcustom` OK?

arnaldo2792 (commented):
I mostly have Nits, but there is a particular comment around an `unwrap`, that's the only thing that concerns me. I can approve it once that is fixed :hamster: .

cbgbt (approved):
🚢

agent/src/agentclient.rs
arnaldo2792: A wild typo appeared `condtion` :+1:

agent/src/apiclient.rs
```diff
@@ -0,0 +1,351 @@
...
+            source: std::io::Error,
+        },
+
+        #[snafu(display("Failed to deserialize Os info: {}", source))]
+        OsContent { source: serde_json::Error },
+
+        #[snafu(display("Failed to deserialize update status: {}", source))]
+        UpdateStatusContent { source: serde_json::Error },
+
+        #[snafu(display("failed to refresh updates or update action performed out of band"))]
```
arnaldo2792, line 328: Nit: there are a few messages that doesn't start with a capital letter, or have a `.` and the following letter is in lower case. Fix for consistency.

arnaldo2792 (approved):
A few non-blocking nits, but I'm OK with those being updated later. Good job! :partying_face:

PR #96, opened 2021-10-25
Add agent yaml generation

gthao313:
**Issue number:**
#80 #78  #79

**Description of changes:**
This change adds an agent component build step which converts agent Rust models into YAML files which can be deployed to kubernetes.

Also mount a service token provided by Token Volume Projection on Pod which gives pod permission to use apiclient to update `bottlerocketCustomresourceNode`.

**Testing done:**
Apply agent yaml file to EKS cluster,  all things had been created successfully
```
kubectl -n brupop-bottlerocket-aws get pods
NAME                               READY   STATUS             RESTARTS   AGE
brupop-agent-4qnjs                 1/1     Running            0          23m
brupop-agent-n4pnk                 1/1     Running            0          23m
brupop-agent-nsdmh                 1/1     Running            0          23m
brupop-apiserver-ff767d4bd-b4m8s   1/1     Running            0          10h
brupop-apiserver-ff767d4bd-h7n6q   1/1     Running            0          10h
brupop-apiserver-ff767d4bd-nqzl5   1/1     Running            0          10h

[tianhg@ip-172-31-39-243 bottlerocket-update-operator]$ kubectl -n brupop-bottlerocket-aws get se
secrets                                     securitygrouppolicies.vpcresources.k8s.aws  serviceaccounts                             services

[tianhg@ip-172-31-39-243 bottlerocket-update-operator]$ kubectl -n brupop-bottlerocket-aws get secrets
NAME                                            TYPE                                  DATA   AGE
brupop                                          kubernetes.io/dockerconfigjson        1      11h
kubernetes.io/service-account-token   3      11h
default-token-q5xzl                             kubernetes.io/service-account-token   3      7d12h

kubectl -n brupop-bottlerocket-aws get clusterrole
brupop-agent-role                                                      2021-10-25T21:31:33Z

[tianhg@ip-172-31-39-243 bottlerocket-update-operator]$ kubectl -n brupop-bottlerocket-aws get namespaces
NAME                      STATUS   AGE
brupop-bottlerocket-aws   Active   7d12h
```

**Terms of contribution:**

### DIFF
--- a/models/src/agent.rs
+++ b/models/src/agent.rs
@@ -1,0 +1,261 @@
+use crate::constants::{
+    AGENT, AGENT_NAME, APP_COMPONENT, APP_MANAGED_BY, APP_PART_OF, BRUPOP,
+    BRUPOP_INTERFACE_VERSION, LABEL_BRUPOP_INTERFACE_NAME, LABEL_COMPONENT, NAMESPACE,
+};
+use k8s_openapi::api::apps::v1::{DaemonSet, DaemonSetSpec};
+use k8s_openapi::api::core::v1::{
+    Affinity, Container, EnvVar, EnvVarSource, HostPathVolumeSource, LocalObjectReference,
+    NodeAffinity, NodeSelector, NodeSelectorRequirement, NodeSelectorTerm, ObjectFieldSelector,
+    PodSpec, PodTemplateSpec, ProjectedVolumeSource, ResourceRequirements, SELinuxOptions,
+    SecurityContext, ServiceAccount, ServiceAccountTokenProjection, Volume, VolumeMount,
+    VolumeProjection,
+};
+use k8s_openapi::api::rbac::v1::{ClusterRole, ClusterRoleBinding, PolicyRule, RoleRef, Subject};
+use k8s_openapi::apimachinery::pkg::api::resource::Quantity;
+use k8s_openapi::apimachinery::pkg::apis::meta::v1::LabelSelector;
+use kube::api::ObjectMeta;
+use maplit::btreemap;
+
+const BRUPOP_AGENT_SERVICE_ACCOUNT: &str = "brupop-agent-service-account";
+const BRUPOP_AGENT_CLUSTER_ROLE: &str = "brupop-agent-role";
+
+/// Defines the brupop-agent service account
+pub fn agent_service_account() -> ServiceAccount {
+    ServiceAccount {
+        metadata: ObjectMeta {
+            name: Some(BRUPOP_AGENT_SERVICE_ACCOUNT.to_string()),
+            namespace: Some(NAMESPACE.to_string()),
+            annotations: Some(btreemap! {
+                "kubernetes.io/service-account.name".to_string() => BRUPOP_AGENT_SERVICE_ACCOUNT.to_string()
+            }),
+            ..Default::default()
+        },
+        ..Default::default()
+    }
+}
+
+/// Defines the brupop-agent cluster role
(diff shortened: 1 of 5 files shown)


gthao313:
Push above add `securityContext` to solve apiclient call permission deny issue

models/src/agent.rs
```diff
@@ -0,0 +1,253 @@
...
+        metadata: ObjectMeta {
+            name: Some(BRUPOP_AGENT_CLUSTER_ROLE.to_string()),
+            namespace: Some(NAMESPACE.to_string()),
+            ..Default::default()
+        },
+        rules: Some(vec![
+            PolicyRule {
+                api_groups: Some(vec!["".to_string()]),
+                resources: Some(vec!["nodes".to_string()]),
+                verbs: vec!["create", "get", "list", "patch", "update", "watch"]
```
cbgbt, line 49: I think this may be a bit too allowing. Do we need more than `get`, `list`, `watch`? I'm wondering if we even need `list`.

I suppose if we're calling the drain API, it may be that we need these. In which case, we would need to later remove them for #97.

models/src/agent.rs
```diff
@@ -0,0 +1,253 @@
...
+                    .collect(),
+                ..Default::default()
+            },
+            PolicyRule {
+                api_groups: Some(vec!["brupop.bottlerocket.aws".to_string()]),
+                resources: Some(vec![
+                    "bottlerocketnodes".to_string(),
+                    "bottlerocketnodes/status".to_string(),
+                ]),
+                verbs: vec!["create", "get", "list", "patch", "update", "watch"]
```
cbgbt: `create`, `patch`, and `update` should all be handled by the `apiserver` component, so we should remove them here.

models/src/agent.rs
```diff
@@ -0,0 +1,253 @@
...
+        image_pull_secret.map(|secret| vec![LocalObjectReference { name: Some(secret) }]);
+
+    DaemonSet {
+        metadata: ObjectMeta {
+            labels: Some(
+                btreemap! {
+                    APP_COMPONENT => "agent",
+                    APP_MANAGED_BY => "brupop",
+                    APP_PART_OF => "brupop",
+                    LABEL_COMPONENT => "agent",
```
cbgbt: It would be ideal to use constants for these.

models/src/agent.rs
```diff
@@ -0,0 +1,253 @@
...
+                }
+                .iter()
+                .map(|(k, v)| (k.to_string(), v.to_string()))
+                .collect(),
+            ),
+            name: Some("brupop-agent".to_string()),
```
cbgbt: Same here, a constant would be ideal for the name.

models/src/agent.rs
```diff
@@ -0,0 +1,253 @@
...
+            namespace: Some(NAMESPACE.to_string()),
+            ..Default::default()
+        },
+        spec: Some(DaemonSetSpec {
+            selector: LabelSelector {
+                match_labels: Some(btreemap! { LABEL_COMPONENT.to_string() => "agent".to_string()}),
```
cbgbt: In brupop 0.1.0, we run on any node with the label `bottlerocket.aws/updater-interface-version=2.0.0`. I think it might be ideal for us to do the same here.

models/src/agent.rs
gthao313: I'll remove it once we confirm drain API does not need it.

cbgbt (approved):
:shipit:

arnaldo2792 (approved):
:tada:

Issue #88, opened 2021-10-07
0.2.0: Remove openssl build from brupop's Dockerfile


cbgbt:
We should integrate our `openssl` compiled against musl into the Bottlerocket SDK image, or otherwise utilize a build process which will allow us to properly dynamically link against openssl at runtime.

One stopgap option could be to dynamically link against the openssl currently in the SDK image, and do away with our `scratch`-based resulting image for the time being.

cbgbt:
Thanks, @jpculp !

Issue #87, opened 2021-10-07
0.2.0: Integrate license-checking using `cargo-deny`


cbgbt:
[`cargo-deny`](https://github.com/EmbarkStudios/cargo-deny) can be used to determine if we are including crates with licenses that we prefer. We should add a make target to run `cargo-deny` during build, and a github workflow to ensure it is run against incoming code changes.

cbgbt:
Thanks, @jpculp !

PR #98, opened 2021-10-29
0.2.0: Improve Makefile

jpculp:
**Issue number:**

#87
#88 

**Description of changes:**

- Updates to Bottlerocket SDK v0.23.0 with OpenSSL built in.

- Adds a `cargo` configuration to ensure that Rust binaries are linked with our toolchain rather than the default `gcc`.

- Moves the implicit fetch step outside of the build, so that an incremental build does not need to download crates again.

- Updates image tag to be consistent with previous releases.

- Adds `make clean`.

- Adds a license check to the Makefile, `make check-licenses`, which is also a dependency of build.

**Testing done:**

Ran `make build` and `make brupop-image` on both x86_64 and aarch64 hosts.

**Terms of contribution:**


jpculp:
- Added `fetch` and `check-licenses` as dependencies for the `brupop-image` target.

cbgbt (approved):
:shipit:

arnaldo2792 (approved):
:hamster:

gthao313 (approved):

Issue #90, opened 2021-10-07
0.2.0: Create initial controller component


cbgbt:
The controller component is responsible for watching for creation and status changes to `BottlerocketNode` custom resources and orchestrating state changes to drive updates to each node.

We should use the [`reflector` abstraction from kube-rs](https://github.com/kube-rs/kube-rs#reflectors) to efficiently query k8s for the node changes.

Issue #82, opened 2021-09-27
0.2.0 controller: Create brupop cluster state machine


cbgbt:
The brupop controller drives updates across nodes in the cluster by modifying the `spec` of `BottlerocketNode` objects associated with each node. To complete this issue, we must create the state machine structure, as well as brupop's logic for driving changes to the state machine based on the cluster's status.

Issue #81, opened 2021-09-27
 brupop 0.2.0 agent: Create k8s deployment targets for development


cbgbt:
Similar to #77 , we must create instructions or a development Make target to add the brupop agent's `DaemonSet` to a kubernetes cluster.

cbgbt:
Closing the remainder of this in favor of #109

PR #102, opened 2021-11-02
0.2.0: Implement Brupop Controller

cbgbt:
**Issue number:** #90,  #82,  #81

**Description of changes:** This adds the implementation of the Brupop controller, as well as a few various fixes and refactoring of related components.

**Testing done:** I've deployed all components (`apiserver`, `agent`, and `controller`) to my own cluster of Bottlerocket nodes. After applying labels and watching the `brn` objects in the cluster, the controller successfully moves all of these nodes from `1.2.1` to `1.3.0`.

``` $kubectl get brn
NAME                                               STATE              VERSION   TARGET STATE       TARGET VERSION
brn-ip-192-168-10-1.us-west-2.compute.internal     WaitingForUpdate   1.3.0     WaitingForUpdate   
brn-ip-192-168-55-52.us-west-2.compute.internal    MonitoringUpdate   1.3.0     WaitingForUpdate   
brn-ip-192-168-81-141.us-west-2.compute.internal   WaitingForUpdate   1.2.1     WaitingForUpdate   <no value>
```

**Terms of contribution:**


controller/src/controller.rs
```diff
@@ -0,0 +1,149 @@
...
+                for brn in nodes.drain(..) {
+                    // Timeouts and errors are logged by instrumentation in `progress_node()`.
+                    #[allow(unused_must_use)]
+                    {
+                        self.progress_node(brn).await;
+                    }
+                }
+            } else {
+                // If there's nothing to operate on, check to see if any other nodes are ready for action.
+                let new_active_node = self.search_for_node_to_update().await;
```
arnaldo2792: Shouldn't this be executed always? Regardless of the active nodes being empty?

models/src/node/client.rs
```diff
@@ -0,0 +1,234 @@
...
+        let br_node_status_patch =
+            serde_json::to_value(br_node_status_patch).context(error::CreateK8SPatch)?;
+
+        let api: Api<BottlerocketNode> =
+            Api::namespaced(self.k8s_client.clone(), constants::NAMESPACE);
+
+        api.patch_status(
+            &selector.brn_resource_name(),
+            &PatchParams::default(),
+            &Patch::Merge(&br_node_status_patch),
```
arnaldo2792, line 192: Do you need to transform `br_node_status_patch`? It looks like `Patch::Merge()` receives an object that derives from `Serialize`, and `BottlerocketNodeStatusPatch` already derives from it.

models/src/node/mod.rs
```diff
@@ -0,0 +1,259 @@
...
+    pub fn new_starting_now(state: BottlerocketNodeState, version: Option<Version>) -> Self {
+        Self::new(state, Some(Utc::now()), version)
+    }
+
+    /// JsonSchema cannot appropriately handle DateTime objects. This accessor returns the transition timestamp
+    /// as a DateTime.
+    pub fn state_timestamp(&self) -> Option<DateTime<Utc>> {
+        self.state_transition_timestamp.as_ref().map(|ts_str| {
+            DateTime::parse_from_rfc3339(ts_str)
+                .expect("state_transition_timestamp must be rfc3339 string.")
```
arnaldo2792: Shouldn't this use `context`?

controller/src/controller.rs
cbgbt: We want to restrict brupop to only operate on a single node at a time. Even if other nodes are ready to become active, we want to wait until no other nodes are active.

controller/src/main.rs
```diff
@@ -1,3 +1,226 @@
...
+
+            // TODO(seankell) Timeout handling will be added in a future PR.
+            Ok(())
+        }
+    }
+
+    /// This function searches all `BottlerocketNode`s for those which can be transitioned to a new state.
+    /// The state transition is then attempted. If successful, this node should be detected as part of the active
+    /// set during the next iteration of the controller's event loop.
+    async fn search_for_node_to_update(&self) -> Option<BottlerocketNode> {
```
gthao313: Question here. Does this function only find one node to update every time? why not finding all nodes and return a list of nodes which are qualified to update, then main loop can handle them.

controller/src/controller.rs
```diff
@@ -0,0 +1,149 @@
...
+impl<T: BottlerocketNodeClient> BrupopController<T> {
+    pub fn new(node_client: T, brn_reader: Store<BottlerocketNode>) -> Self {
+        BrupopController {
+            node_client,
+            brn_reader,
+        }
+    }
+
+    /// Returns a list of all `BottlerocketNode` objects in the cluster.
+    fn all_nodes(&self) -> Vec<BottlerocketNode> {
```
gthao313, line 32: Curious if we need this function here? Since you mentioned in my PR before that those kind functions essentially just getter functions written as top-level functions. Should we just use self.brn_reader.state() when we need it? : )

models/src/node/mod.rs
cbgbt: I had the debate about writing this on whether or not this indicated a software bug or an actual error and have decided that even in either scenario, returning an `Err` value is probably best. I'll change it.

controller/src/controller.rs
cbgbt: Good question! In this case, this function *is* a getter function. Getter functions are fine -- they let us change the inner workings of the struct without changing the interface -- but they should probably be provided as a method on the struct that they are retrieving data from. When I mentioned a top-level function, I meant that the function in the other PR was implemented on the module, rather than the struct it was accessing.

To clarify, I think either making an inner value `pub` could be appropriate if it's okay for customers to reach in and change the value as well as retrieve it, or something like this:
```
struct A {
    inner_value: String
}

impl A {
    pub fn inner_value(&self) -> &str {
        &self.inner_value
    }
}
```

Is more appropriate than something like this:

```
struct A {
    inner_value: String,
}

pub fn get_inner_value(my_a: &A) -> &str {
    a.inner_value
}
```

models/src/node/client.rs
cbgbt: Nice catch! Fixed.

cbgbt:
^ FIxes concerns raised by Arnaldo.

gthao313 (approved):

agent/src/agentclient.rs
```diff
@@ -227,16 +225,16 @@ impl BrupopAgent {
...
                         update().await.context(error::UpdateActions {
                             action: "Perform".to_string(),
                         })?;
                     }
-                    BottlerocketNodeState::RebootingToUpdate => {
+                    BottlerocketNodeState::RebootedToUpdate => {
                         log::info!("Rebooting node to complete update");
 
-                        if ensure_reboot_happened(&bottlerocket_node_status)? {
+                        if running_desired_version(&bottlerocket_node_spec).await? {
```
bcressey, line 237: If the update is bad and we roll back to the previous partition bank, could this put the node in an infinite reboot loop? Or is that handled by some other means?

yamlgen/deploy/brupop-resources.yaml
```diff
@@ -0,0 +1,376 @@
...
+              memory: 50Mi
+            requests:
+              cpu: 10m
+              memory: 50Mi
+          securityContext:
+            seLinuxOptions:
+              level: s0
+              role: system_r
+              type: super_t
+              user: system_u
```
bcressey: FWIW, you can set `privileged: true` in `securityContext` to be able to interact with the API.

The major permissions that `super_t` grants are:
* relabeling files with arbitrary SELinux labels
* modifying files in the host's `/etc` and the datastore `/var/lib/bottlerocket` 

If we don't need to do those things - and we shouldn't! - then the `control_t` label granted by "privileged" should be fine.

yamlgen/deploy/brupop-resources.yaml
```diff
@@ -0,0 +1,376 @@
...
+  name: brupop-controller-deployment
+  namespace: brupop-bottlerocket-aws
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      brupop.bottlerocket.aws/component: brupop-controller
+  strategy:
+    rollingUpdate:
+      maxUnavailable: 100
```
bcressey: Should this be 100%?

models/src/node/client.rs
```diff
@@ -0,0 +1,231 @@
...
+        selector: &BottlerocketNodeSelector,
+        spec: &BottlerocketNodeSpec,
+    ) -> Result<()> {
+        (**self).update_node_spec(selector, spec).await
+    }
+}
+
+#[derive(Clone)]
+/// Concrete implementation of the `BottlerocketNodeClient` trait. This implementation will almost
+/// certainly be used in ggany case that isn't a unit test.
```
bcressey:
```suggestion
/// certainly be used in any case that isn't a unit test.
```

agent/src/agentclient.rs
cbgbt: Yes, it can. It's not yet handled, but the plan is to track the number of attempted updates, eventually giving up and emitting a failure metric to Prometheus via opentelemetry. That handling wasn't codified in an issue, so I've opened #123

yamlgen/deploy/brupop-resources.yaml
cbgbt: The controller runs as a single node, to avoid having to coordinate on which nodes to progress next. If it dies, we reschedule it.

I think the right move would be to move away from `rollingUpdate` strategy to [`recreate`](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#recreate-deployment)

models/src/node/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+    /// Records the desired state of the `BottlerocketNode`
+    pub state: BottlerocketNodeState,
+    /// The time at which the most recent state was set as the desired state.
+    state_transition_timestamp: Option<String>,
+    /// The desired update version, if any.
+    #[validate(regex = "SEMVER_RE")]
+    version: Option<String>,
+}
+
+impl BottlerocketNode {
```
somnusfish, line 138: Where is this coming from? Is it a struct? But I couldn't find the definition for this struct.

models/src/node/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+        self.version.as_ref().map(|v| Version::from_str(v).unwrap())
+    }
+}
+
+/// `BottlerocketNodeStatus` surfaces the current state of a bottlerocket node. The status is updated by the host agent,
+/// while the spec is updated by the brupop controller.
+#[derive(Clone, Debug, Default, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
+pub struct BottlerocketNodeStatus {
+    #[validate(regex = "SEMVER_RE")]
+    current_version: String,
```
somnusfish, line 198: Why do we store `String` here and convert between String and Version? Why didn't we directly store Version?

models/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 pub mod agent;
 pub mod apiserver;
 pub mod constants;
+pub mod controller;
```
somnusfish, line 4: The mods inside models have a very similar structure for the crates under brupop, how do you design what should be in models and what should be in other lib+bin?

controller/src/controller.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// Returns the set of BottlerocketNode objects which is currently being acted upon.
+    ///
+    /// Nodes are being acted upon if they are not in the `WaitingForUpdate` state, or if their desired state does
+    /// not match their current state.
+    #[instrument(skip(self))]
+    fn active_node_set(&self) -> BTreeMap<String, BottlerocketNode> {
+        self.all_nodes()
+            .iter()
+            .filter(|brn| {
+                brn.status.as_ref().map_or(false, |brn_status| {
```
somnusfish, line 45: Would map_or_else be better here?

controller/src/controller.rs
```diff
@@ -0,0 +1,149 @@
...
+            // TODO(seankell) Timeout handling will be added in a future PR.
+            Ok(())
+        }
+    }
+
+    /// This function searches all `BottlerocketNode`s for those which can be transitioned to a new state.
+    /// The state transition is then attempted. If successful, this node should be detected as part of the active
+    /// set during the next iteration of the controller's event loop.
+    #[instrument(skip(self))]
+    async fn search_for_node_to_update(&self) -> Option<BottlerocketNode> {
```
somnusfish: This method secretly changed node's spec with a name seems only pick up some nodes.

models/src/node/mod.rs
cbgbt: This struct is generated when deriving `CustomResource` and the `kube` macro on the `BottlerocketNodeSpec` struct. See [`CustomResource`](https://docs.rs/kube-derive/0.64.0/kube_derive/derive.CustomResource.html) documentation.

models/src/node/mod.rs
cbgbt: This `BottlerocketNodeStatus` is being serialized into an object in Kubernetes. For this to work, it needs to support `Serialize`, `Deserialize`, and `JsonSchema`, which I believe isn't all true for `semver::Version`. Kubernetes, however, does allow you to serialize strings and add regular expressions to the schema which ensure they match a pattern. This `validate` macro adds the schema requirement for us.

models/src/lib.rs
cbgbt: So far, anything that is written to Kubernetes as an object, or object models that are shared between many crates end up in the `models` crate. There's definitely some discretion to placing code.

controller/src/controller.rs
cbgbt: It's possible I'm misunderstanding your suggestion, but I think `map_or_else` is generally better if the default value is computed by calling a function. If you use `map_or`, the default argument is evaluated when the statement is parsed, which could be wasteful if you end up not needing the value.

cbgbt:
^ Addresses feedback from @somnusfish

cbgbt:
^ Fixes a file that wasn't autoformatted by my IDE.

controller/src/controller.rs
somnusfish: Here is what I found: 
```Arguments passed to map_or are eagerly evaluated; if you are passing the result of a function call, it is recommended to use map_or_else, which is lazily evaluated.```
I originally thought map_or evaluate both value and map_or_else lazily evaluate the brn_status is Some(), but looks like it applies only on default value.

somnusfish (approved):

bcressey (approved):

PR #114, opened 2021-11-04
0.2.0 Improve Readability of State Machine State Names

cbgbt:
**Testing done:**
Straightforward change to the state names for our update state machine, after some in-person discussions that trended towards these being less ambiguous than the previous names.

**Terms of contribution:**

### DIFF
--- a/controller/src/controller.rs
+++ b/controller/src/controller.rs
@@ -43,7 +43,7 @@ impl<T: BottlerocketNodeClient> BrupopController<T> {
             .iter()
             .filter(|brn| {
                 brn.status.as_ref().map_or(false, |brn_status| {
-                    brn_status.current_state != BottlerocketNodeState::WaitingForUpdate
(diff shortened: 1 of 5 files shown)


gthao313 (approved):
Thanks for doing this! :rocket:

controller/src/controller.rs
```diff
@@ -43,7 +43,7 @@ impl<T: BottlerocketNodeClient> BrupopController<T> {
             .iter()
             .filter(|brn| {
                 brn.status.as_ref().map_or(false, |brn_status| {
```
somnusfish, line 45: Seems `map_or_else` would be more suitable here?

bcressey (approved):

controller/src/controller.rs
cbgbt: I chatted with @somnusfish about this elsewhere. To summarize, we would only want `map_or_else` if the default value is something that needs to be computed.

Issue #99, opened 2021-10-29
0.2.0 agent: add retries with exponential backoff to network functions


gthao313:
> I think it would be a good improvement to add retries with exponential backoff to this.

Adding retries with exponential backoff to create_metadata_custom_resource and update_metadata_custom_resource functions.

Issue #97, opened 2021-10-26
0.2.0: Integrate `drain` API into apiserver


cbgbt:
We need to validate that bottlerocketnode drainage is being invoked by the daemon on the pod which is being drained. In order to do this, we must implement a draining API method in the `apiserver`, and then invoke this from the agent.

PR #115, opened 2021-11-05
0.2.0: Add client abstraction for apiserver

cbgbt:
**Issue number:** Beginning to implement #99 #97 #80

**Description of changes:**

Add apiserver client abstraction
    
This unifies calls to the apiserver under a single client. The intent is to start using HTTP headers to hold k8s Node and authN information so that it can be validated by actix-web middleware.

The client implements the header changes, as well as retries with exponential backoff.

**Testing done:**
Unit tests pass. Ran the controller and noted that the brupop agent successfully made writes to `brn` objects.
```
$ kubectl get brn
NAME                                               STATE   VERSION   TARGET STATE   TARGET VERSION
brn-ip-192-168-26-177.us-west-2.compute.internal   Idle    1.2.1     StagedUpdate   1.3.0
brn-ip-192-168-42-241.us-west-2.compute.internal   Idle    1.2.1     Idle           <no value>
```

**Terms of contribution:**

### DIFF
--- a/apiserver/Cargo.toml
+++ b/apiserver/Cargo.toml
@@ -5,35 +5,40 @@ edition = "2018"
 publish = false
 license = "Apache-2.0 OR MIT"
 
+[features]
+default = ["client", "server"]
+client = []
+server = []
+
+
 [dependencies]
 models = { path = "../models", version = "0.1.0" }
 
-actix-web = { version = "4.0.0-beta.9", default-features = false }
-
 # tracing-actix-web version must align with actix-web version
-tracing-actix-web = "0.4.0-beta.14"
+actix-web = { version = "4.0.0-beta.9", default-features = false }
 tracing = "0.1"
-tracing-subscriber = { version = "0.3", features = ["registry", "env-filter"] }
+tracing-actix-web = "0.4.0-beta.14"
 tracing-bunyan-formatter = "0.3"
+tracing-subscriber = { version = "0.3", features = ["registry", "env-filter"] }
 opentelemetry = { version = "0.16", features = ["rt-tokio-current-thread"]}
 tracing-opentelemetry = "0.16"
-
-
-env_logger = "0.9"
-futures = "0.3"
 
 # k8s-openapi must match the version required by kube and enable a k8s version feature
 k8s-openapi = { version = "0.13.1", default-features = false, features = ["v1_20"] }
 kube = { version = "0.62.0", default-features = true, features = [ "derive"] }
 
+async-trait = "0.1"
+futures = "0.3"
 lazy_static = "1.4"
 log = "0.4"
-reqwest = { version = "0.11", features =  [ "json" ] }
+mockall = { version = "0.10", optional = true }
+reqwest = { version = "0.11", features =  [ "json", "native-tls" ] }
 schemars = "0.8"
 serde = { version = "1", features = [ "derive" ] }
 serde_json = "1"
 snafu = "0.6"
 tokio = { version = "1", features = ["macros", "rt-multi-thread", "time"] }
+tokio-retry = "0.3"
 
 [dev-dependencies]
 mockall = "0.10"
--- a/apiserver/src/api.rs
+++ b/apiserver/src/api.rs
@@ -1,18 +1,62 @@
-use crate::error::{self, Result};
-use crate::telemetry;
+use crate::{
+    constants::{
+        HEADER_BRUPOP_K8S_AUTH_TOKEN, HEADER_BRUPOP_NODE_NAME, HEADER_BRUPOP_NODE_UID,
+        NODE_RESOURCE_ENDPOINT,
+    },
+    error::{self, Result},
+    telemetry,
+};
 use models::constants::APISERVER_HEALTH_CHECK_ROUTE;
 use models::node::{BottlerocketNodeClient, BottlerocketNodeSelector, BottlerocketNodeStatus};
 
 use actix_web::{
+    http::HeaderMap,
     web::{self, Data},
     App, HttpRequest, HttpResponse, HttpServer, Responder,
 };
-use serde::{Deserialize, Serialize};
 use serde_json::json;
-use snafu::ResultExt;
+use snafu::{OptionExt, ResultExt};
 use tracing_actix_web::{RootSpan, TracingLogger};
 
-const NODE_RESOURCE_ENDPOINT: &'static str = "/bottlerocket-node-resource";
+use std::convert::TryFrom;
+
+/// A struct containing information intended to be passed to the apiserver via HTTP headers.
+struct ApiserverCommonHeaders {
+    node_selector: BottlerocketNodeSelector,
+    _k8s_auth_token: String,
+}
+
+/// Returns a string value extracted from HTTP headers.
+fn extract_header_string(headers: &HeaderMap, key: &'static str) -> Result<String> {
+    Ok(headers
+        .get(key)
+        .context(error::HTTPHeaderParse {
+            missing_header: key,
+        })?
+        .to_str()
+        .map_err(|_| error::Error::HTTPHeaderParse {
+            missing_header: key,
+        })?
+        .to_string())
+}
+
+impl TryFrom<&HeaderMap> for ApiserverCommonHeaders {
+    type Error = error::Error;
+
+    fn try_from(headers: &HeaderMap) -> Result<Self> {
+        let node_name = extract_header_string(headers, HEADER_BRUPOP_NODE_NAME)?;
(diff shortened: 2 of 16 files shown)


apiserver/src/error.rs
```diff
@@ -13,6 +13,9 @@ pub enum Error {
     #[snafu(display("Unable to create client: '{}'", source))]
     ClientCreate { source: kube::Error },
 
+    #[snafu(display("Unable to parse HTTP header. Missing '{}'", missing_header))]
```
somnusfish, line 16: I remember Tom mentioned something like it's not recommended to define different errors in an error file. Do we want to make some adjustments in that case?

apiserver/src/lib.rs
```diff
@@ -1,3 +1,39 @@
...
+pub mod client;
+
+pub(crate) mod constants;
+
+use models::node::{BottlerocketNodeSelector, BottlerocketNodeStatus};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+/// Describes a node for which a BottlerocketNode custom resource should be constructed.
+pub struct CreateBottlerocketNodeRequest {
```
somnusfish, line 18: The name is a little bit confusing. Looks like in this context, bottlerocketnode = bottlerocketnode custom resource, but my first thought on bottlerocketnode is actually a host that runs bottlerocket os. I am not familiar with the naming conventions here, maybe it would be better for us to make the name more clear?

apiserver/Cargo.toml
```diff
@@ -19,15 +24,15 @@ opentelemetry = { version = "0.16", features = ["rt-tokio-current-thread"]}
...
 k8s-openapi = { version = "0.13.1", default-features = false, features = ["v1_20"] }
 kube = { version = "0.62.0", default-features = true, features = [ "derive"] }
 
-log = "0.4"
-reqwest = { version = "0.11", features =  [ "json" ] }
+async-trait = "0.1"
+mockall = { version = "0.10", optional = true }
+futures = "0.3"
+reqwest = { version = "0.11", features =  [ "json", "native-tls" ] }
+tokio-retry = "0.3"
```
somnusfish: Does rust dependencies recommended following alphabetical order?

apiserver/Cargo.toml
```diff
@@ -5,12 +5,17 @@ edition = "2018"
 publish = false
 license = "Apache-2.0 OR MIT"
 
+[features]
```
somnusfish, line 8: This is just a question. Could you give an example about the usage for features in our case and explain what benefits we gain using this features?

apiserver/src/client/error.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use snafu::Snafu;
+
+/// The client result type.
+pub type Result<T> = std::result::Result<T, ClientError>;
+
+/// Error type representing issues using an apiserver client.
+#[derive(Debug, Snafu)]
+#[snafu(visibility = "pub")]
+pub enum ClientError {
```
somnusfish, line 11: Why not define errors for UncordonBottlerocketNodeRequest and DrainAndCordonBottlerocketNodeRequest ?

apiserver/src/client/webclient.rs
```diff
@@ -0,0 +1,176 @@
...
+        })
+        .await
+    }
+
+    #[instrument]
+    async fn drain_and_cordon_node(
+        &self,
+        _req: DrainAndCordonBottlerocketNodeRequest,
+    ) -> Result<()> {
+        unimplemented!()
```
somnusfish: I feel this is something you are plan to implement in future, would `todo!` making more sense?

apiserver/src/client/webclient.rs
```diff
@@ -0,0 +1,176 @@
...
+    strategy::{jitter, ExponentialBackoff},
+    Retry,
+};
+use tracing::instrument;
+
+// The web client uses exponential backoff.
+// These values configure how long to delay between tries.
+const RETRY_BASE_DELAY: Duration = Duration::from_millis(100);
+const RETRY_MAX_DELAY: Duration = Duration::from_secs(10);
+const NUM_RETRIES: usize = 5usize;
```
gthao313: Is this a typo or a syntax? `5usize`

apiserver/src/error.rs
cbgbt: I think this is a good idea. I've opened #121 to track moving errors to submodules, since these span multiple existing crates. Is it okay to defer fixing that to a future PR?

apiserver/src/lib.rs
cbgbt: That's fair feedback, but I think this might also be a more sweeping change. I've opened #121. Do you mind providing some feedback on naming there?

apiserver/Cargo.toml
cbgbt: Sure! So, in this case the `agent` needs to communicate with the `apiserver` using the code in `apiserver::client`.

So in the `agent`'s `Cargo.toml`, we add a dependency on `apiserver` but only enable the `client` feature. In this case, when the compiler builds the `agent` crate, it will not compile in the `apiserver`'s server code, which means we can't accidentally use code which is meant only for the server component. It also improves compile times.

apiserver/src/client/error.rs
cbgbt: I'm leaving implementing these for when we handle #97

apiserver/src/client/webclient.rs
cbgbt: It's syntax! Rust allows you to specify numeric literal types as part of literals to avoid having to add type annotations otherwise.

I think it's actually not necessary here, since we add a `usize` type annotation to the constant. I also think you can add `_` to these suffixes to make them more readable, e.g. `5_usize` would be one way to express it.

apiserver/src/error.rs
somnusfish: Sure.

cbgbt:
^ Addresses feedback from @somnusfish and @gthao313

apiserver/src/client/webclient.rs
```diff
@@ -0,0 +1,176 @@
...
+
+#[async_trait]
+impl APIServerClient for K8SAPIServerClient {
+    #[instrument]
+    async fn create_bottlerocket_node(
+        &self,
+        req: CreateBottlerocketNodeRequest,
+    ) -> Result<BottlerocketNode> {
+        Retry::spawn(retry_strategy(), || async {
+            let http_client = reqwest::Client::new();
```
somnusfish, line 92: Just curious why not use actix_web::client here to match the server.

apiserver/src/api.rs
```diff
@@ -1,95 +1,126 @@
...
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
-/// Describes updates to a BottlerocketNode object's `status`.
-pub struct UpdateBottlerocketNodeRequest {
-    pub node_selector: BottlerocketNodeSelector,
-    pub node_status: BottlerocketNodeStatus,
-}
-
 /// HTTP endpoint which updates the `status` of a BottlerocketNode custom resource on behalf of the caller.
 async fn update_bottlerocket_node_resource<T: BottlerocketNodeClient>(
```
somnusfish, line 99: Actually when I read this file, I have a feeling that it might be better if we put this api.rs to a folder and divide each handler to a separate file, I am not sure if this is suitable for rust code structure arrange though

gthao313 (approved):
:hatched_chick:

apiserver/src/client/webclient.rs
cbgbt: I actually didn't know `actix_web` had a client! I personally think that I trust `reqwest` a bit more, I know it's quite popular in the community (more than 6x more crates.io downloads it seems), but I don't have any strong reason to choose one over the other.

apiserver/src/api.rs
cbgbt: It can certainly be done! Sometimes traversing more files can be a nuisance, but there's not much coupling here so I think it should be ok. I'll try it.

apiserver/src/client/webclient.rs
somnusfish: I am okay with either. The reason I ask this question was I saw the actix-web-opentelemetry had the feature to support telemetry for client, but we can decide which way to go later if we want track client as well.

cbgbt:
^ Rebases atop `0.2.0`

apiserver/src/api.rs
cbgbt: I've implemented this in #124

somnusfish (approved):

Issue #103, opened 2021-11-02
Move creation of BRN & Status into Core Event Loop of Agent


cbgbt:
**Issue or Feature Request:**
We should add creation of the BRN and status objects to the start of the `loop{}` in `agent.run()`. This would ensure that BRN objects that get deleted will be re-created. This should put the object back into its initial `WaitingForUpdate` state, and the agent should handle this appropriately by disregarding any previous update activity.

Issue #100, opened 2021-11-01
0.2.0 agent: Improving fetch_custom_resource to use reflector


gthao313:
We should avoid polling etcd to get bottlerocket node metadata. Instead of, it's better to use [reflector](https://docs.rs/kube/0.18.0/kube/api/struct.Reflector.html) to watch and cache event.

cbgbt:
We should see if we can also use a `reflector` to improve fetching the `Node` object associated with the node, in `get_node_selector()`.

PR #118, opened 2021-11-10
0.2.0 agentclient: use reflector to watch and cache event.

gthao313:
**Issue number:**
#100 #103

**Description of changes:**

1. add creation of the BRN and status objects to the start of the loop{} in agent.run()
2. using reflector to talk with customer resource and nodes instead of polling etcd
3. fix some typods

**Testing done:**
Apply agent component to EKS cluster and verify if it create BRNs accurately.
```
NAME                                           READY   STATUS             RESTARTS   AGE
brupop-agent-9nskt                             1/1     Running            1          35m
brupop-agent-lb7wn                             0/1     Running   1          35m
brupop-agent-nfkw7                             1/1     Running            3          35m
brupop-apiserver-677496b7c4-6kq6q              1/1     Running            1          35m
brupop-apiserver-677496b7c4-cstt4              1/1     Running            1          35m
brupop-apiserver-677496b7c4-rw5m8              1/1     Running            0          35m
brupop-controller-deployment-dbb96dcdd-f9vmv   1/1     Running            0          35m
```
```
NAME                                               STATE   VERSION   TARGET STATE   TARGET VERSION
brn-ip-192-168-70-212.us-west-2.compute.internal   Idle    1.4.2     Idle           <no value>
brn-ip-192-168-71-122.us-west-2.compute.internal   Idle    1.1.1     StagedUpdate   1.4.1
brn-ip-192-168-75-219.us-west-2.compute.internal   Idle    1.4.1     Idle
```

**Terms of contribution:**

### DIFF
--- a/agent/src/agentclient.rs
+++ b/agent/src/agentclient.rs
@@ -1,8 +1,13 @@
 use k8s_openapi::api::core::v1::Node;
 use kube::Api;
 use snafu::{OptionExt, ResultExt};
-use std::env;
 use tokio::time::{sleep, Duration};
+use tokio_retry::{
+    strategy::{jitter, ExponentialBackoff},
+    Retry,
+};
+
+use kube::runtime::reflector::Store;
 
 use crate::{
     apiclient::{boot_update, get_os_info, list_available, prepare, update},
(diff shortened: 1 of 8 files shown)


agent/src/agentclient.rs
```diff
@@ -3,102 +3,123 @@ use k8s_openapi::api::core::v1::Node;
...
         }
     }
 
-    pub async fn check_node_custom_resource_exists(&mut self) -> Result<bool> {
-        let associated_bottlerocketnode_name = node_resource_name(&self.get_node_selector().await?);
+    pub async fn check_node_custom_resource_exists(&self) -> Result<bool> {
         let bottlerocket_nodes: Api<BottlerocketNode> =
             Api::namespaced(self.k8s_client.clone(), NAMESPACE);
 
         Ok(bottlerocket_nodes
```
gthao313: Should I use reflector and query BRN object if node custom resource exists here? The reason why I did not update it because querying `Store` sometime will delay, so sometime when I query `Store`, it return empty to me, which would be recognized  BRN had not been created. To solve this issue, I added retry and sleep to give it more time to make objects data ready. Do you think It's appropriate to use retry and sleep here?

agent/src/agentclient.rs
```diff
@@ -3,102 +3,123 @@ use k8s_openapi::api::core::v1::Node;
...
 use models::node::{
-    node_resource_name, BottlerocketNode, BottlerocketNodeSelector, BottlerocketNodeState,
-    BottlerocketNodeStatus,
+    BottlerocketNode, BottlerocketNodeSelector, BottlerocketNodeState, BottlerocketNodeStatus,
 };
 
 const AGENT_SLEEP_DURATION: Duration = Duration::from_secs(5);
 const NODE_RESOURCE_ENDPOINT: &'static str = "/bottlerocket-node-resource";
+const MAX_ATTEMPTS: i8 = 5;
+const STREAM_SLEEP_DURATION: Duration = Duration::from_secs(10);
```
cbgbt: Would you mind adding some comments elaborating on these constants?

agent/src/agentclient.rs
cbgbt: If this is called during the event loop, it might not be a great idea to always make the API call.

I would either:
* Check the store. If it's not present in the store, make the API call.
* OR do retries with backoff on the store.

I think option 1 makes the most sense, since it returns more quickly.

agent/src/main.rs
```diff
@@ -27,17 +38,60 @@ async fn run_agent() -> Result<()> {
...
+    let brn_lp = ListParams::default()
+        .fields(format!("metadata.name={}", associated_bottlerocketnode_name).as_str());
+    let brn_store = reflector::store::Writer::<BottlerocketNode>::default();
+    let brn_reader = brn_store.as_reader();
+    let brn_reflector = reflector::reflector(brn_store, watcher(brns, brn_lp));
+    let brn_drainer = try_flatten_touched(brn_reflector)
+        .filter_map(|x| async move { std::result::Result::ok(x) })
+        .for_each(|_brn| {
+            // TODO: utilize tracing event
+            log::info!("Processed event for BottlerocketNodes");
```
cbgbt: I would reduce the log level to `debug!` here, as I believe these events are quite noisy.

agent/src/main.rs
```diff
@@ -27,17 +38,60 @@ async fn run_agent() -> Result<()> {
...
+        ListParams::default().fields(format!("metadata.name={}", associated_node_name).as_str());
+    let nodes: Api<Node> = Api::all(k8s_client.clone());
+    let nodes_store = reflector::store::Writer::<Node>::default();
+    let node_reader = nodes_store.as_reader();
+    let node_reflector = reflector::reflector(nodes_store, watcher(nodes, node_lp));
+    let node_drainer = try_flatten_touched(node_reflector)
+        .filter_map(|x| async move { std::result::Result::ok(x) })
+        .for_each(|_node| {
+            // TODO: utilize tracing event
+            log::info!("Processed event for node");
```
cbgbt: Same here, consider reducing the log level to `debug!`.

gthao313:
push above improve reflector retry method to use `ExponentialBackoff` and improve `check_node_custom_resource_exists` to query store.

agent/src/agentclient.rs
```diff
@@ -3,102 +3,134 @@ use k8s_openapi::api::core::v1::Node;
...
+            let bottlerocket_nodes: Api<BottlerocketNode> =
+                Api::namespaced(self.k8s_client.clone(), NAMESPACE);
+
+            Ok(bottlerocket_nodes
+                .get(&self.associated_bottlerocketnode_name.clone())
+                .await
+                .context(error::BottlerocketNodeNotExist {
+                    node_name: &self.associated_bottlerocketnode_name.clone(),
+                })
+                .is_ok())
```
cbgbt: One problem with this is that it could return `Ok(false)` under cases where the `brn` *does* exist but communication with the k8s API fails for some other reasons. I wonder if it's possible to `match` on an error variant that tells us if it doesn't exist?

cbgbt:
Looks good, thanks! Just one comment.

gthao313:
Push above address @cbgbt comments. Thanks

cbgbt (approved):

agent/src/agentclient.rs
```diff
@@ -20,90 +25,139 @@ use models::{
...
             apiserver_client,
-            node_selector: None,
+            brn_reader,
+            node_reader,
+            associated_node_name,
+            associated_bottlerocketnode_name,
         }
     }
 
     pub async fn check_node_custom_resource_exists(&mut self) -> Result<bool> {
```
somnusfish, line 65: These two name`check_custom_resource_status_exists` and `check_node_custom_resource_exists` are really confusing for me.
Seems like they use different `associated_bottlerocketnode_name` to fetch the customer resources from Kubernetes. `check_node_custom_resource_exists ` use the one binded with agent, `check_custom_resource_status_exists` use something related to "brn-{}" with value get from node selector, and the node_selector fetches stuff from `node_reader`.

Could you please explain more so I could better understand?

1. What is the difference between `brn_reader` and `node_reader`?
From my understand `brn_reader`is related to the customer resource you defined for monitor the status for each kubernetes node. Like if brupop is going to monitor 3 kubernetes hosts in the cluster, it will have 3 `BottlerocketNode` customer resources so the agent and controller can work together to update the status and spec to upgrade the host.
But I don't understand the `node_reader` here, what does that do? 

2. Could you explain the differences between `check_custom_resource_status_exists` and `check_node_custom_resource_exists`? And maybe choose a better name or add more comments on it?

agent/src/agentclient.rs
```diff
@@ -137,15 +191,15 @@ impl<T: APIServerClient> BrupopAgent<T> {
 
     /// update the custom resource associated with this node
     async fn update_metadata_custom_resource(
-        &mut self,
-        current_metadata: BottlerocketNodeStatus,
+        &self,
+        _current_metadata: BottlerocketNodeStatus,
```
somnusfish: Curious why do you add underscore here? I thought only variable that you won't use start with underscore.

agent/src/agentclient.rs
gthao313: > What is the difference between brn_reader and node_reader?

I think your understanding is correct. We watch two resources separately, One is brn (custom resource), and another one is  kubernetes node. Each agent only monitor the resources which agent currently live on. For example, agent lives on `A` node, so agent only watch `A` node and the `brn-A`( the brn created for `A` node).

> Could you explain the differences between check_custom_resource_status_exists and 
check_node_custom_resource_exists

`check_node_custom_resource_exists` means check if BottlerocketNode(custom resource) for this k8s node has been created. If not, agent needs create one. 
`check_custom_resource_status_exists` means check if BottlerocketNode(custom resource) has `Status`. If not, we need initialize it ([BottlerocketNodeStatus](https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/0.2.0/models/src/node/mod.rs#L196)). Usually a new created BottlerocketNode(custom resource) doesn't have status value. So this function helps to initialize it.

Overall those two functions help to create BottlerocketNode(custom resource) and initialize  BottlerocketNodeStatus. Yeah, I'd like to improve the name or comments. Thanks.

agent/src/agentclient.rs
gthao313: Yeah. I think I should remove it. Somehow `Rust-analyzer` suggests me to add `_`. thanks

cbgbt (approved):

agent/src/agentclient.rs
```diff
@@ -20,90 +25,139 @@ use models::{
...
+        if associated_bottlerocketnode.len() != 0 {
+            Ok(true)
+        } else {
+            let bottlerocket_nodes: Api<BottlerocketNode> =
+                Api::namespaced(self.k8s_client.clone(), NAMESPACE);
+
+            // handle the special case which custom resource does exist but communication with the k8s API fails for other errors.
+            if let Err(e) = bottlerocket_nodes
+                .get(&self.associated_bottlerocketnode_name.clone())
+                .await
```
somnusfish, line 79: Is this fetch from the reflector first and then try to get from k8s etcd? Why do you need fetch from k8s etcd again? Is it because reflector is like cache that you need miss once before the data get moved inside, in this case why does `fetch_custom_resource` only fetch from reflector and error out if failed?

gthao313: No, We tried to fetch from the reflector first. If it doesn't work because of delay on reflector . Sometime when I query `Store`, it return empty to me, which would be recognized BRN had not been created, but it just because of delay. Therefore, If delay on reflector, we will fetch k8s etcd to get the metadata. Here is the discuss with Sean https://github.com/bottlerocket-os/bottlerocket-update-operator/pull/118#discussion_r750732385

somnusfish (approved):

Issue #128, opened 2021-12-03
0.2.x: Integrate Prometheus metrics into `controller` component


cbgbt:
**Issue or Feature Request:**
We want metrics emitted when hosts are pushed through the update state machine. It would also be ideal to provide traces from the outputs `tracing` integration.

PR #132, opened 2021-12-15
Controller metrics

somnusfish:
**Issue number:**
#128

**Description of changes:**
Emitting prometheus metrics for host's bottlerocket version and host's brupop state.
Adding HTTP server in the controller to vend prometheus metrics.

**Testing done:**
Provisioned 5 old bottlerocket hosts in the cluster and run Brupop, curled the controller endpoint to verify the metrics.

Sample metrics during brupop updating hosts.
```
[ec2-user@ip-192-168-53-104 ~]$ curl 10.100.135.35/metrics
# HELP brupop_hosts_state Brupop host's state
# TYPE brupop_hosts_state gauge
brupop_hosts_state{state="Idle"} 6
brupop_hosts_state{state="MonitoringUpdate"} 1
brupop_hosts_state{state="PerformedUpdate"} 1
brupop_hosts_state{state="RebootedIntoUpdate"} 1
brupop_hosts_state{state="StagedUpdate"} 1
# HELP brupop_hosts_version Brupop host's bottlerocket version
# TYPE brupop_hosts_version gauge
brupop_hosts_version{bottlerocket_version="1.2.0"} 4
brupop_hosts_version{bottlerocket_version="1.5.1"} 2
```

Deployed prometheus on the cluster and use port forward  to get he graphic analysis locally.

**Terms of contribution:**

### DIFF
--- a/controller/Cargo.toml
+++ b/controller/Cargo.toml
@@ -15,8 +15,13 @@ kube = { version = "0.63.2", default-features = true, features = [ "derive", "ru
 
 snafu = "0.6"
 tokio = { version = "1", features = ["macros", "rt-multi-thread", "time"] }
-tracing = "0.1"
-tracing-subscriber = { version = "0.3", features = ["registry", "env-filter"] }
-tracing-bunyan-formatter = "0.3"
+
+actix-web = { version = "4.0.0-beta.9", default-features = false }
+http = "0.2.5"
 opentelemetry = { version = "0.16", features = ["rt-tokio-current-thread"] }
+opentelemetry-prometheus = "0.9"
+prometheus = "0.12.0"
+tracing = "0.1"
+tracing-bunyan-formatter = "0.3"
 tracing-opentelemetry = "0.16"
+tracing-subscriber = { version = "0.3", features = ["registry", "env-filter"] }
--- a/controller/Cargo.toml
+++ b/controller/Cargo.toml
@@ -21,6 +21,7 @@ http = "0.2.5"
 opentelemetry = { version = "0.16", features = ["rt-tokio-current-thread"] }
 opentelemetry-prometheus = "0.9"
 prometheus = "0.12.0"
+serde_plain = "1.0.0"
 tracing = "0.1"
 tracing-bunyan-formatter = "0.3"
 tracing-opentelemetry = "0.16"
--- a/controller/Cargo.toml
+++ b/controller/Cargo.toml
@@ -6,22 +6,19 @@ publish = false
 license = "Apache-2.0 OR MIT"
 
 [dependencies]
-models = { path = "../models", version = "0.1.0" }
-
+actix-web = { version = "4.0.0-beta.9", default-features = false }
 futures = "0.3"
+http = "0.2.5"
 # k8s-openapi must match the version required by kube and enable a k8s version feature
 k8s-openapi = { version = "0.13.1", default-features = false, features = ["v1_20"] }
 kube = { version = "0.63.2", default-features = true, features = [ "derive", "runtime" ] }
-
-snafu = "0.6"
-tokio = { version = "1", features = ["macros", "rt-multi-thread", "time"] }
-
-actix-web = { version = "4.0.0-beta.9", default-features = false }
-http = "0.2.5"
+models = { path = "../models", version = "0.1.0" }
 opentelemetry = { version = "0.16", features = ["rt-tokio-current-thread"] }
 opentelemetry-prometheus = "0.9"
 prometheus = "0.12.0"
 serde_plain = "1.0.0"
+snafu = "0.6"
+tokio = { version = "1", features = ["macros", "rt-multi-thread", "time"] }
 tracing = "0.1"
 tracing-bunyan-formatter = "0.3"
 tracing-opentelemetry = "0.16"
--- a/controller/src/controller.rs
+++ b/controller/src/controller.rs
@@ -82,9 +82,13 @@ impl<T: BottlerocketNodeClient> BrupopController<T> {
                 .context(error::UpdateNodeSpec);
 
             // Update metrics to guide the next operation
-            let current_operation = desired_spec.operation();
             if let Some(metrics) = &self.metrics {
-                metrics.op(current_operation);
+                if ret.is_ok() {
+                    let current_operation = desired_spec.operation();
+                    metrics.op(current_operation);
+                } else {
+                    metrics.error();
+                }
             }
 
             ret
--- a/controller/src/controller.rs
+++ b/controller/src/controller.rs
@@ -20,15 +20,19 @@ const ACTION_INTERVAL: Duration = Duration::from_secs(2);
 pub struct BrupopController<T: BottlerocketNodeClient> {
     node_client: T,
     brn_reader: Store<BottlerocketNode>,
-    metrics: Option<BrupopControllerMetrics>,
+    metrics: BrupopControllerMetrics,
 }
 
 impl<T: BottlerocketNodeClient> BrupopController<T> {
     pub fn new(node_client: T, brn_reader: Store<BottlerocketNode>) -> Self {
+        // Creates brupop-controller meter via the configured
+        // GlobalMeterProvider which is setup in PrometheusExporter
+        let meter = global::meter("brupop-controller");
+        let metrics = BrupopControllerMetrics::new(meter);
         BrupopController {
             node_client,
             brn_reader,
-            metrics: None,
+            metrics,
         }
     }
 
@@ -65,7 +69,7 @@ impl<T: BottlerocketNodeClient> BrupopController<T> {
     async fn progress_node(&self, node: BottlerocketNode) -> Result<()> {
         if node.has_reached_desired_state() {
             // Emit metrics to show the existing status
-            self.emit_metrics();
+            self.emit_metrics()?;
 
             let desired_spec = determine_next_node_spec(&node);
 
@@ -113,24 +117,17 @@ impl<T: BottlerocketNodeClient> BrupopController<T> {
     }
 
     #[instrument(skip(self))]
-    fn init_metrics(&mut self) {
-        let meter = global::meter("brupop-controller");
-        self.metrics = Some(BrupopControllerMetrics::new(meter));
-    }
-
-    #[instrument(skip(self))]
-    fn emit_metrics(&self) {
-        if let Some(metrics) = &self.metrics {
-            let data = self.fetch_data();
-            metrics.emit_metrics(data);
-        }
+    fn emit_metrics(&self) -> Result<()> {
+        let data = self.fetch_data()?;
+        self.metrics.emit_metrics(data);
+        Ok(())
     }
 
     /// Fetch the custom resources status for all resources
     /// to gather the information on hosts's bottlerocket version
     /// and brupop state.
     #[instrument(skip(self))]
-    fn fetch_data(&self) -> BrupopHostsData {
+    fn fetch_data(&self) -> Result<BrupopHostsData> {
         let mut hosts_version_count_map = HashMap::new();
         let mut hosts_state_count_map = HashMap::new();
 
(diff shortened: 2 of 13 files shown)


somnusfish:
Add controller as k8s service in this PR as well.

## Test:
Build and deploy on the local k8s cluster

### Deployment
```
➜  deploy git:(controller_metrics) ✗ kubectl get service
NAME                       TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)   AGE
brupop-apiserver           ClusterIP   10.100.27.11     <none>        80/TCP    52s
brupop-controller-server   ClusterIP   10.100.190.112   <none>        80/TCP    51s
```

### Query metrics
```
[ec2-user@ip-192-168-53-104 ~]$ curl 192.168.29.223:8080/metrics
# HELP brupop_controller_op Brupop controller operations
# TYPE brupop_controller_op counter
brupop_controller_op{operation="no_op"} 23
[ec2-user@ip-192-168-53-104 ~]$ curl 10.100.190.112/metrics
# HELP brupop_controller_op Brupop controller operations
# TYPE brupop_controller_op counter
brupop_controller_op{operation="no_op"} 29
```

controller/src/controller.rs
```diff
@@ -68,13 +72,26 @@ impl<T: BottlerocketNodeClient> BrupopController<T> {
                 "BottlerocketNode has reached desired status. Modifying spec."
             );
 
-            self.node_client
+            let ret = self
```
cbgbt: `update_result` might be a better variable name here.

controller/src/controller.rs
```diff
@@ -105,6 +122,10 @@ impl<T: BottlerocketNodeClient> BrupopController<T> {
         None
     }
 
+    fn init_metrics(&mut self) {
+        let meter = global::meter("brupop-controller");
+        self.metrics = Some(BrupopControllerMetrics::new(meter));
+    }
```
cbgbt: I think it would be best to configure the metrics as part of the constructor for the Controller, rather than initialize them at runtime.

models/src/node/mod.rs
```diff
@@ -188,6 +188,17 @@ impl BottlerocketNodeSpec {
         // We know this won't panic because we have a regex requirement on this attribute, which is enforced by the k8s schema.
         self.version.as_ref().map(|v| Version::from_str(v).unwrap())
     }
+
+    /// Returns the operation for current spec.
+    pub fn operation(&self) -> String {
```
cbgbt: Consider using [`serde_plain`](https://crates.io/crates/serde_plain) to serialize these. We also use this library in the Bottlerocket repo for enum serialization, e.g. in [buildsys](https://github.com/bottlerocket-os/bottlerocket/blob/00ecda47174e3867e5549ced93579fc2a886eadb/tools/buildsys/src/main.rs#L112)

controller/src/controller.rs
somnusfish: I tried it before. However it only works when I init meter inside run() method, neither parse meter as argument nor initialize in the constructor works.

somnusfish:
As per discussion, update the metrics from controller spec change itself to monitoring the whole state for brupop cluster.

controller/src/controller.rs
cbgbt: Can you elaborate on what went wrong? I don't understand why it wouldn't work.

models/src/node/mod.rs
```diff
@@ -195,7 +195,7 @@ impl BottlerocketNodeSpec {
 #[derive(Clone, Debug, Default, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
 pub struct BottlerocketNodeStatus {
     #[validate(regex = "SEMVER_RE")]
-    current_version: String,
+    pub current_version: String,
```
cbgbt: I don't want to allow this to be `pub` because we need to ensure that the value stored here is a valid semver string. If it's pub, someone could create a mutable `BottlerocketNodeStatus` and then create an invalid object by disregarding the semver rule.

Can we instead use `current_version().to_string()` and keep this member variable private?

controller/Cargo.toml
```diff
@@ -15,8 +15,14 @@ kube = { version = "0.63.2", default-features = true, features = [ "derive", "ru
 
 snafu = "0.6"
 tokio = { version = "1", features = ["macros", "rt-multi-thread", "time"] }
+
+actix-web = { version = "4.0.0-beta.9", default-features = false }
```
webern: nit: I would keep all of the dependencies together without newlines and sorted alphabetically.

models/src/node/mod.rs
webern: You might be able to use `SemVer` as the type or create a type to use instead of `String`, but I agree with keeping it private and using an accessor.

controller/src/controller.rs
webern: It does have some global state (obviously), but not sure why it would cause a problem. https://github.com/open-telemetry/opentelemetry-rust/blob/5d6d000c47f9f5da307e32c559bf43c40a08f951/opentelemetry/src/global/metrics.rs#L6

controller/src/main.rs
```diff
@@ -62,6 +76,9 @@ async fn main() -> Result<()> {
         _ = controller_runner => {
             event!(Level::ERROR, "controller exited");
         },
+        _ = prometheus_server => {
```
webern, line 82: Super nit: the TODO on line 71 could now change to "// TODO if any of these fail..."

webern (approved):

controller/src/metrics.rs
```diff
@@ -0,0 +1,86 @@
...
+        };
+        // Observer for cluster host's bottlerocket version
+        let _brupop_hosts_version = meter
+            .u64_value_observer("brupop_hosts_version", hosts_version_callback)
+            .with_description("Brupop host's bottlerocket version")
+            .init();
+
+        let hosts_data_clone_for_state = Arc::clone(&brupop_shared_hosts_data);
+        let hosts_state_callback = move |res: ObserverResult<u64>| {
+            let data = hosts_data_clone_for_state.lock().unwrap();
```
gthao313, line 68: I think we should avoid using `unwrap` : )

somnusfish: Do you mean we need write some error handle here?

gthao313: Yeah, we should handle the error here . You can replace `unwrap` with `method.context(error::MyCoolError)?`. Sorry about my unclear expression.

somnusfish: I was trying to handle the error here but the use case here is a little bit different.

Here is the definition for the callback clousure:
```
        let hosts_state_callback = move |res: ObserverResult<u64>| {
            let data = hosts_data_clone_for_state.lock().unwrap();
            for (host_state, count) in &data.hosts_state_count_map {
                let labels = vec![HOST_STATE_KEY.string(host_state.clone())];
                res.observe(*count, &labels);
            }
        };
```

And here is the definition for the type of the required callback:
```
pub fn u64_value_observer<T, F>(
    &self,
    name: T,
    callback: F
) -> ValueObserverBuilder<'_, u64> where
    T: Into<String>,
    F: Fn(ObserverResult<u64>) + Send + Sync + 'static,  <-------- callback type
```

I can't return something like `?` with a closure which should return ().

controller/src/controller.rs
somnusfish: The differences are when init meter inside `run()`, the meter is 
```
{"v":0,"name":"brupop-controller","msg":"[NEW - END]","level":30,"hostname":"brupop-controller-deployment-54cb49c8c6-flsrt","pid":1,"time":"2022-01-12T18:41:24.716404272+00:00","target":"controller::metrics","line":45,"file":"controller/src/metrics.rs","meter":"Meter { instrumentation_library: InstrumentationLibrary { name: \"brupop-controller\", version: None }, core: UniqueInstrumentMeterCore { inner: Accumulator(AccumulatorCore { current: {}, async_instruments: Mutex { data: AsyncInstrumentState { runners: [], instruments: [] }, poisoned: false, .. }, current_epoch: AtomicNumber(0), processor: BasicProcessor { aggregator_selector: Histogram([0.5, 0.9, 0.99]), export_selector: Cumulative, state: Mutex { data: BasicProcessorState { config: BasicProcessorConfig { memory: true }, values: {}, process_start: SystemTime { tv_sec: 1642012884, tv_nsec: 715485442 }, interval_start: SystemTime { tv_sec: 1642012884, tv_nsec: 715485513 }, interval_end: SystemTime { tv_sec: 1642012884, tv_nsec: 715485573 }, started_collection: 0, finished_collection: 0 }, poisoned: false, .. } }, resource: Resource { attrs: {} } }), sync_state: Mutex { data: {}, poisoned: false, .. }, async_state: Mutex { data: {}, poisoned: false, .. } } }","elapsed_milliseconds":0}
```

Whereas when init meter in the constructor/main the meter is 
```
{"v":0,"name":"brupop-controller","msg":"[NEW - START]","level":30,"hostname":"brupop-controller-deployment-54cb49c8c6-njh5m","pid":1,"time":"2022-01-12T19:05:54.096074979+00:00","target":"controller::metrics","line":48,"file":"controller/src/metrics.rs","meter":"Meter { instrumentation_library: InstrumentationLibrary { name: \"brupop-controller\", version: None }, core: NoopMeterCore { _private: () } }"}
```

somnusfish: I don't find an good example on how to set up for `UniqueInstrumentMeterCore` manually,  it seems can be done via the [registry](https://github.com/open-telemetry/opentelemetry-rust/blob/ccc8fb210c046856825e283940850478521c9c8b/opentelemetry/src/metrics/registry.rs) , but only example I found [here](https://github.com/open-telemetry/opentelemetry-rust/blob/02ddf1b9eac1ead48510504fc6bd4cdccb7e23fb/opentelemetry/src/sdk/metrics/controllers/push.rs#L174-L182)  seems overkilled.

controller/src/metrics.rs
webern: If the closure returns (), then we have to decide whether it is better to panic and crash the program, or to log the error and let the program continue as if nothing bad happened. I guess those are the only two choices!

controller/src/controller.rs
```diff
@@ -105,6 +112,43 @@ impl<T: BottlerocketNodeClient> BrupopController<T> {
...
+        let mut hosts_state_count_map = HashMap::new();
+
+        for brn in self.all_nodes() {
+            if let Some(brn_status) = brn.status {
+                let current_version = brn_status.current_version().to_string();
+                let current_state = brn_status.current_state;
+
+                *hosts_version_count_map.entry(current_version).or_default() += 1;
+                *hosts_state_count_map
+                    .entry(serde_plain::to_string(&current_state).unwrap())
```
cbgbt: Nit: We know this can never actually return an error unless we have made a programming mistake, since the enum should always be serializable. I've had some conversations with @webern about this and I think it might be best to still transform this into a `Result` type, where the `Err` variant has something like `Assertion` or some other name that implies it can only occur when the software is wrong, rather than when there's some transient error.

Happy to chat more about this!

controller/src/metrics.rs
cbgbt: It turns out that `Mutex` returns an error on `lock()` [only when it is poisoned](https://doc.rust-lang.org/std/sync/struct.Mutex.html#poisoning), meaning another thread holding the lock has paniced. I think in that scenario, it is acceptable to panic.

controller/src/main.rs
```diff
@@ -54,14 +57,28 @@ async fn main() -> Result<()> {
             futures::future::ready(())
         });
 
-    // TODO if either of these fails, we should write to the k8s termination log and exit.
+    // Setup Http server to vend prometheus metrics
+    let exporter = opentelemetry_prometheus::exporter().init();
+    let prometheus_server = HttpServer::new(move || {
+        App::new()
+            .app_data(Data::new(exporter.clone()))
```
cbgbt, line 67: nit: I think this `clone()` might be unnecessary?

cbgbt (approved):

controller/src/controller.rs
cbgbt: I'm having a hard time working out why this would happen.

controller/src/main.rs
somnusfish: It would get the following error.
```
error[E0507]: cannot move out of `exporter`, a captured variable in an `Fn` closure
  --> controller/src/main.rs:64:33
   |
61 |     let exporter = opentelemetry_prometheus::exporter().init();
   |         -------- captured outer variable
...
64 |             .app_data(Data::new(exporter))
   |                                 ^^^^^^^^ move occurs because `exporter` has type `PrometheusExporter`, which does not implement the `Copy` trait
```

controller/src/controller.rs
somnusfish: Found out why, comments in latest change.

controller/src/controller.rs
somnusfish: Fixed

cbgbt (approved):

cbgbt (approved):

gthao313 (approved):

PR #155, opened 2022-02-11
Refactor agent run to support crash loop handle

somnusfish:
**Issue number:**

**Description of changes:**
Refactor the agent run method with a wrapper over Bottlerocket Update API call for later support crash loop handle. Changes include:
* Make agent readonly
* Abstract few methods in run()

**Testing done:**
Provisioned two hosts with old bottlerocket version and installed Brupop. Monitored the update on the hosts.

**Terms of contribution:**

### DIFF
--- a/agent/src/agentclient.rs
+++ b/agent/src/agentclient.rs
@@ -65,9 +65,11 @@ impl<T: APIServerClient> BrupopAgent<T> {
     }
 
     #[instrument(skip(self), err)]
-    pub async fn check_node_custom_resource_exists(&mut self) -> Result<bool> {
-        // try to check if node custom resource exist in the store first. If it's not present in the
-        // store(either node custom resource doesn't exist or store data delays), make the API call for second round check.
+    async fn check_node_shadow_exists(
+        &self,
+    ) -> std::result::Result<bool, agentclient_error::BottlerocketShadowRWError> {
+        // try to check if node associated BottlerocketShadow exist in the store first. If it's not present in the
+        // store(either associated BottlerocketShadow doesn't exist or store data delays), make the API call for second round check.
 
         let associated_bottlerocketshadow = self.brs_reader.state().clone();
         if associated_bottlerocketshadow.len() != 0 {
@@ -76,13 +78,13 @@ impl<T: APIServerClient> BrupopAgent<T> {
             let bottlerocket_shadows: Api<BottlerocketShadow> =
                 Api::namespaced(self.k8s_client.clone(), NAMESPACE);
 
-            // handle the special case which custom resource does exist but communication with the k8s API fails for other errors.
+            // handle the special case which associated BottlerocketShadow does exist but communication with the k8s API fails for other errors.
             if let Err(e) = bottlerocket_shadows
                 .get(&self.associated_bottlerocketshadow_name.clone())
                 .await
             {
                 match e {
-                    // 404 not found response error is OK for this use, which means custom resource doesn't exist
+                    // 404 not found response error is OK for this use, which means associated BottlerocketShadow doesn't exist
                     kube::Error::Api(error_response) => {
                         if error_response.code == 404 {
                             return Ok(false);
@@ -93,7 +95,7 @@ impl<T: APIServerClient> BrupopAgent<T> {
                             .fail();
                         }
                     }
-                    // Any other type of errors can not present that custom resource doesn't exist, need return error
+                    // Any other type of errors can not present that associated BottlerocketShadow doesn't exist, need return error
                     _ => {
                         return Err(e).context(agentclient_error::UnableFetchBottlerocketShadow {
                             node_name: &self.associated_bottlerocketshadow_name.clone(),
@@ -106,8 +108,8 @@ impl<T: APIServerClient> BrupopAgent<T> {
     }
 
     #[instrument(skip(self), err)]
-    pub async fn check_custom_resource_status_exists(&self) -> Result<bool> {
-        Ok(self.fetch_custom_resource().await?.status.is_some())
+    async fn check_shadow_status_exists(&self) -> Result<bool> {
+        Ok(self.fetch_shadow().await?.status.is_some())
     }
 
     #[instrument(skip(self), err)]
@@ -136,15 +138,15 @@ impl<T: APIServerClient> BrupopAgent<T> {
         .await
     }
 
-    // fetch associated BottlerocketShadow (custom resource) and help to get node `metadata`, `spec`, and  `status`.
+    // fetch associated BottlerocketShadow and help to get node `metadata`, `spec`, and  `status`.
     #[instrument(skip(self), err)]
-    async fn fetch_custom_resource(&self) -> Result<BottlerocketShadow> {
-        // get associated BottlerocketShadow (custom resource) name, and we can use this
-        // BottlerocketShadow name to fetch targeted BottlerocketShadow (custom resource) and get node `metadata`, `spec`, and  `status`.
+    async fn fetch_shadow(&self) -> Result<BottlerocketShadow> {
+        // get associated BottlerocketShadow name, and we can use this
+        // BottlerocketShadow name to fetch targeted BottlerocketShadow and get node `metadata`, `spec`, and  `status`.
 
         Retry::spawn(retry_strategy(), || async {
             // Agent specifies node reflector only watch and cache the BottlerocketShadow which is associated with the node that agent pod currently lives on,
-            // so vector of brs_reader only have one object. Therefore, fetch_custom_resource uses index 0 to extract BottlerocketShadow object.
+            // so vector of brs_reader only have one object. Therefore, fetch_shadow uses index 0 to extract BottlerocketShadow object.
             let associated_bottlerocketshadow = self.brs_reader.state().clone();
             if associated_bottlerocketshadow.len() != 0 {
                 return Ok(associated_bottlerocketshadow[0].clone());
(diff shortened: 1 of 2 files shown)


agent/src/agentclient.rs
```diff
@@ -257,45 +257,133 @@ impl<T: APIServerClient> BrupopAgent<T> {
...
+    async fn create_custom_resource_if_not_exist(&self) -> Result<()> {
+        let custom_resource_exists = self.check_node_custom_resource_exists().await?;
+        if !custom_resource_exists {
+            self.create_metadata_custom_resource().await?;
+        }
+        Ok(())
+    }
+
+    #[instrument(skip(self))]
+    async fn initialize_custom_resource_if_not_initialzed(&self) -> Result<()> {
```
cbgbt: nit: typo, plus suggestion to improve clarity.

```suggestion
    async fn initialize_shadow_if_not_initialized(&self) -> Result<()> {
```

cbgbt: I wonder if it generally makes sense to refer to the `custom_resource` as the `shadow` in the agent. We avoided it at first because `bottlerocket_node` was ambiguous, but now we have an unambiguous name to refer to.

agent/src/agentclient.rs
```diff
@@ -312,119 +400,48 @@ impl<T: APIServerClient> BrupopAgent<T> {
...
+                Err(e) => {
+                    match e {
+                        // Errors from update_status_in_custom_resource should be skipped and retry in next loop
+                        agentclient_error::Error::MissingBottlerocketShadowStatus
+                        | agentclient_error::Error::BottlerocketShadowStatusVersion { source: _ }
+                        | agentclient_error::Error::BottlerocketShadowStatusChosenUpdate {
+                            source: _,
+                        }
+                        | agentclient_error::Error::UpdateBottlerocketShadowResource {
+                            source: _,
```
cbgbt: I'm worried about this list becoming stale. Can we reclassify errors from that function using a new `Error` type, that way we can just refer to the overarching type rather than having to disambiguate this way?

agent/src/agentclient.rs
```diff
@@ -312,119 +400,48 @@ impl<T: APIServerClient> BrupopAgent<T> {
...
-                                sleep(AGENT_SLEEP_DURATION).await;
-                                continue;
-                            };
-                        } else {
-                            boot_update()
-                                .await
-                                .context(agentclient_error::UpdateActions {
-                                    action: "Reboot".to_string(),
-                                })?;
+            match self.handle_update(&bottlerocket_shadow).await {
```
cbgbt: nit: We should pick a more particular name here, maybe `handle_state_transition`?

cbgbt (changes requested):

agent/src/agentclient.rs
somnusfish: I think I can create a new Error enum related to operate Shadow and put it under agentclient_error::Error. I will do that.

agent/src/agentclient.rs
```diff
@@ -257,45 +257,133 @@ impl<T: APIServerClient> BrupopAgent<T> {
         Ok(())
     }
 
+    #[instrument(skip(self))]
+    async fn create_custom_resource_if_not_exist(&self) -> Result<()> {
```
gthao313: According to Sean's comment below, I think might be better to be consistent to use `shadow` instead of `custom resource`.
```suggestion
    async fn create_shadow_if_not_exist(&self) -> Result<()> {
```

agent/src/agentclient.rs
gthao313: Yeah, Same feeling here. And another concern would be if we have more errors needed to be handled in future, this  part has to be updated often and become lengthy. So that might be worth to pack it like you mentioned above create a new Error enum.

agent/src/agentclient.rs
```diff
@@ -65,7 +65,7 @@ impl<T: APIServerClient> BrupopAgent<T> {
     }
 
     #[instrument(skip(self), err)]
-    pub async fn check_node_custom_resource_exists(&mut self) -> Result<bool> {
+    pub async fn check_node_custom_resource_exists(&self) -> Result<bool> {
```
gthao313:
```suggestion
    pub async fn check_shadow_exists(&self) -> Result<bool> {
```

agent/src/agentclient.rs
```diff
@@ -185,7 +185,7 @@ impl<T: APIServerClient> BrupopAgent<T> {
 
     /// create the custom resource associated with this node
     #[instrument(skip(self), err)]
-    pub async fn create_metadata_custom_resource(&mut self) -> Result<()> {
+    pub async fn create_metadata_custom_resource(&self) -> Result<()> {
```
gthao313:
```suggestion
    pub async fn create_shadow_resource(&self) -> Result<()> {
```

agent/src/agentclient.rs
```diff
@@ -65,9 +65,11 @@ impl<T: APIServerClient> BrupopAgent<T> {
     }
 
     #[instrument(skip(self), err)]
-    pub async fn check_node_custom_resource_exists(&mut self) -> Result<bool> {
-        // try to check if node custom resource exist in the store first. If it's not present in the
-        // store(either node custom resource doesn't exist or store data delays), make the API call for second round check.
+    async fn check_node_shadow_exists(
+        &self,
+    ) -> std::result::Result<bool, agentclient_error::BottlerocketShadowRWError> {
```
somnusfish, line 70: I was struggling to return agentclient_error::Error but somehow it's very hard to implement that. Since this is a private method and the Err will later be turned into agentclient_error::Error, I would assume it is okay to just return agentclient_error::BottlerocketShadowRWError here.

somnusfish:
Changes:
* Address comments to use BottlerocketShadow/shadow for custom resources.
* Fix typo.
* Add `agentclient_error::BottlerocketRWError` to better aggregate errors
* Make most method in agentclient.rs privdate since they are not used outside.

agent/src/agentclient.rs
```diff
@@ -257,52 +258,137 @@ impl<T: APIServerClient> BrupopAgent<T> {
         Ok(())
     }
 
+    #[instrument(skip(self))]
+    async fn create_shadow_if_not_exist(&self) -> Result<()> {
```
etungsten, line 262: Nit:
```suggestion
    async fn create_shadow(&self) -> Result<()> {
```

I'm not sure if it necessary to include the condition in the function name. Unless there is a different `create_shadow` function that creates a shadow regardless of whether one already exists

agent/src/agentclient.rs
```diff
@@ -257,52 +258,137 @@ impl<T: APIServerClient> BrupopAgent<T> {
...
+        let shadow_exists = self.check_node_shadow_exists().await?;
+        if !shadow_exists {
+            self.create_metadata_shadow().await?;
+        }
+        Ok(())
+    }
+
+    #[instrument(skip(self))]
+    async fn initialize_shadow_if_not_initialized(&self) -> Result<()> {
```
etungsten, line 271:
```suggestion
    async fn initialize_shadow(&self) -> Result<()> {
```

Same idea as above

agent/src/agentclient.rs
```diff
@@ -257,52 +258,137 @@ impl<T: APIServerClient> BrupopAgent<T> {
...
+        // - Gather metadata about the system and update the BottlerocketShadow associated with this node
+        // - Determine if the spec on the system's BottlerocketShadow demands the node take action. If so, begin taking that action.
 
         loop {
-            // Create a BottlerocketShadow (custom resource) if associated BottlerocketShadow does not exist
-            if self.check_node_custom_resource_exists().await.is_err() {
-                // Errors checking if brs exists are ignored (and also logged by `check_node_custom_resource_exists()`).
-                event!(Level::WARN, "An error occurred when checking if BottlerocketShadow exists. Restarting event loop");
+            if self.create_shadow_if_not_exist().await.is_err() {
+                event!(Level::WARN, "An error occurred when try to create BottlerocketShadow BottlerocketShadow. Restarting event loop.");
```
etungsten: nit:
```suggestion
                event!(Level::WARN, "An error occurred when try to create BottlerocketShadow. Restarting event loop.");
```

agent/src/agentclient.rs
```diff
@@ -508,14 +488,52 @@ pub mod agentclient_error {
             source: apiclient_error::Error,
         },
 
+        #[snafu(display("Unable to operate on BottlerocketShadow for {}", error))]
```
etungsten: nit:
```suggestion
        #[snafu(display("Unable to operate on BottlerocketShadow: '{}'", error))]
```

agent/src/agentclient.rs
```diff
@@ -508,14 +488,52 @@ pub mod agentclient_error {
...
+    impl From<BottlerocketShadowRWError> for Error {
+        fn from(err: BottlerocketShadowRWError) -> Self {
+            Self::BottlerocketShadowError { error: err }
+        }
+    }
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility = "pub")]
+    pub enum BottlerocketShadowRWError {
+        #[snafu(display("Unable to gather system version metadata: {}", source))]
```
etungsten: Nit: we single quote the source error in other error variants
```suggestion
        #[snafu(display("Unable to gather system version metadata: '{}'", source))]
```

etungsten (commented):
Some small nits. Otherwise looks good to me.

agent/src/agentclient.rs
somnusfish: In the `run` loop, these two methods are called at the beginning of the loop. I feel with the condition in the name it would be more easy to understand.

agent/src/agentclient.rs
somnusfish: Same explaination above.

etungsten (approved):

cbgbt (approved):

Issue #123, opened 2021-11-23
0.2.0: Handle update-reboot failures/ "crash loops"


cbgbt:
If an update-reboot fails and is forced to rollback to the previous partition, brupop will continue to attempt to add the node to the active set, updating it endlessly and needlessly.

We should record attempted updates (or more generically, attempted state transitions) and give up after some number of attempts, recording a failure metric.

cbgbt:
This is merged, but we want to use the migrations work in #161 to ensure a migration exists from existing 0.2.0 CRD. We'll release this in another pre-GA release.

Issue #12, opened 2019-11-11
Add delay and smarter verification between node restarts


jahkeup:
**What I'd like:**

Dogswatch should add some delay between the restart of Nodes in a cluster. During this time, the Controller should check in with the Node that has been updated to confirm that it has come up healthy and that Workloads have returned to it. After this, the Controller should have a configurable duration used to delay between each Node restart.

samuelkarp:
This seems potentially related to [before and after reboot checks](https://github.com/coreos/container-linux-update-operator/blob/master/doc/before-after-reboot-checks.md).

anguslees:
Indeed.  In case you want concrete use cases for before/after reboot checks, I use them (with coreos/flatcar update operator) to delay until the rook/ceph cluster is healthy[1], and to signal to rook/ceph that the rook storage cluster should set the "noout" flag[2].  After reboot clears the noout flag and again blocks until cluster is healthy again.

[1] eg: Data is replicated sufficiently.  This signal is "global" and much more complex than what a single pod readinessProbe can represent, which is why it can't be just a PodDisruptionBudget.  A better implementation might only consider the redundancy of the data on "this" node.  In particular, a naive time "delay", or generic check that pods were running again (as suggested in the issue description) would not be sufficient here.

[2] noout means the node outage that is about to happen is expected to be brief, and rook should not start frantically re-replicating "lost" data onto new nodes.

This wasn't my idea at all, the standard rook docs for this are: https://rook.io/docs/rook/v1.4/container-linux.html

Having used this for a long time now, it works great.  What might not be obvious at first is that the reboot script itself is deployed as a daemonset limited to nodes with the "before-reboot" label. That means it automatically "finds" and installs itself only on the relevant nodes, and only for the relevant time, which is pretty neat.  Debugging the system when updates are not proceeding does require an understanding of the various state machine interactions though, of course.

I would expect very similar challenges exist for something like an elasticsearch cluster, where data replication is important and also not represented in the "health" of any specific container.  I agree this probably points to a missing feature in PodDisruptionBudget, since it _is_ still fundamentally a question of "is it ok to make $pod unavailable now".

chancez:
[...]

PR #160, opened 2022-02-18
Add reset algorithm to better handle crash loop.

somnusfish:
Exponential backoff in retry with ~24 hours counter reset to prevent
the agent crashes infinitely. With each retry starting in Idle state,
this algorithm allows the agent to fetch the latest Bottlerocket for
each retry. By waiting the agent going back to Idle for recovery, this
algorithm prevents the controller brown out the cluster.

**Issue number:**
#123 

**Description of changes:**
* The agent shall signal that it has encountered an error while attempting a state transition before crashing by setting its state to ErrorReset.
* When the controller progress node, if the controller detects that the agent has crashed, it shall set the spec state to Idle, record the current time as a timestamp variable in Bottlerocket update circle, increment the crash_count variable and emit failure metrics. The controller shall wait for the agent to uncordon the node if needed and set state to Idle before moving forward.
* When find_and_update_ready_node, the controller shall tend to find and update nodes with lower crash_count.
* When progress a node crashed before, the controller shall treat the node as usual node if it has reached 2^crash_count seconds since the first crash. Otherwise, the controller shall check if it reached 24 hours for reset. If it has been more than 24 hours since the first crash, the controller shall reset crash_count to 0 and crash timestamp to None, then treat the node as usual node. Otherwise the controller shall skip.
* When the agent restarts, if the state is ErrorReset and spec is Idle, the agent shall uncordon the node based on the recorded previous state, and set the current status to Idle, else the agent shall try to process as usual.
* Upon a successful Bottlerocket update circle, the crash_count variable shall be reset to 0 and crash time stamp shall be set to None.

**Testing done:**
#### Without crash
Provisioned 3 old hosts with Brupop installed, monitored cluster's update.

#### With crash
Intentionally make `update()` error to make sure StagedUpdate->PerformedUpdate always fails. Montoring the cluster to confirm the following feature:
* Agent set the status to ErrorReset when something failed
* Controller guide the agent to recover
* Controller tend to pick the node with lower crash time
* Controller retry using exponential backoff

#### Not covered in test
* Controller reset the crash ~24 hours since the first crash.

**Terms of contribution:**

### DIFF
--- a/agent/src/agentclient.rs
+++ b/agent/src/agentclient.rs
@@ -1,15 +1,3 @@
-use k8s_openapi::api::core::v1::Node;
-use kube::Api;
-use snafu::{OptionExt, ResultExt};
-use tokio::time::{sleep, Duration};
-use tokio_retry::{
-    strategy::{jitter, ExponentialBackoff},
-    Retry,
-};
-use tracing::{event, instrument, Level};
-
-use kube::runtime::reflector::Store;
-
 use crate::apiclient::{boot_update, get_chosen_update, get_os_info, prepare, update};
 use apiserver::{
     client::APIServerClient,
@@ -24,6 +12,18 @@ use models::{
     },
 };
 
+use chrono::{DateTime, Utc};
+use k8s_openapi::api::core::v1::Node;
+use kube::runtime::reflector::Store;
+use kube::Api;
+use snafu::{OptionExt, ResultExt};
+use tokio::time::{sleep, Duration};
+use tokio_retry::{
+    strategy::{jitter, ExponentialBackoff},
+    Retry,
+};
+use tracing::{event, instrument, Level};
+
 // The reflector uses exponential backoff.
 // These values configure how long to delay between tries.
 const RETRY_BASE_DELAY: Duration = Duration::from_millis(1000);
@@ -34,6 +34,21 @@ const AGENT_SLEEP_DURATION: Duration = Duration::from_secs(5);
 
 /// The module-wide result type.
 pub type Result<T> = std::result::Result<T, agentclient_error::Error>;
+
+#[derive(Clone, Default, Debug)]
+struct ShadowErrorInfo {
+    pub crash_count: u32,
+    pub state_transition_failure_timestamp: Option<DateTime<Utc>>,
+}
+
+impl ShadowErrorInfo {
+    fn new(crash_count: u32, state_transition_failure_timestamp: Option<DateTime<Utc>>) -> Self {
+        ShadowErrorInfo {
+            crash_count,
+            state_transition_failure_timestamp,
+        }
+    }
+}
 
 #[derive(Clone)]
 pub struct BrupopAgent<T: APIServerClient> {
@@ -162,9 +177,10 @@ impl<T: APIServerClient> BrupopAgent<T> {
 
     /// Gather metadata about the system, node status provided by spec.
     #[instrument(skip(self, state), err)]
-    async fn gather_system_metadata(
+    async fn shadow_status_with_refreshed_system_matadata(
         &self,
         state: BottlerocketShadowState,
+        shadow_error_info: ShadowErrorInfo,
     ) -> Result<BottlerocketShadowStatus> {
         let os_info = get_os_info()
             .await
@@ -182,6 +198,8 @@ impl<T: APIServerClient> BrupopAgent<T> {
             os_info.version_id.clone(),
             update_version,
             state,
+            shadow_error_info.crash_count,
+            shadow_error_info.state_transition_failure_timestamp,
         ))
     }
 
@@ -223,7 +241,10 @@ impl<T: APIServerClient> BrupopAgent<T> {
     #[instrument(skip(self), err)]
     async fn initialize_metadata_shadow(&self) -> Result<()> {
         let update_node_status = self
-            .gather_system_metadata(BottlerocketShadowState::Idle)
+            .shadow_status_with_refreshed_system_matadata(
+                BottlerocketShadowState::Idle,
+                ShadowErrorInfo::default(),
+            )
             .await?;
 
         self.update_metadata_shadow(update_node_status).await?;
@@ -280,20 +301,30 @@ impl<T: APIServerClient> BrupopAgent<T> {
     async fn update_status_in_shadow(
         &self,
         bottlerocket_shadow: &BottlerocketShadow,
+        state: BottlerocketShadowState,
+        shadow_error_info: ShadowErrorInfo,
     ) -> Result<()> {
         let bottlerocket_shadow_status = bottlerocket_shadow
             .status
             .as_ref()
             .context(agentclient_error::MissingBottlerocketShadowStatus)?;
-        let bottlerocket_shadow_spec = &bottlerocket_shadow.spec;
 
         let updated_node_status = self
-            .gather_system_metadata(bottlerocket_shadow_spec.state.clone())
+            .shadow_status_with_refreshed_system_matadata(state, shadow_error_info)
             .await?;
 
         if updated_node_status != *bottlerocket_shadow_status {
             self.update_metadata_shadow(updated_node_status).await?;
         }
+        Ok(())
+    }
+
+    /// Prepare the node to be ready to work from rebooting or crashing.
+    #[instrument(skip(self))]
+    async fn handle_recover(&self) -> Result<()> {
+        // This recover logic might need to be improved in future based on adding
+        // more features when agent drain the node.
+        self.uncordon().await?;
         Ok(())
     }
 
@@ -318,12 +349,19 @@ impl<T: APIServerClient> BrupopAgent<T> {
             );
 
             match bottlerocket_shadow_spec.state {
-                BottlerocketShadowState::Idle => {
-                    event!(
-                        Level::INFO,
-                        "Ready to finish monitoring and start update process"
-                    );
-                }
+                // Right now we consider the recover state for ErrorReset is Idle
+                BottlerocketShadowState::Idle => match bottlerocket_shadow_status.current_state {
+                    BottlerocketShadowState::ErrorReset => {
+                        self.handle_recover().await?;
+                        event!(Level::INFO, "Recovered from ErrorReset");
+                    }
+                    _ => {
+                        event!(
+                            Level::INFO,
+                            "Ready to finish monitoring and start update process"
+                        );
+                    }
+                },
                 BottlerocketShadowState::StagedUpdate => {
                     event!(Level::INFO, "Preparing update");
                     prepare().await.context(agentclient_error::UpdateActions {
@@ -343,7 +381,15 @@ impl<T: APIServerClient> BrupopAgent<T> {
 
                     if running_desired_version(bottlerocket_shadow_spec).await? {
                         // Make sure the status in BottlerocketShadow is up-to-date from the reboot
-                        self.update_status_in_shadow(bottlerocket_shadow).await?;
+                        self.update_status_in_shadow(
+                            bottlerocket_shadow,
+                            bottlerocket_shadow_spec.state.clone(),
+                            ShadowErrorInfo::new(
+                                bottlerocket_shadow_status.crash_count(),
+                                bottlerocket_shadow_status.failure_timestamp().unwrap(),
+                            ),
+                        )
+                        .await?;
                     } else {
                         boot_update()
                             .await
@@ -354,13 +400,21 @@ impl<T: APIServerClient> BrupopAgent<T> {
                 }
                 BottlerocketShadowState::MonitoringUpdate => {
                     event!(Level::INFO, "Monitoring node's healthy condition");
-                    self.uncordon().await?;
+                    self.handle_recover().await?;
                     // TODO: we need add some criterias here by which we decide to transition
                     // from MonitoringUpdate to WaitingForUpdate.
                 }
+                BottlerocketShadowState::ErrorReset => {
+                    // Spec state should never be ErrorReset
+                    event!(Level::ERROR, "Spec state should never be ErrorReset");
+                    return agentclient_error::Assertion {
+                        message: "Spec state should never be ErrorReset.".to_string(),
+                    }
+                    .fail();
+                }
             }
         } else {
-            event!(Level::INFO, "Did not detect action demand.");
+            event!(Level::DEBUG, "Did not detect action demand.");
         }
         Ok(())
     }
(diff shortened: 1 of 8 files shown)


agent/src/agentclient.rs
```diff
@@ -297,6 +297,27 @@ impl<T: APIServerClient> BrupopAgent<T> {
         Ok(())
     }
 
+    #[instrument(skip(self))]
+    async fn update_failure_status_in_shadow(
```
cbgbt: Maybe instead of replicating the update function we can just change it to take a desired `BottlerocketShadowState` as an additional parameter?

agent/src/agentclient.rs
```diff
@@ -318,11 +339,24 @@ impl<T: APIServerClient> BrupopAgent<T> {
...
-                        "Ready to finish monitoring and start update process"
-                    );
+                    match bottlerocket_shadow_status.current_state {
+                        BottlerocketShadowState::ErrorReset => {
+                            // The state before crashing is ignored since there won't be any
+                            // issue to uncordon an already uncordoned node.
+                            // This recover logic might need to be improved in future based on adding
+                            // more features when agent drain the node.
+                            self.uncordon().await?;
+                            event!(Level::INFO, "Recover from ErrorReset");
```
cbgbt: I wonder if we can pull `uncordon` out into some abstraction so that the same "make node ready for work" code is used in case of `ErrorReset` or also if we're just coming back up after reboot?

This would make it less likely that we need to modify multiple places to add more update prep behavior. We'd just want to document on that function that it's used in cases of monitoring successful updates or recovering from an error.

agent/src/agentclient.rs
```diff
@@ -358,6 +393,10 @@ impl<T: APIServerClient> BrupopAgent<T> {
                     // TODO: we need add some criterias here by which we decide to transition
                     // from MonitoringUpdate to WaitingForUpdate.
                 }
+                BottlerocketShadowState::ErrorReset => {
+                    // Spec should never be ErrorReset
+                    panic!("Spec state should never be ErrorReset");
+                }
```
cbgbt: I agree that `panic!` makes sense since this state indicates that the software is broken. We'd been handling cases like this elsewhere by making an `Assertion` error variant that we return, which makes sure that the software bubbles up through usual error handling code (like writing the termination log).

agent/src/agentclient.rs
```diff
@@ -399,40 +438,48 @@ impl<T: APIServerClient> BrupopAgent<T> {
...
                         agentclient_error::Error::BottlerocketShadowError { error: _ } => {
                             event!(Level::WARN, "An error occurred when updating BottlerocketShadow status. Restarting event loop.");
-                            sleep(AGENT_SLEEP_DURATION).await;
-                            continue;
                         }
                         _ => {
-                            // TODO Handle agent crash loop by Lu
                             event!(
-                                Level::ERROR,
+                                Level::WARN,
```
cbgbt, line 505: I wonder if `Level::ERROR` might still be better?

controller/src/controller.rs
```diff
@@ -86,6 +88,32 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
...
                 .await
                 .context(error::UpdateNodeSpec)
+        } else if node.has_crashed() {
+            self.emit_metrics()?;
+
+            let desired_spec = determine_next_node_spec(&node);
+
+            event!(
+                Level::INFO,
+                "BottlerocketShadow's status is ErrorReset, try to recover from crash."
```
cbgbt: nit:

```suggestion
                "BottlerocketShadow's status is ErrorReset, trying to recover from crash."
```

controller/src/controller.rs
```diff
@@ -86,6 +88,32 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
...
+                )
+                .await
+                .context(error::UpdateNodeSpec)?;
+
+            // Rely on active_node_set to determine whether the agent
+            // change status back to Idle
+            while !self.active_node_set().is_empty() {
+                event!(Level::INFO, "Wait for the agent to recover");
+                sleep(AGENT_RECOVERY_INTERVAL).await;
+            }
```
cbgbt: I don't think we should sleep loop here, because it could mean that the controller ceases to emit metrics for the duration of time that this node attempts to recover.

If we consider this `Error` node to be part of the active set, then allowing the controller to rerun it's event loop should still proceed correctly, not adding additional nodes to the loop until this one has resolved.

controller/src/statemachine.rs
```diff
@@ -36,14 +83,27 @@ pub fn determine_next_node_spec(brs: &BottlerocketShadow) -> BottlerocketShadowS
...
                     brs.spec.version(),
+                    brs.spec.crash_count(),
+                    brs.spec.failure_timestamp().unwrap(),
                 ),
             }
         }
     }
 }
+
+fn calculated_time_since_crash_in_minutes(brs: &BottlerocketShadow) -> Option<i64> {
```
cbgbt: Would it make sense to return a `Duration` here, and allow the caller to boil it down to an integer type?

controller/src/statemachine.rs
```diff
@@ -10,24 +13,68 @@ pub fn determine_next_node_spec(brs: &BottlerocketShadow) -> BottlerocketShadowS
...
-                        .filter(|target_version| &node_status.current_version() != target_version)
-                        .map(|target_version| {
+                    if node_status.current_version() != target_version {
+                        if let Some(crash_time_period) = calculated_time_since_crash_in_minutes(brs)
+                        {
+                            // Node crashed before
+                            event!(Level::INFO, since_last_crash = ?crash_time_period, "Detected gap since last crash");
+                            if crash_time_period
+                                > 2_i64.pow(brs.spec.crash_count().try_into().unwrap())
+                            {
```
cbgbt: Can we pull the exponential backoff code into a function and add unit tests?

controller/src/statemachine.rs
```diff
@@ -10,24 +13,68 @@ pub fn determine_next_node_spec(brs: &BottlerocketShadow) -> BottlerocketShadowS
...
+                                BottlerocketShadowSpec::new_starting_now(
+                                    BottlerocketShadowState::StagedUpdate,
+                                    Some(target_version.clone()),
+                                    brs.spec.crash_count(),
+                                    brs.spec.failure_timestamp().unwrap(),
+                                )
+                            } else if crash_time_period > 60 * 12 {
```
cbgbt: We should replace these integers with `CONSTANTS`.

controller/src/statemachine.rs
```diff
@@ -10,24 +13,68 @@ pub fn determine_next_node_spec(brs: &BottlerocketShadow) -> BottlerocketShadowS
...
+                                // Reset after ~ 24 hours since the first ErrorReset
+                                BottlerocketShadowSpec::new_starting_now(
```
cbgbt: I wonder if we actually want to reset on a timer? It might make sense to reset only when the node successfully updates. This would give us a stronger suspicion that the node will succeed.

models/src/node/mod.rs
```diff
@@ -148,26 +157,50 @@ impl BottlerocketShadow {
...
+    }
 }
 
 impl BottlerocketShadowSpec {
     pub fn new(
         state: BottlerocketShadowState,
         state_transition_timestamp: Option<DateTime<Utc>>,
         version: Option<Version>,
+        crash_count: u64,
+        state_transition_failure_timestamp: Option<DateTime<Utc>>,
```
cbgbt: I'm thinking it might actually make more sense to include `crash_count` in the `status`, since it has more to do with how the node is currently progressing rather than its desired state.

We may need the nodes themselves to be responsible for updating the crash count, which would help ensure that we don't double-count crashes, but I'm not 100% on this.

I was hoping to draw inspiration from how the `restarts` are stored for Pods in Kubernetes itself, but it seems like they might not be associated directly with `Pod` objects. Still trying to work that out.

cbgbt (changes requested):

agent/src/agentclient.rs
somnusfish: I might not understand what you meant. Do you mean something similar to 
```
    let token_path = token_path.to_str().context(error::Assertion {
        message: "Token path (defined in models/agent.rs) is not valid unicode.",
    })?;
```
But this just gives out an error to let the caller handle. In this case, we want the program to panic since this should not happen. The difference would just be panic here or the caller panic later.

agent/src/agentclient.rs
somnusfish: All Errs can be handled by retry are `Level::WARN` in `agentclient.rs`. I would like to make it uniformed.

controller/src/statemachine.rs
somnusfish: I will do that.

controller/src/statemachine.rs
somnusfish: I feel it would be better to reset the timer. The exponential backoff escalate pretty fast that once it reaches 1 day, it would go 2 days, 4 days, etc. Then Brupop need wait that long to fetch a fixed Bottlerocket.

If we decided without the reset, we should at least limit the max wait time to some value say maybe 24 hours so we don't miss the correct update too long.

controller/src/statemachine.rs
somnusfish: The reason that I use the `Option` here was I want to simplify the logic to check crash/not crash. I feel return i64 rather than Duration would make it caller's logic easier. Otherwise the caller would need to compare `crash_time_period> Duration::minutes(2_i64.pow(brs.spec.crash_count().try_into().unwrap()))`. 

I will decide which to use with the implementation for exponential backoff  as a separate function.

models/src/node/mod.rs
somnusfish: The reasons that `crash_count` is in the spec are:
1. We have the reset logic that requires the controller to reset the crash_count. And since the time calculate logic is already in the controller, we don't want the agent repeat the time calculate logic for the reset.

2. In a successful update, `crash_count` gets reset during the state transfer MonitoringUpdate -> Idle. On the agent perspective, if the target state is Idle, the agent won't know whether it is from ErrorReset or MonitoringUpdate, which will requires us to track the previous state in the status.

controller/src/controller.rs
somnusfish: In the existing logic that crash_count is in the spec, the controller was in charge of incrementing the crash_count once it sees the node in ErrorReset, so letting it go to another loop would cause false incrementing the crash_count.

>I don't think we should sleep loop here, because it could mean that the controller ceases to emit metrics for the duration of time that this node attempts to recover.

In the previous controller logic, the controller only emits metrics once it detects the node reached the desired state, which means the controller also ceases to emit metrics for the duration of time the node work on state transfer.

agent/src/agentclient.rs
cbgbt: Yeah. I think the idea is to have the caller handle gracefully terminating the program, since our outer error handler is writing to the termination log it will be properly logged. If we let the program panic here, it will be really difficult to determine why it failed.

agent/src/agentclient.rs
cbgbt: After thinking about it a bit more, I agree that you're right to use `WARN`.

controller/src/statemachine.rs
cbgbt: I think I would prefer limiting to a maximum backoff time of 24 hours, that seems good.

controller/src/statemachine.rs
cbgbt: Yeah. I think abstracting the comparison code into its own function might be preferred, we should try to use the most appropriate type here to represent the return value.

agent/src/agentclient.rs
somnusfish: Do you know if there is better way to write this?
```
agentclient_error::Error::Assertion { message: msg } => {
                            return agentclient_error::Assertion { message: msg }.fail();
                        }
```

agent/src/agentclient.rs
```diff
@@ -165,6 +165,8 @@ impl<T: APIServerClient> BrupopAgent<T> {
     async fn gather_system_metadata(
```
cbgbt: Nit: I think this function name may be showing that it was a strange choice. Maybe something like `current_shadow_status` would be clearer? I don't exactly love that name either. There may be something better.

agent/src/agentclient.rs
```diff
@@ -280,15 +284,17 @@ impl<T: APIServerClient> BrupopAgent<T> {
     async fn update_status_in_shadow(
         &self,
         bottlerocket_shadow: &BottlerocketShadow,
+        state: BottlerocketShadowState,
+        crash_count: u32,
+        state_transition_failure_timestamp: Option<DateTime<Utc>>,
```
cbgbt: Something that might help with the proliferation of these arguments would be to encapsulate the error handling state in a struct, like `ShadowErrorInfo` or something like that. This could be something we implement the [`Default`](https://doc.rust-lang.org/std/default/trait.Default.html) trait on so that we don't need to use the constants `0` and `None` when constructing the Status.

agent/src/agentclient.rs
```diff
@@ -318,12 +333,19 @@ impl<T: APIServerClient> BrupopAgent<T> {
...
-                    event!(
-                        Level::INFO,
-                        "Ready to finish monitoring and start update process"
-                    );
-                }
+                // Right now we consider the recover state for ErrorReset is Idle
+                BottlerocketShadowState::Idle => match bottlerocket_shadow_status.current_state {
+                    BottlerocketShadowState::ErrorReset => {
+                        self.handle_recover().await?;
+                        event!(Level::INFO, "Recover from ErrorReset");
```
cbgbt: Nit: tense

```suggestion
                        event!(Level::INFO, "Recovered from ErrorReset");
```

agent/src/agentclient.rs
```diff
@@ -354,10 +382,18 @@ impl<T: APIServerClient> BrupopAgent<T> {
...
+                    // Spec state should never be ErrorReset
+                    event!(Level::ERROR, "Spec state should never be ErrorReset");
+                    return agentclient_error::Assertion {
+                        message: "Spec state should never be ErrorReset.".to_string(),
+                    }
+                    .fail();
+                }
             }
         } else {
             event!(Level::INFO, "Did not detect action demand.");
```
cbgbt: I wonder if this log message should be `DEBUG` level? I'm realizing this might be quite noisy since it would likely be emitted on most runs through the event loop.

agent/src/agentclient.rs
```diff
@@ -398,41 +434,81 @@ impl<T: APIServerClient> BrupopAgent<T> {
...
             match self.handle_state_transition(&bottlerocket_shadow).await {
-                Ok(()) => {}
+                Ok(()) => {
+                    let crash_count;
+                    let state_transition_failure_timestamp;
+                    match bottlerocket_shadow_status.current_state {
+                        // Reset crash_count and state_transition_failure_timestamp for a successful update loop
+                        BottlerocketShadowState::MonitoringUpdate => {
+                            crash_count = 0;
+                            state_transition_failure_timestamp = None;
```
cbgbt: I think we'll want to wait to apply these until the time that we are transitioning from `MonitoringForUpdate` into `Idle`.

Reason being is that we want to be able to add support for customers to add custom checks that the node is healthy (#12, which references [this similar features in CLUO](https://github.com/coreos/container-linux-update-operator/blob/master/doc/before-after-reboot-checks.md)). Basically, we want any check that fails here to result the node being reverted to the previous version, and maintaining its crash loop state.

We can implement that functionality in a future release, but we should still make sure to only assign the fresh state after we've reached `Idle`.

models/src/node/mod.rs
```diff
@@ -148,6 +156,25 @@ impl BottlerocketShadow {
...
+    pub fn has_crashed(&self) -> bool {
+        self.status.as_ref().map_or(false, |node_status| {
+            node_status.current_state == BottlerocketShadowState::ErrorReset
+        })
+    }
+
+    /// Order BottleRocketShadow based on crash_count in status
+    /// to determine the priority to be handled by the controller.
+    /// Uninitialized status should be considered as lowest priority.
+    pub fn compare_status(&self, other: &Self) -> Ordering {
```
cbgbt: Naming suggestion:

```suggestion
    pub fn compare_crash_count(&self, other: &Self) -> Ordering {
```

I really like the use of `cmp` here, by the way :).

controller/src/statemachine.rs
```diff
@@ -1,6 +1,12 @@
-use models::node::{BottlerocketShadow, BottlerocketShadowSpec, BottlerocketShadowState};
+use models::node::{
+    BottlerocketShadow, BottlerocketShadowSpec, BottlerocketShadowState, BottlerocketShadowStatus,
+};
 
+use chrono::Utc;
 use tracing::instrument;
+use tracing::{event, Level};
+
+const RETRY_MAX_DELAY_IN_MINUTES: i64 = 24 * 60;
```
cbgbt, line 9: This should likewise be stored as a `Duration` to avoid having to encode the type in the name (`MINUTES`)

controller/src/statemachine.rs
```diff
@@ -47,3 +71,47 @@ pub fn determine_next_node_spec(brs: &BottlerocketShadow) -> BottlerocketShadowS
         }
     }
 }
+
+fn is_time_reached(node_status: &BottlerocketShadowStatus) -> bool {
```
cbgbt: This name is a bit too vague. I might suggest something like this with a docstring:

```suggestion
/// Returns whether or not an Idle node is allowed to enter an update workflow.
/// This returns false if the node has previously encountered an error and not yet
/// passed its retry timer.
fn node_allowed_to_update(node_status: &BottlerocketShadowStatus) -> bool {
```

agent/src/agentclient.rs
somnusfish: I think we need to make it clear that this method refetch the system metadata. Maybe `shadow_status_with_refetch_system_matadata` ?

agent/src/agentclient.rs
somnusfish: Sure, I can do that.

models/src/node/mod.rs
somnusfish: It's tricky to implement `cmp` for `BottlerocketShadow` though, that's why I did this method.

agent/src/agentclient.rs
somnusfish: `shadow_status_with_refreshed_system_matadata`

agent/src/agentclient.rs
somnusfish: Till here agent already handled the state transitioning from `MonitoringForUpdate` into `Idle`.

The logic for state transition is in `self.handle_state_transition`, this part is just the part to set the correct status to update. What you mentioned can be `self.handle_state_transition` and causing Err() with the return value which goes into the error handling part.

controller/src/statemachine.rs
somnusfish: I believe Duration in chrono is not constants function, I have the following error:

```
calls in constants are limited to constant functions, tuple structs and tuple variants
```

The Duration in tokio and std::time seems only operate on seconds and below, I am kind of feel that could cause more confusion. So I will keep this here for the next version. We can discuss more here.

agent/src/agentclient.rs
cbgbt: I like that suggestion!

controller/src/statemachine.rs
cbgbt: I think we'd want to use `std::time::Duration`. It does seem to mostly rely on second granularity for its interface, but I think it's usually a good idea to encode a measured unit in your type, rather than in your variable name.

I think with the lack of minute interface, I'm willing to disagree but ship it in spite of this.

cbgbt (approved):

cbgbt:
I think this is ready. There are a few nits, but they can be iteratively cleaned up or added here, but I'm overall happy with the behavior.

etungsten (approved):

Issue #130, opened 2021-12-09
0.2.0: Add integration testing subcommand which monitors new nodes for successful updates


cbgbt:
**Issue or Feature Request:**
The integration subcrate is currently written to manually assemble a cluster that can allow a human to quickly verify that nodes are being updated using `kubectl get brn`.

One way we could expand this into a pass/fail test mechanism would be to create a subcommand of the integration test subcrate (or a script) which provides userdata to the launched instances to lock to a particular target version for updates, then use an AMI of an older version of Brupop. After waiting some amount of time, we could confirm that the instances successfully made it to the target version and the Idle state.

PR #165, opened 2022-03-08
add a new integration test subcommand - monitor

gthao313:
**Issue number:**
#130

**Description of changes:**
A mechanism to help human automatically verify if that nodes are being updated to target Bottlerocket version

**Testing done:**
Unit test
```
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests (target/debug/deps/integ-6e89cca0ffdbf236)

running 3 tests
test monitor::test::test_check_brupop_resources_health ... ok
test monitor::test::test_confirm_update_success ... ok
test monitor::test::test_check_shadows_health ... ok
```

integration test
```
cargo run --bin integ monitor --cluster-name bottlerocket-test --region us-west-2

.........
[Not ready] keep monitoring!
brs: "brs-ip-192-168-133-128.us-west-2.compute.internal"      current_version: "1.7.0"       current_state: Idle
brs: "brs-ip-192-168-135-205.us-west-2.compute.internal"      current_version: "1.7.0"       current_state: Idle
brs: "brs-ip-192-168-145-37.us-west-2.compute.internal"      current_version: "1.7.0"       current_state: Idle
[Complete]: All nodes has been successfully updated to latest version!
```

**Terms of contribution:**


integ/src/monitor.rs
```diff
@@ -0,0 +1,533 @@
...
+        let brss: Api<BottlerocketShadow> = Api::namespaced(self.k8s_client.clone(), NAMESPACE);
+        let brss_object_list = brss
+            .list(&ListParams::default())
+            .await
+            .context(monitor_error::FindBrupopResources {})?;
+
+        Ok(brss_object_list)
+    }
+
+    async fn fetch_brupop_resources(&self) -> Result<ObjectList<Pod>> {
```
cbgbt: nit:

```suggestion
    async fn fetch_brupop_pods(&self) -> Result<ObjectList<Pod>> {
```

integ/src/monitor.rs
```diff
@@ -0,0 +1,533 @@
...
+        }
+    }
+
+    // verify if Burpop resources (agent, api-server, controller) are in `running` status.
+    async fn check_resources_health(&self, pods: &ObjectList<Pod>) -> Result<bool> {
+        for pod in pods {
+            if !is_pod_running(&pod) {
+                return Ok(false);
+            }
+        }
```
cbgbt: Not sure if you've played much with the Iterator API, but this might be more succinctly written as:

```rust
Ok(pods.iter().any(|pod| !is_pod_running(pod))
```

The current implementation is fine, it's just a tip. I think you might be able to return just a `bool` here instead of a `Result` though.

cbgbt (approved):
Nice work with the tests! Just a few nits.

🚢

models/src/node/mod.rs
```diff
@@ -196,9 +196,9 @@ impl BottlerocketShadowSpec {
 #[derive(Clone, Debug, Default, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
 pub struct BottlerocketShadowStatus {
     #[validate(regex = "SEMVER_RE")]
-    current_version: String,
+    pub current_version: String,
```
somnusfish: It might be better to keep these versions private and create a method to read these values since your monitor only tried to fetch the value. The reason we want it to be private is we don't want any method accidentally update these values which could cause issue for the behavior of agent and controller.

integ/src/monitor.rs
```diff
@@ -0,0 +1,533 @@
...
+
+        Ok(true)
+    }
+
+    // confirm that the instances successfully made it to the target version and the Idle state
+    async fn confirm_update_success(
+        &self,
+        bottlerocketshadows: &ObjectList<BottlerocketShadow>,
+    ) -> Result<bool> {
+        let mut update_success = 1;
```
somnusfish: This seems to try to use int to do bool's job. Maybe use bool would be more clear? like
```
let mut update_success = true;

if ...


{
   update_success = update_success & false;
}

Ok(update_success)
```

integ/src/monitor.rs
```diff
@@ -0,0 +1,533 @@
...
+        impl Clone for Monitor {
+            fn clone(&self) -> Self;
+        }
+    }
+}
+
+// compute the estimated update time to trigger monitor exit
+// formula: number_of_node*300 secs + 300 secs
+fn estimate_expire_time() -> i32 {
+    let expire_time = DEFAULT_INSTANCE_COUNT * 300 + 300;
```
somnusfish: Where is this 300 coming from? Would it be better to predefine in a const?

models/src/node/mod.rs
cbgbt: One nice thing about Rust is that values are immutable by default, unless you use the `mut` keyword. I think for that reason interfaces can be a bit looser with making fields public. Since we have the code around setting this using a timestamp though, I agree that using a method is probably best.

cbgbt (approved):

somnusfish (approved):

Issue #162, opened 2022-02-22
Implement creating the instances via nodegroups.


gthao313:
> I recommend creating the instances via nodegroups via eksctl configuration files. You can specify any arbitrary setting you want. So you can have the node-labels set as a Bottlerocket settings directly.

According to Erikson's suggestion on improving old version BR instance, we think it's better to create nodes via nodegroups.

Meanwhile, Currently Integration test needs users to input the old AMIs `--instance-id`. We can have a functionality about getting AMI version which gets fetched via SSM. Let user choose the BR version they want and Integration test to get the AMI.

PR #167, opened 2022-03-10
Add find_ami_id logic instead of pass argument `--instance-id`

gthao313:
**Issue number:**
#162 

**Description of changes:**
```
Date:   Thu Mar 10 01:43:23 2022 +0000

    Add find_ami_id logic instead of pass argument `--instance-id`

    currently integration test needs human manually find bottlerocket ami id
    and pass it to integration test via `--instance-id`, which isn't efficient
    and convenient; therefore, we add a new logic about finding ami id
    via aws-sdk-ssm according to arch, bottlerocket-version, and eks-version
    provided by the users.
```

**Testing done:**
```
cargo run --bin integ integration-test --cluster-name bottlerocket-test --region us-west-2 --bottlerocket-version 1.5.0  --arch x86_64
```

```
cargo run --bin integ integration-test --cluster-name bottlerocket-test --region us-west-2 --bottlerocket-version 1.5.0  --arch arm64
```

**Terms of contribution:**

### DIFF
--- a/integ/src/ec2_provider.rs
+++ b/integ/src/ec2_provider.rs
@@ -40,17 +40,24 @@ impl Ec2Creator {}
 
 pub async fn create_ec2_instance(
     cluster: ClusterInfo,
-    node_ami: &str,
+    ami_arch: &str,
+    bottlerocket_version: &str,
 ) -> ProviderResult<CreatedEc2Instances> {
     // Setup aws_sdk_config and clients.
     let region_provider = RegionProviderChain::first_try(Some(Region::new(cluster.region.clone())));
     let shared_config = aws_config::from_env().region(region_provider).load().await;
     let ec2_client = aws_sdk_ec2::Client::new(&shared_config);
+    let ssm_client = aws_sdk_ssm::Client::new(&shared_config);
 
     // Prepare security groups
     let mut security_groups = vec![];
     security_groups.append(&mut cluster.nodegroup_sg.clone());
     security_groups.append(&mut cluster.clustershared_sg.clone());
+
+    // Prepare ami id
+    //default eks_version to the version that matches cluster
+    let eks_version = cluster.version;
+    let node_ami = find_ami_id(&ssm_client, ami_arch, bottlerocket_version, &eks_version).await?;
 
     // Prepare instance type
     let instance_type = instance_type(&ec2_client, &node_ami).await?;
@@ -136,6 +143,30 @@ pub async fn terminate_ec2_instance(cluster: ClusterInfo) -> ProviderResult<()> 
     .await
     .context("Timed-out waiting for instances to reach the `terminated` state.")??;
     Ok(())
+}
+
+// Find the node ami id to use.
+async fn find_ami_id(
+    ssm_client: &aws_sdk_ssm::Client,
+    arch: &str,
+    br_version: &str,
+    eks_version: &str,
+) -> ProviderResult<String> {
+    let parameter_name = format!(
+        "/aws/service/bottlerocket/aws-k8s-{}/{}/{}/image_id",
+        eks_version, arch, br_version
+    );
+    let ami_id = ssm_client
+        .get_parameter()
+        .name(parameter_name)
+        .send()
+        .await
+        .context("Unable to get ami id")?
+        .parameter
+        .context("Unable to get ami id")?
+        .value
+        .context("ami id is missing")?;
+    Ok(ami_id)
 }
 
 /// Determine the instance type to use. If provided use that one. Otherwise, for `x86_64` use `m5.large`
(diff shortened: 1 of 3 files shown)


integ/src/main.rs
```diff
@@ -20,6 +20,11 @@ const DEFAULT_KUBECONFIG_FILE_NAME: &str = "kubeconfig.yaml";
 const DEFAULT_REGION: &str = "us-west-2";
 const CLUSTER_NAME: &str = "brupop-integration-test";
 
+//The default values for AMI ID
+const AMI_ARCH: &str = "x86_64";
+const BOTTLEROCKET_VERSION: &str = "1.6.1";
+const EKS_VERSION: &str = "1.21";
```
gthao313: Not sure if we should default those values here since we might need maintain them periodically.

cbgbt: I agree that maybe the `BOTTLEROCKET_VERSION` shouldn't have a default. But we may want to ensure that the architecture and EKS versions match the cluster that can be generated?

cbgbt (approved):

integ/src/main.rs
gthao313: Yeah. I'll different strategies for those variables.
`BOTTLEROCKET_VERSION`: No default, completely provided by user.
`AMI_ARCH ` : Default to `x86_64`, and `arm64` specified by user.
`EKS_VERSION `:  Default to the version that matches the cluster (might be able to be generated).

gthao313:
Above address Sean's comment. 
> Yeah. I'll different strategies for those variables.
BOTTLEROCKET_VERSION: No default, completely provided by user.
AMI_ARCH : Default to x86_64, and arm64 specified by user.
EKS_VERSION : Default to the version that matches the cluster (might be able to be generated).

cbgbt (approved):

integ/src/ec2_provider.rs
```diff
@@ -138,6 +145,30 @@ pub async fn terminate_ec2_instance(cluster: ClusterInfo) -> ProviderResult<()>
...
 
+// Find the node ami id to use.
+async fn find_ami_id(
+    ssm_client: &aws_sdk_ssm::Client,
+    arch: &str,
+    br_version: &str,
+    eks_version: &str,
+) -> ProviderResult<String> {
+    let parameter_name = format!(
+        "/aws/service/bottlerocket/aws-k8s-{}/{}/{}/image_id",
```
somnusfish, line 156: Maybe define this as a pattern variable?

integ/src/main.rs
```diff
@@ -60,9 +63,12 @@ enum SubCommand {
...
 // Stores user-supplied arguments for the 'integration-test' subcommand.
 #[derive(StructOpt, Debug)]
-struct IntegrationTestArgs {
-    #[structopt(long = "--instance-ami-id")]
-    instance_ami_id: String,
+pub struct IntegrationTestArgs {
+    #[structopt(long = "--bottlerocket-version")]
+    bottlerocket_version: String,
+
+    #[structopt(long = "--arch", default_value = AMI_ARCH)]
```
somnusfish, line 78: Could you add the validation of input arches? For example, predefine a list of valid arches.

integ/src/ec2_provider.rs
gthao313: Can you give an example about it?  thanks

somnusfish: Something similar to 
```
const PARAMETER_PATERN = "/aws/service/bottlerocket/aws-k8s-{}/{}/{}/image_id"
let parameter_name = format!(PARAMETER_PATERN, ...)
```

gthao313: I don't think we can use `format!(PARAMETER_PATERN, ...)` since format! needs syntax like format!({}, ...)`. We can change it to 
```
const PARAMETER_PATERN = "/aws/service/bottlerocket/aws-k8s-"
let parameter_name = format!({}{}/{}/{}/image_id, PARAMETER_PATERN, ...)
```
But I don't think this is readable, so I would keep current format. Thanks. 👍

integ/src/main.rs
```diff
@@ -20,9 +23,21 @@ const DEFAULT_KUBECONFIG_FILE_NAME: &str = "kubeconfig.yaml";
...
 const CLUSTER_NAME: &str = "brupop-integration-test";
 
+//The default values for AMI ID
+const AMI_ARCH: &str = "x86_64";
+
 /// This value configure how long it sleeps between create instance and label instance.
 const INTEGRATION_TEST_DELAY: Duration = Duration::from_secs(60);
 
+lazy_static! {
+    static ref ARCH: Vec<ArchitectureValues> = {
```
cbgbt: As a nit, Rust defines a macro to make Vec creation a bit easier:

```rust
static ref ARCHES: Vec<ArchitectureValues> = vec![ArchitectureValues::Arm64, ArchitectureValues::X8664];
```

cbgbt (approved):

somnusfish (approved):

PR #177, opened 2022-03-31
create instances via nodegroup

gthao313:
**Issue number:**
#162

**Description of changes:**
Currently we create bottlerocket nodes via launching instances directly, and in this PR we're going to create bottlerocket node via nodegroups. From this we can specify any arbitrary setting we want. and we can have the node-labels set as a Bottlerocket settings directly.

Tasks

- Nodegroup creation and termination
- Launch template creation and termination
- One-time iam role and profile creation and deletion
- Mapping iam role and profile to cluster

**Testing done:**
Test on ipv4 and ipv6 clusters
```
cargo run --bin integ integration-test --cluster-name bottlerocket-test --region us-west-2 --bottlerocket-version 1.5.0  --arch x86_64 --nodegroup-name=tianhg-test-x64
```

```
cargo run --bin integ clean --cluster-name bottlerocket-test --region us-west-2 --nodegroup-name=tianhg-test-x64
```

```
cargo run --bin integ integration-test --cluster-name bottlerocket-test --region us-west-2 --bottlerocket-version 1.5.0  --arch arm64 --nodegroup-name=tianhg-test-arm64
```
```
cargo run --bin integ clean --cluster-name bottlerocket-test --region us-west-2  --nodegroup-name=tianhg-test-arm64
```

**Terms of contribution:**


integ/src/nodegroup_provider.rs
```diff
@@ -0,0 +1,573 @@
...
+            "--username",
+            "system:node:{{EC2PrivateDNSName}}",
+        ])
+        .output()
+        .context("Unable to map iam identity.")?;
+
+    Ok(())
+}
+
+fn eks_role_policy_document() -> String {
```
cbgbt: Should this just be a `const &str`?

integ/src/nodegroup_provider.rs
```diff
@@ -0,0 +1,573 @@
...
+    let nodegroup = eks_client
+        .describe_nodegroup()
+        .nodegroup_name(nodegroup_name)
+        .cluster_name(cluster_name)
+        .send()
+        .await;
+
+    match nodegroup {
+        Err(_resource_not_found_exception) => Ok(true),
+        _ => Ok(false),
```
cbgbt, line 549: Are these bools reversed? I would expect `resource_not_found` to result in `false`.

integ/src/nodegroup_provider.rs
```diff
@@ -0,0 +1,573 @@
...
+        .tags(
+            Tag::builder()
+                .key(INSTANCE_TAG_NAME)
+                .value(INSTANCE_TAG_VALUE)
+                .build(),
+        )
+        .build()
+}
+
+fn userdata(endpoint: &str, cluster_name: &str, certificate: &str) -> String {
```
cbgbt: We should probably use a [TOML serializer](https://docs.rs/toml/latest/toml/ser/fn.to_string_pretty.html) with serde for this, since it would be pretty easy to break the TOML formatting with the right input.

integ/src/nodegroup_provider.rs
gthao313: I think the function name `confirm_nodegroup_exists` is not accurate, and I'll rename it to `confirm_nodegroup_deleted`. thanks

integ/src/nodegroup_provider.rs
gthao313: Do you mean something like below? I tried this but `create_launch_template().user_data( input)`, the input need a valid base64 string. Can I still use TOML serializer to output base64 string? 
```
n userdata(endpoint: &str, cluster_name: &str, certificate: &str) -> ProviderResult<String> {
    let user_data = toml::ser::to_string_pretty(&format!(
        r#"[settings.updates]
            ignore-waves = true
                
            [settings.kubernetes]
            api-server = "{}"
            cluster-name = "{}"
            cluster-certificate = "{}"
        "#,
        endpoint, cluster_name, certificate
    ))
    .context("Unable convert userdata to pretty string")?;

    Ok(user_data)
```

cbgbt: I meant more like create a series of structs like:

```rust
#[derive(Serialize)]
struct Settings {
    updates: UpdateSettings,
    kubernetes: KubernetesSettings,
}

#[derive(Serialize)]
struct UpdateSettings {
    ignore_waves: bool,
}

#[derive(Serialize)]
struct KubernetesSettings {
    api_server: String,
    // etc...
}
```

Then you could do something like:

```rust
let settings = Settings {
    updates: UpdateSettings {
        ignore_waves: true,
    },
    kubernetes: KubernetesSettings {
        // etc...
    },
}
toml::ser::to_string_pretty(settings);
```

somnusfish (approved):

integ/src/nodegroup_provider.rs
gthao313: I tried the method that @cbgbt mentioned above; however, the string formate of userdata which was created by `toml::ser::to_string_pretty()` can be recognized by 'early-boot-config'. We can't use this method and I think it's better to keep the original method.
Error
```
[    5.826817] early-boot-config[1483]: Provider error: Unable to serialize settings from instance user data: TOML data did not contain 'settings' section
```
userdata created by toml
```
"[\"settings.updates\"]\n
ignore-waves = true\n\n
[\"settings.kubernetes\"]\n
api-server = 'xxxx7.us-west-2.eks.amazonaws.com'\n
cluster-name = 'bottlerocket-test'\n
cluster-certificate = '=xxxxx'\n"
```

integ/src/pods-template.yaml
```diff
@@ -31,7 +31,7 @@ spec:
     spec:
       containers:
       - name: nginx
-        image: k8s.gcr.io/nginx-slim:0.8
+        image: nginx:1.21.6
```
gthao313, line 34: Update nginx image version to `1.21.6` so that able to create arm/amd pods. Old version Nginx doesn't support arm.

gthao313:
Push above address comments, update third party packages.

webern (approved):

cbgbt (approved):

integ/src/nodegroup_provider.rs
```diff
@@ -0,0 +1,623 @@
...
+        .tags(
+            Tag::builder()
+                .key(INSTANCE_TAG_NAME)
+                .value(INSTANCE_TAG_VALUE)
+                .build(),
+        )
+        .build()
+}
+
+fn userdata(
```
gthao313, line 456: resolve conflict on this part. The code is from [IPV6 PR](https://github.com/bottlerocket-os/bottlerocket-update-operator/commit/e803d0eaa2f4be468415ea047627df33201b832a)

gthao313:
Push above rebase and update aws rust dependencies to `0.13.0`

integ/src/main.rs
```diff
@@ -27,11 +28,11 @@ const DEFAULT_KUBECONFIG_FILE_NAME: &str = "kubeconfig.yaml";
 const DEFAULT_REGION: &str = "us-west-2";
 const CLUSTER_NAME: &str = "brupop-integration-test";
 
-//The default values for AMI ID
+// The default values for AMI ID
 const AMI_ARCH: &str = "x86_64";
 
-/// This value configure how long it sleeps between create instance and label instance.
-const INTEGRATION_TEST_DELAY: Duration = Duration::from_secs(60);
+// The default name for nodegroup
```
arnaldo2792: pedantic nit:

```suggestion
// The default name for the nodegroup
```

integ/src/main.rs
```diff
@@ -232,38 +223,55 @@ async fn run() -> Result<()> {
...
+            info!("Terminating nodegroup ...");
+            terminate_nodegroup(cluster_info, &args.nodegroup_name)
+                .await
+                .context(error::TerminateNodeGroup)?;
+
+            // If EKS cluster still has running nodes which need brupop, Integration-test shouldn't uninstall brupop, delete test pods, and kubeconfig file.
+            if !nodes_exist(k8s_client).await.context(error::RunBrupop)? {
+                // Clean up all brupop resources like namespace, deployment on brupop test
+                info!(
+                    "Deleting all brupop cluster resources which created by integration test ..."
```
arnaldo2792: nit:

```suggestion
                    "Deleting all brupop cluster resources created by integration test ..."
```

integ/src/updater.rs
```diff
@@ -195,6 +148,18 @@ pub async fn process_pods_test(action: Action, kube_config_path: &str) -> Update
     Ok(())
 }
 
+// Find if any node is running in cluster
+pub async fn nodes_exist(k8s_client: kube::client::Client) -> UpdaterResult<bool> {
+    let nodes: Api<Node> = Api::all(k8s_client.clone());
+    // let a = nodes.list_node().await.context(update_error::FindBrupopPods {})?;
```
arnaldo2792: Extra comment here?

integ/src/pods-template.yaml
arnaldo2792, line 34: Should you have a different commit for this change?

integ/src/main.rs
```diff
@@ -232,38 +223,55 @@ async fn run() -> Result<()> {
...
+                &KubeConfigOptions::default(),
+            )
+            .await
+            .context(error::LoadKubeConfig)?;
+            let k8s_client =
+                kube::client::Client::try_from(config).context(error::CreateK8sClient)?;
 
-            // delete all created pods for testing.
-            info!(
+            // Terminate nodegroup which created by integration test.
```
arnaldo2792: pendantic nit:

```suggestion
            // Terminate nodegroup created by integration test.
```

integ/src/updater.rs
```diff
@@ -195,6 +148,18 @@ pub async fn process_pods_test(action: Action, kube_config_path: &str) -> Update
     Ok(())
 }
 
+// Find if any node is running in cluster
```
arnaldo2792: nit:

```suggestion
// Find if any node is running in the cluster
```

somnusfish (approved):

Issue #161, opened 2022-02-21
Implement Custom Resource Migrations


cbgbt:
We should implement [webhook conversion](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion) to allow migrating between custom resource definitions. As it stands, there's not a great way for us to make changes to the custom resource definition without making all new fields optional-only.

Providing a mechanism for migrating BRS data would much more complicated upgrade scenarios.

PR #183, opened 2022-04-05
Add the structure for versions in CustomResourceDefinition.

somnusfish:
This is the first part for CustomResourceDefinition migration.
The change includes re-architect the BottlerocketShadow in
models/src/node, also adjusts the way to generate CRD's yaml file.
Right now webhook is not suported yet, the default version
conversion is None.

**Issue number:**
#161 

**Description of changes:**
This is the first part for CustomResourceDefinition migration.
The change includes re-architect the BottlerocketShadow in
models/src/node, also adjusts the way to generate CRD's yaml file.
Right now webhook is not suported yet, the default version
conversion is None.

**Testing done:**
Build and tested the updated image, Brupop was able to upgrade cluster properly.

**Terms of contribution:**


models/src/node/crd/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+    crd
+}
+
+pub fn combined_crds() -> CustomResourceDefinition {
+    let mut crds: Vec<CustomResourceDefinition> = BOTTLEROCKETSHADOW_CRD_METHODS
+        .iter()
+        .map(|crd_method| crd_method())
+        .collect();
+    let latest_crd = crds.pop().unwrap();
```
gthao313, line 129: Can we avoid using `unwarp()` here?  Not sure if you have any concerns here but I think we might able to use `context` to handle error here?

models/src/node/crd/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "({}, {})", self.node_name, self.node_uid)
+    }
+}
+
+pub fn brs_name_from_node_name(node_name: &str) -> String {
+    format!("brs-{}", node_name)
+}
+
+pub fn combine_crd(
```
gthao313: I think this function might not need to be pub.

models/src/node/crd/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+    latest_crd: CustomResourceDefinition,
+    old_crds: Vec<CustomResourceDefinition>,
+) -> CustomResourceDefinition {
+    // Combine all different versions of customer resources into one CustomResourceDefinition yaml
+    // Related open issue: https://github.com/kube-rs/kube-rs/issues/569
+    let mut crd = latest_crd;
```
gthao313: Why not directly using `latest_crd` variable instead of assign it to `crd`?

somnusfish: Not sure if I understand the syntax correctly. But I want to make latest_crd mutable. I am not sure if directly use latest_crd could do that. I will try.

models/src/node/crd/mod.rs
somnusfish: I would say this combined_crds() will be called by build.rs on compiling time. And I would expect this to panic if this unwrap() failed, which means there is no CRD defined.

gthao313: That makes sense. Thanks

models/src/node/crd/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+        crd_methods.push(v2::BottlerocketShadow::crd as fn()->CustomResourceDefinition);
+        crd_methods
+    };
+}
+
+pub trait Selector {
+    fn selector(&self) -> error::Result<BottlerocketShadowSelector>;
+}
+
+impl<T: kube::ResourceExt> Selector for T {
```
cbgbt: I think blanket implementing this for all `ResourceExt` is too broad, since owner references could be any type of kubernetes object, and this function will probably not do what people expect if they aren't using it on *our* resources specifically.

Should we require our BottlerocketShadow CRDs to all extend an additional trait, which requires `ResourceExt` be implemented?

e.g.

```rust
trait BottlerocketShadowResource: kube::ResourceExt {}
```

Even if the trait doesn't have any methods, it could be used as a [marker trait](doc.rust-lang.org/std/marker/index.html) to denote what types are implemented by this `Selector` trait.

models/src/node/crd/mod.rs
cbgbt: I believe you can resolve this by writing the function signature like this:

```rust
fn combine_crd(
    mut latest_crd: CustomResourceDefinition,
    old_crds: Vec<CustomResourceDefinition>,
) -> CustomResourceDefinition;
```

models/src/node/crd/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+pub fn brs_name_from_node_name(node_name: &str) -> String {
+    format!("brs-{}", node_name)
+}
+
+pub fn combine_crd(
+    latest_crd: CustomResourceDefinition,
+    old_crds: Vec<CustomResourceDefinition>,
+) -> CustomResourceDefinition {
+    // Combine all different versions of customer resources into one CustomResourceDefinition yaml
+    // Related open issue: https://github.com/kube-rs/kube-rs/issues/569
```
cbgbt: We should elevate some of this comment to be a function docstring. I think it might be a good idea to also write a module-level docstring explaining the layout of the `node::crd` package, and why it's done this way to enable migrations.

models/src/node/crd/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+    let mut crd = latest_crd;
+    for old_crd in old_crds {
+        let mut old_versions = old_crd.spec.versions;
+
+        // Adjust storage value via #derive(CustomResource) is supported yet.
+        // Change old versions' storage to false since only one stroage true is allowed
```
cbgbt: nit:

```suggestion
        // Change old versions' storage to false since only one storage true is allowed
```

models/src/node/crd/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+lazy_static! {
+    static ref BOTTLEROCKETSHADOW_CRD_METHODS: Vec<fn() -> CustomResourceDefinition> = {
+        // A list of CRD methods for different BottlerocketShadow version
+        // The latest version should be added at the end of the vector.
+        let mut crd_methods = Vec::new();
+        crd_methods.push(v1::BottlerocketShadow::crd as fn()->CustomResourceDefinition);
+        crd_methods.push(v2::BottlerocketShadow::crd as fn()->CustomResourceDefinition);
+        crd_methods
+    };
+}
```
cbgbt, line 54: I think you could add the result of `combine_crds()` here as a static value, since it should never change.

cbgbt (commented):
This is cool. Nice work!

models/src/node/crd/mod.rs
somnusfish: Somehow `serde::ser::Serialize` get lost when I put it in lazy_static, do I would just keep this method as it is.

somnusfish:
Addressed feedbacks.

models/src/node/crd/mod.rs
```diff
@@ -0,0 +1,131 @@
+/// A package aim to solve the Custom Resource Migrations issue.
```
cbgbt: nit: I would frame it as enabling migrations to newer versions rather than as an issue:

```suggestion
/// A package aim to enable migrations to new Custom Resource Definitions.
```

cbgbt (approved):
Nice! :ship: :rocket:

gthao313 (approved):

PR #184, opened 2022-04-06
Do not authorize requests with non-singular owner

cbgbt:
**Issue number:** n/a

**Description of changes:**

Tokens should be unique to a single Pod. If the Kubernetes API suggests
that a token is owned by multiple Pods, we should reject the request.

This also fixes up some inconsistencies in error messages.

**Testing done:** Unit tests pass. I also tested a cluster update from 1.4.2 to 1.7.0, which was successful.

**Terms of contribution:**

### DIFF
--- a/apiserver/src/auth/authorizor.rs
+++ b/apiserver/src/auth/authorizor.rs
@@ -139,13 +139,18 @@ impl<T: TokenReviewer> K8STokenAuthorizor<T> {
         token_review_status: &TokenReviewStatus,
         node_selector: &BottlerocketShadowSelector,
     ) -> Result<(), AuthorizationError> {
-        let pod_name = token_review_status
+        let pod_names = token_review_status
             .user
             .as_ref()
             .and_then(|user| user.extra.as_ref())
             .and_then(|extra| extra.get(POD_NAME_INFO_KEY))
-            .and_then(|pod_names| pod_names.first())
             .context(TokenReviewMissingPodName)?;
+
+        // Only one Pod should own a given token. If multiple Pods are present in the response,
+        // we should not proceed.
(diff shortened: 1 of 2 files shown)


gthao313 (approved):
LGTM

apiserver/src/auth/authorizor.rs
```diff
@@ -139,14 +139,16 @@ impl<T: TokenReviewer> K8STokenAuthorizor<T> {
...
+        let pod_names = token_review_status
             .user
             .as_ref()
             .and_then(|user| user.extra.as_ref())
             .and_then(|extra| extra.get(POD_NAME_INFO_KEY))
-            .and_then(|pod_names| pod_names.first())
             .context(TokenReviewMissingPodName)?;
 
+        snafu::ensure!(pod_names.len() <= 1, TokenReviewMultiplePodNames);
+        let pod_name = pod_names.first().context(TokenReviewMissingPodName)?;
```
arnaldo2792: If I'm not misunderstanding, you can just check that `pod_names.len()` is 1, since you error out when there are more than one, or if there aren't any.

```suggestion
        snafu::ensure!(pod_names.len() == 1, TokenReviewUnexpectedPodNames);
```

cbgbt: My problem was that I wanted `TokenReviewMissingPodName` on <1 and `TokenReviewMultiplePodNames` on >1. I suppose I could also solve this with a match, but I'd still want to call `pod_names.first()` anyways.

somnusfish: It might be better if you could add some comments here to describe what you are trying to match. It is very confusing on the first read.

cbgbt:
Pushed comments to clarify the error check.

somnusfish (approved):

gthao313 (approved):

Issue #178, opened 2022-03-31
Bruop doesn't start on EKS cluster with IPv6 enabled.


pmacieje:
**Image I'm using:** bottlerocket-update-operator:v0.2.0

**Issue or Feature Request:** bruop-controller, apiservers and agent's doesn't start at all.

Errors:

**Controller:** "Error: ClientCreate { source: InferConfig(InferConfigError { in_cluster: ParseClusterUrl(InvalidUri(InvalidAuthority)), kubeconfig: ReadConfig(Os { code: 2, kind: NotFound, message: "No such file or directory" }, "/.kube/config") }) }"

**ApiServer:** {"v":0,"name":"apiserver","msg":"brupop apiserver failed.","level":50,"hostname":"brupop-apiserver-745c5cffd9-n7tdb","pid":1,"time":"2022-03-31T11:40:59.531118923+00:00","target":"apiserver","line":23,"file":"apiserver/src/main.rs","error":"Unable to create client: 'Failed to infer configuration: failed to infer config: in-cluster: (failed to parse cluster url: invalid authority), kubeconfig: (failed to read kubeconfig from '\"/.kube/config\"': No such file or directory (os error 2))'"}

**Additional info:** I've deployed bruop on EKS cluster v.1.21 with IPv6 enabled. Also deployed on IPv4 EKS cluster and there looks like work just fine.

cbgbt:
Thanks for raising this issue. We'll take a look at Brupop in IPv6 clusters and see if we can reproduce.

somnusfish:
I was able to reproduce this issue in the IPv6 cluster.

This issue is caused by `kube::client` inferring the default config from the cluster didn't format the IPv6 URI correctly. I have submitted a [bug report](https://github.com/kube-rs/kube-rs/pull/875) with a fix. Right now we are blocked on the upstream release.

To prevent this happen in future, we will add integration tests for IPv6 cluster (#186 ).

somnusfish:
This issue could be resolved with the release v0.2.1 which added support the IPv6 cluster. Please feel free to reopen this issue if you have future questions/concerns.

PR #188, opened 2022-04-13
Add support for IPv6 cluster

somnusfish:
**Issue number:**
#178 

**Description of changes:**
* "rustls-tls" feature must be used in kube 0.71.0 so kube::client could
build IPv6 client properly.
* API server would reply on environment variable KUBERNETES_SERVICE_HOST
to detect the ip family for the cluster
* Corresponding changes related to the return value change by `Store`

** Note **
Not sure about Brupop behavior in Dual Ip family cluster. 

**Testing done:**
Tested on IPv6 and IPv4 cluster.

**Terms of contribution:**

### DIFF
--- a/apiserver/src/api/mod.rs
+++ b/apiserver/src/api/mod.rs
@@ -34,6 +34,7 @@ use kube::{
 };
 use opentelemetry::global::meter;
 use snafu::{OptionExt, ResultExt};
+use std::env;
 use tracing::{event, Level};
 use tracing_actix_web::TracingLogger;
 
@@ -130,6 +131,21 @@ pub async fn run_server<T: 'static + BottlerocketShadowClient>(
     );
 
     // Set up the actix server.
+
+    // Use IP for KUBERNETES_SERVICE_HOST to decide the IP family for the cluster,
+    // Match API server IP family same as cluster
+    let k8s_service_addr =
+        env::var("KUBERNETES_SERVICE_HOST").context(error::MissingClusterIPFamiliy)?;
+    let server_addr = if k8s_service_addr.contains(":") {
+        // IPv6 format
+        format!("[::]:{}", server_port)
+    } else {
+        // IPv4 format
+        format!("0.0.0.0:{}", server_port)
+    };
+
+    event!(Level::DEBUG, ?server_addr, "Server addr localhost.");
+
     let server = HttpServer::new(move || {
         App::new()
             .wrap(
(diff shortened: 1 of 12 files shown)


apiserver/src/api/mod.rs
```diff
@@ -130,6 +131,18 @@ pub async fn run_server<T: 'static + BottlerocketShadowClient>(
     );
 
     // Set up the actix server.
+    let server_addr;
+
+    // Use IP for KUBERNETES_SERVICE_HOST to decide the IP family for the cluster
+    let k8s_service_addr = env::var("KUBERNETES_SERVICE_HOST").context(error::MissEnvVar)?;
+    if k8s_service_addr.contains(":") {
```
gthao313: Nit: I'd like to see more description about what the formate would be for `k8s_service_addr` in IPV6 and IPV4. Can  you put some comments here to tell why you use `:` to differentiate IPV4 and IPV6.

gthao313 (approved):

apiserver/src/api/mod.rs
somnusfish: A normal IPv6 address using `:`, and a normal IPv4 address using `.`

somnusfish: A normal IPv6 address uses `:` to divide, and a normal IPv4 address uses `.` to divide.

apiserver/src/api/mod.rs
```diff
@@ -130,6 +131,18 @@ pub async fn run_server<T: 'static + BottlerocketShadowClient>(
     );
 
     // Set up the actix server.
+    let server_addr;
```
arnaldo2792: Nit:

```suggestion
    let server_addr;
    
    let server_addr = if k8s_service_addr {
    } else {
    }
```

I wonder if you should  validate the IP format?

apiserver/src/error.rs
```diff
@@ -16,6 +16,9 @@ pub enum Error {
     #[snafu(display("Unable to parse HTTP header. Missing '{}'", missing_header))]
     HTTPHeaderParse { missing_header: &'static str },
 
+    #[snafu(display("Unable to detect cluster IP family. For '{}'", source))]
```
arnaldo2792, line 19: The message isn't "generic" enough for the name of this error, I think it should be something along the lines of `Missing env var <>`. Or else, have an error that is specific for this message `MissingClusterIPFamiliy`.

apiserver/src/api/mod.rs
somnusfish: Which IP format? If it is for k8s service, we don't need validation since we just want to get the ip family information for the cluster.

arnaldo2792 (approved):

gthao313 (approved):

Issue #186, opened 2022-04-12
Add integration test for IPv6 cluster


somnusfish:
**Image I'm using:**

**Issue or Feature Request:**

Integration test for IPv6 cluster.

Some ideas:
* Maybe a shared account for integration test.

PR #193, opened 2022-04-15
make integ test support ipv6 cluster

gthao313:
testing with ipv6 cluster, we need specify dns-ip setting on node
userdata. So add a logic to find dns-ip from cluster info and transform
to ideal format.

**Issue number:**
#186 

**Description of changes:**
```
Date:   Thu Apr 14 23:47:49 2022 +0000

    make integ test support ipv6 cluster

    testing with ipv6 cluster, we need specify dns-ip setting on node
    userdata. So add a logic to find dns-ip from cluster info and transform
    to ideal format.
```

Meanwhile remove a strict and not generic filter to find instance profile.

**Testing done:**
Test with `brupop-ipv6` and `brupop-ipv4` cluster on brupop integration test account. 

ipv6
```
[tianhg@ip-172-31-39-243 bottlerocket-update-operator]$ kubectl -n brupop-bottlerocket-aws get brs
NAME                                                STATE              VERSION   TARGET STATE   TARGET VERSION
brs-ip-192-168-137-245.us-west-2.compute.internal   Idle               1.6.1     Idle           <no value>
brs-ip-192-168-140-207.us-west-2.compute.internal   Idle               1.7.0     Idle
brs-ip-192-168-143-103.us-west-2.compute.internal   MonitoringUpdate   1.7.0     Idle           1.7.0
```
ipv4
```
brs-ip-192-168-138-196.us-west-2.compute.internal   StagedUpdate   1.6.1     PerformedUpdate   1.7.0
brs-ip-192-168-155-104.us-west-2.compute.internal   Idle           1.6.1     Idle              <no value>
brs-ip-192-168-156-212.us-west-2.compute.internal   Idle           1.7.0     Idle
```

**Terms of contribution:**

### DIFF
--- a/integ/Cargo.toml
+++ b/integ/Cargo.toml
@@ -6,11 +6,11 @@ edition = "2018"
 publish = false
 
 [dependencies]
-aws-config = "0.0.26-alpha"
-aws-sdk-ec2 = "0.0.26-alpha"
-aws-sdk-eks = "0.0.26-alpha"
-aws-sdk-iam = "0.0.26-alpha"
-aws-sdk-ssm = "0.0.26-alpha"
+aws-config = "0.10.1"
+aws-sdk-ec2 = "0.10.1"
+aws-sdk-eks = "0.10.1"
+aws-sdk-iam = "0.10.1"
+aws-sdk-ssm = "0.10.1"
 base64 = "0.13.0"
 console_log = { version = "0.2", features = ["color"] }
 env_logger = "0.9"
--- a/integ/src/eks_provider.rs
+++ b/integ/src/eks_provider.rs
@@ -6,13 +6,21 @@
 use aws_config::meta::region::RegionProviderChain;
 use aws_sdk_ec2::model::{Filter, SecurityGroup, Subnet};
 use aws_sdk_ec2::Region;
+use aws_sdk_eks::model::IpFamily;
 
 use crate::error::{IntoProviderError, ProviderError, ProviderResult};
 
 use serde::{Deserialize, Serialize};
 use std::process::Command;
 
-#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, Default)]
+const IPV4_OCTET: &str = "10";
+const IPV6_HEXTET: &str = "a";
+const IPV4_DIVIDER: &str = ".";
+const IPV6_DIVIDER: &str = ":";
+
+pub type ClusterDnsIpInfo = (IpFamily, Option<String>);
+
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct ClusterInfo {
     pub name: String,
     pub region: String,
@@ -25,6 +33,7 @@ pub struct ClusterInfo {
     pub controlplane_sg: Vec<String>,
     pub clustershared_sg: Vec<String>,
     pub iam_instance_profile_arn: String,
+    pub dns_ip_info: ClusterDnsIpInfo,
 }
 
 pub fn write_kubeconfig(
@@ -109,8 +118,9 @@ pub async fn get_cluster_info(cluster_name: &str, region: &str) -> ProviderResul
             .collect();
 
     let node_instance_role = cluster_iam_identity_mapping(cluster_name, region)?;
-    let iam_instance_profile_arn =
-        instance_profile(&iam_client, cluster_name, &node_instance_role).await?;
+    let iam_instance_profile_arn = instance_profile(&iam_client, &node_instance_role).await?;
+
+    let dns_ip_info = dns_ip(&eks_client, cluster_name).await?;
 
     Ok(ClusterInfo {
         name: cluster_name.to_string(),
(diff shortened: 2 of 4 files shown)


integ/Cargo.toml
```diff
@@ -6,11 +6,11 @@ edition = "2018"
 publish = false
 
 [dependencies]
-aws-config = "0.0.26-alpha"
-aws-sdk-ec2 = "0.0.26-alpha"
-aws-sdk-eks = "0.0.26-alpha"
-aws-sdk-iam = "0.0.26-alpha"
-aws-sdk-ssm = "0.0.26-alpha"
+aws-config = "0.8"
```
somnusfish: The latest version is 0.10.1, is there anything blocking us from updating to the latest version? Otherwise we will be bugged by the bot again.

integ/src/eks_provider.rs
```diff
@@ -124,9 +131,79 @@ pub async fn get_cluster_info(cluster_name: &str, region: &str) -> ProviderResul
...
+                .as_ref()
+                .context("Cluster missing kubernetes_network_config field")?
+                .service_ipv4_cidr
+                .as_ref()
+                .context("Ipv6 idr missing data")
+                .map(|ids| ids.clone())?;
+
+            // EKS clusters derive the cluster dns IP by setting the last octet of the IPv4 CIDR to `10`,
+            // so we can use this logic to get cluster dns IP
+            let mut ipv4_vec: Vec<String> = ipv4_cidr.split(".").map(|s| s.to_string()).collect();
```
somnusfish: Could you abstract this logic in IPv4 and IPv6 to a separate method, and add an unit test for this method?

integ/src/eks_provider.rs
```diff
@@ -310,24 +387,11 @@ async fn instance_profile(
         .send()
         .await
         .context("Unable to list instance profiles")?;
-    let eksctl_prefix = format!("eksctl-{}", cluster_name);
     list_result
         .instance_profiles
         .as_ref()
         .context("No instance profiles found")?
         .iter()
-        .filter(|x| {
```
somnusfish, line 319: Is this from previous PR for the sg group?

integ/src/eks_provider.rs
```diff
@@ -124,9 +131,79 @@ pub async fn get_cluster_info(cluster_name: &str, region: &str) -> ProviderResul
...
+        .name(cluster_name)
+        .send()
+        .await
+        .context("Unable to get eks describe cluster")?;
+
+    let ip_family = describe_results
+        .cluster
+        .as_ref()
+        .context("Response missing cluster field")?
+        .kubernetes_network_config
```
somnusfish: This `kubernetes_network_config` was fetched 3 times in `dns_ip()`,  does it make sense to store the reference as a variable?

integ/src/eks_provider.rs
somnusfish: Would it be better if you could name `.` and `:`  to something like IPv4_divider as a const or a enum_str?

integ/src/eks_provider.rs
gthao313: It's about finding instance profile not related to SG. The logic here is pretty strict and not suitable for different situations, so I think we don't need this anymore. Actually I have this code change in my NodeGroup PR, but I think it might be worth to add here.

integ/Cargo.toml
gthao313: When I implemented this pr, the version `0.10.1` hadn't been released. I'm pretty confident with `0.8` and I haven't test with `0.10.1` version. I'll try with the latest version. : )

integ/src/eks_provider.rs
gthao313: Yeah, I think so. I was thinking keep it there for better readable. I'll update that.

gthao313:
Push above address Lu's comments.

integ/src/eks_provider.rs
```diff
@@ -124,9 +132,71 @@ pub async fn get_cluster_info(cluster_name: &str, region: &str) -> ProviderResul
...
+
+    match ip_family {
+        IpFamily::Ipv4 => {
+            let ipv4_cidr = kubernetes_network_config
+                .service_ipv4_cidr
+                .as_ref()
+                .context("Ipv6 idr missing data")
+                .map(|ids| ids.clone())?;
+
+            let ipv4 = transform_dns_ip(&ipv4_cidr, IPV4_DIVIDER, IPV4_OCTET);
```
etungsten: Cluster DNS IP conversion for IPv4 clusters is slightly more complicated. See https://github.com/awslabs/amazon-eks-ami/blob/02823975992a9291609498b00c00b6434e0e4d19/files/bootstrap.sh#L387-L408

~~If the service ipv4 cidr, starts with a `10` for the first octet, then we return `10.100.0.10`. Otherwise, we return `172.20.0.10`. We have the same logic in pluto: 
https://github.com/bottlerocket-os/bottlerocket/blob/62c2e64df8885c88534624288630c69342a589be/sources/api/pluto/src/main.rs#L214-L216~~

etungsten (changes requested):

integ/src/eks_provider.rs
```diff
@@ -124,9 +132,71 @@ pub async fn get_cluster_info(cluster_name: &str, region: &str) -> ProviderResul
...
+        .await
+        .context("Unable to get eks describe cluster")?;
+
+    let kubernetes_network_config = describe_results
+        .cluster
+        .as_ref()
+        .context("Response missing cluster field")?
+        .kubernetes_network_config
+        .as_ref()
+        .context("Cluster missing kubernetes_network_config field")?;
```
etungsten: non-blocking nit:
Whenever you run into situations where you want to retrieve deeply nested object, you can combine all the Option checks with [`and_then()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then).

integ/src/eks_provider.rs
etungsten: Sorry, correction. If service IPv4 cider does not exist, then we should let the bottlerocket host determine the cluster-dns-ip itself by not setting `cluster-dns-ip` in the instance userdata.

etungsten (approved):

integ/src/eks_provider.rs
```diff
@@ -124,9 +134,75 @@ pub async fn get_cluster_info(cluster_name: &str, region: &str) -> ProviderResul
...
         clustershared_sg,
         iam_instance_profile_arn,
+        dns_ip_info,
     })
 }
 
+async fn dns_ip(
+    eks_client: &aws_sdk_eks::Client,
+    cluster_name: &str,
+) -> ProviderResult<(IpFamily, Option<String>)> {
```
somnusfish: Maybe use `CLUSTER_DNS_IP_INFO` here to enforce type check?

somnusfish (approved):

Issue #207, opened 2022-06-07
Integration test - Monitor unable to handle when brupop resources are not created


gthao313:
### What is the issue:
I was trying to run integration test Monitor subcommand to monitor brupop update status; however, it was failed to process when brupop resources are not created. It still recognize that situation as completion status.

### How to fix:

Add logic to determine if brupop resources are created.

PR #209, opened 2022-06-13
integration test - monitor bug fix

gthao313:
Monitor deals with the situation improperly, which brupop resources
(agent,controller...) are not created. Integration test monitor will
recoginze it as all updates have been completed which is incorrect. Add
new logic to determine if brupop resources are created and make Monitor
bail out if brupop resources aren't created

**Issue number:**
#207 

**Description of changes:**
```
Date:   Mon Jun 13 22:45:01 2022 +0000

    integration test - monitor bug fix

    Monitor deals with the situation improperly, which brupop resources
    (agent,controller...) are not created. Integration test monitor will
    recoginze it as all updates have been completed which is incorrect. Add
    new logic to determine if brupop resources are created and make Monitor
    bail out if brupop resources aren't created
```

**Testing done:**
```
brs: "brs-ip-192-168-134-247.us-west-2.compute.internal"      current_version: "1.8.0"       current_state: Idle
brs: "brs-ip-192-168-146-250.us-west-2.compute.internal"      current_version: "1.8.0"       current_state: Idle
brs: "brs-ip-192-168-147-125.us-west-2.compute.internal"      current_version: "1.8.0"       current_state: Idle
[Complete]: All nodes has been successfully updated to latest version!
```

**Terms of contribution:**

### DIFF
--- a/integ/src/monitor.rs
+++ b/integ/src/monitor.rs
@@ -88,26 +88,24 @@ impl<T: BrupopClient> BrupopMonitor<T> {
         }
     }
 
-    // verify if Burpop pods (agent, api-server, controller) are in `running` status.
+    // verify if Brupop pods (agent, api-server, controller) are in `running` status.
     fn check_pods_health(&self, pods: &ObjectList<Pod>) -> bool {
-        pods.iter().all(|pod| is_pod_running(pod))
+        if pods.items.is_empty() {
+            return false;
+        } else {
+            return pods.iter().all(|pod| is_pod_running(pod));
+        }
     }
 
     // verify if brs has been created properly and initialized `status`.
-    async fn check_shadows_health(
-        &self,
-        bottlerocketshadows: &ObjectList<BottlerocketShadow>,
-    ) -> Result<bool> {
-        for bottlerocketshadow in bottlerocketshadows {
-            // verify if brs has been updated `status` properly
-            if bottlerocketshadow.status.is_some() {
-                continue;
-            } else {
-                return Ok(false);
-            }
+    fn check_shadows_health(&self, bottlerocketshadows: &ObjectList<BottlerocketShadow>) -> bool {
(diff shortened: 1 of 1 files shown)


integ/src/monitor.rs
```diff
@@ -90,23 +90,30 @@ impl<T: BrupopClient> BrupopMonitor<T> {
 
     // verify if Burpop pods (agent, api-server, controller) are in `running` status.
     fn check_pods_health(&self, pods: &ObjectList<Pod>) -> bool {
```
cbgbt, line 92: This function is pretty well self-contained. It would be a good idea to write unit tests for this function.

I would have thought that the `all()` call would cover this case!

integ/src/monitor.rs
```diff
@@ -90,23 +90,30 @@ impl<T: BrupopClient> BrupopMonitor<T> {
...
+        } else {
+            return pods.iter().all(|pod| is_pod_running(pod));
+        }
     }
 
     // verify if brs has been created properly and initialized `status`.
     async fn check_shadows_health(
         &self,
         bottlerocketshadows: &ObjectList<BottlerocketShadow>,
     ) -> Result<bool> {
```
cbgbt: I think we don't need to return a `Result` here, since it seems like the function can't fail. If that's the case, you might be able to simplify this using the iterator pattern you used above.

integ/src/monitor.rs
gthao313: I already have an unit test for [check_pods_health](https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/develop/integ/src/monitor.rs#L305-L343) function : )

gthao313:
Push above address Sean's comments.

cbgbt (approved):

somnusfish (approved):

Issue #14, opened 2019-11-11
Protect controller from becoming unscheduleable


jahkeup:
Also, the Operator Controller should handle killing itself, especially so in a single noded cluster! It should not prevent itself from getting scheduled in a Cluster.

_Originally posted by @jahkeup in https://github.com/amazonlinux/PRIVATE-thar/pull/239#issuecomment-552590081_

jahkeup:
One way to protect the controller could be to have the controller save its hosting node for to be updated last. Once it updated through the other nodes, the controller would delete its Pod to be rescheduled and only once started elsewhere would it continue to update that last node.

The controller's deployment should then include `bottlerocket.aws/update-available` in its `antiAffinity` weighted selector (preferring `update-available==false`) so that it lands on updated hosts first.

This method wouldn't account for a single noded cluster or one where only *a node* was Ready and Schedulable. The controller will have to check that it considers itself to be reschedulable prior to stopping its Pod.

cbgbt:
Some thoughts from conversation with @somnusfish:
* Consider evicting the controller first when doing a drain
* Add a timeout to drains, error out and trigger our new crash loop handling code (#123) if they get stuck. Drains should never roll forward on timeouts.
* Add PDBs to apiserver deployment to ensure we always have at least 2 running in the cluster.

cbgbt:
We want to add the ability to allow brupop to update many nodes simultaneously, which makes this more important. Adding this to the 1.0.0 release milestone.

PR #214, opened 2022-06-21
Protect controller from becoming unschedulable

gthao313:
**Issue number:**
#14 

**Description of changes:**
```
Date:   Tue Jun 21 03:03:50 2022 +0000

    Make the controller a high-priority

    As part of Protect controller from becoming unschedulable, this approach
    means to give controller a high priority so that controller may be scheduled
    sooner than Pods with lower priority if its scheduling requirements are met.

commit c7716c2109894f41c2b2ab960c5cfcd8c1f5b489
Date:   Tue Jun 21 03:00:36 2022 +0000

    Update the node which hosts controller last

    As part of Protect controller from becoming unschedulable, this part can
    make brupop Update the node which hosts controller last. Therefore, this
    approach can reduce the effect when controller pods is dead as much as possible.

commit 7eff6e627247d58c574bf1c105a065d43a2bebe3
Date:   Mon Jun 20 21:48:47 2022 +0000

    redesign the responsibilities of controller and agent

    As part of prevent controller from becoming unschedualable, we redesign
    the responsibilities of controler and agent. Basically new design makes
    agent be able to complete a node update even lack controller's signal.
```

**Testing done:**

- [x] Test controller state machine
- [x] Test if the node which hosts controller pod update last
- [x] Make sure controller pod has desired priority value. 

The node which hosts controller pod is ` ip-192-168-131-7.us-west-2.compute.internal
The priority class value is `1000000`

```
Name:                 brupop-controller-deployment-8484df4fb4-vcjxb
Namespace:            brupop-bottlerocket-aws
Priority:             100000
Priority Class Name:  brupop-controller-high-priority
Node:                  ip-192-168-131-7.us-west-2.compute.internal/192.168.129.76
```

```
[kubectl -n brupop-bottlerocket-aws get brs
NAME                                                STATE   VERSION   TARGET STATE               TARGET VERSION   CRASH COUNT
brs-ip-192-168-131-7.us-west-2.compute.internal     Idle    1.6.1     Idle                       <no value>       0
brs-ip-192-168-134-90.us-west-2.compute.internal    Idle    1.6.1     Idle                       <no value>       0
brs-ip-192-168-137-157.us-west-2.compute.internal   Idle    1.6.1     StagedAndPerformedUpdate   1.8.0            0
```

**Terms of contribution:**

### DIFF
--- a/agent/src/agentclient.rs
+++ b/agent/src/agentclient.rs
@@ -362,15 +362,12 @@ impl<T: APIServerClient> BrupopAgent<T> {
                         );
                     }
                 },
-                BottlerocketShadowState::StagedUpdate => {
+                BottlerocketShadowState::StagedAndPerformedUpdate => {
                     event!(Level::INFO, "Preparing update");
                     prepare().await.context(agentclient_error::UpdateActions {
                         action: "Prepare".to_string(),
                     })?;
 
-                    self.cordon_and_drain().await?;
-                }
-                BottlerocketShadowState::PerformedUpdate => {
                     event!(Level::INFO, "Performing update");
                     update().await.context(agentclient_error::UpdateActions {
                         action: "Perform".to_string(),
@@ -390,7 +387,9 @@ impl<T: APIServerClient> BrupopAgent<T> {
                             ),
                         )
                         .await?;
+                        self.handle_recover().await?;
                     } else {
+                        self.cordon_and_drain().await?;
                         boot_update()
                             .await
                             .context(agentclient_error::UpdateActions {
@@ -400,9 +399,8 @@ impl<T: APIServerClient> BrupopAgent<T> {
                 }
                 BottlerocketShadowState::MonitoringUpdate => {
                     event!(Level::INFO, "Monitoring node's healthy condition");
-                    self.handle_recover().await?;
-                    // TODO: we need add some criterias here by which we decide to transition
-                    // from MonitoringUpdate to WaitingForUpdate.
+                    // TODO: we left space here for customer if they need add customized criteria
+                    // which uses to decide to transition from MonitoringUpdate to WaitingForUpdate.
                 }
                 BottlerocketShadowState::ErrorReset => {
                     // Spec state should never be ErrorReset
--- a/controller/src/controller.rs
+++ b/controller/src/controller.rs
@@ -4,7 +4,8 @@ use super::{
     statemachine::determine_next_node_spec,
 };
 use models::node::{
-    BottlerocketShadow, BottlerocketShadowClient, BottlerocketShadowState, Selector,
+    brs_name_from_node_name, BottlerocketShadow, BottlerocketShadowClient, BottlerocketShadowState,
+    Selector,
 };
 
 use kube::runtime::reflector::Store;
@@ -12,6 +13,7 @@ use kube::ResourceExt;
 use opentelemetry::global;
 use snafu::ResultExt;
 use std::collections::{BTreeMap, HashMap};
+use std::env;
 use tokio::time::{sleep, Duration};
 use tracing::{event, instrument, Level};
 
@@ -104,16 +106,16 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
     /// The state transition is then attempted. If successful, this node should be detected as part of the active
     /// set during the next iteration of the controller's event loop.
     #[instrument(skip(self))]
-    async fn find_and_update_ready_node(&self) -> Option<BottlerocketShadow> {
+    async fn find_and_update_ready_node(&self) -> Result<Option<BottlerocketShadow>> {
         let mut shadows: Vec<BottlerocketShadow> = self.all_nodes();
 
-        shadows.sort_by(|a, b| a.compare_crash_count(b));
-        for brs in shadows {
+        sort_shadows(&mut shadows, &get_associated_bottlerocketshadow_name()?);
+        for brs in shadows.drain(..) {
             // If we determine that the spec should change, this node is a candidate to begin updating.
             let next_spec = determine_next_node_spec(&brs);
             if next_spec != brs.spec {
                 match self.progress_node(brs.clone()).await {
-                    Ok(_) => return Some(brs),
+                    Ok(_) => return Ok(Some(brs)),
                     Err(_) => {
                         // Errors connecting to the k8s API are ignored (and also logged by `progress_node()`).
                         // We'll just move on and try a different node.
@@ -122,7 +124,7 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
                 }
             }
         }
-        None
+        Ok(None)
     }
 
     #[instrument(skip(self))]
@@ -189,7 +191,7 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
                 }
             } else {
                 // If there's nothing to operate on, check to see if any other nodes are ready for action.
-                let new_active_node = self.find_and_update_ready_node().await;
+                let new_active_node = self.find_and_update_ready_node().await?;
                 if let Some(brs) = new_active_node {
                     event!(Level::INFO, name = %brs.name(), "Began updating new node.")
                 }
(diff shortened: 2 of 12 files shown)


agent/src/agentclient.rs
```diff
@@ -362,15 +362,12 @@ impl<T: APIServerClient> BrupopAgent<T> {
                         );
                     }
                 },
-                BottlerocketShadowState::StagedUpdate => {
+                BottlerocketShadowState::StagedAndPerformUpdate => {
```
cbgbt: nit: verb tense
```suggestion
                BottlerocketShadowState::StagedAndPerformedUpdate => {
```

agent/src/agentclient.rs
```diff
@@ -362,15 +362,12 @@ impl<T: APIServerClient> BrupopAgent<T> {
...
                     event!(Level::INFO, "Preparing update");
                     prepare().await.context(agentclient_error::UpdateActions {
                         action: "Prepare".to_string(),
                     })?;
```
cbgbt, line 369: Is it safe to call `prepare()` twice? In case this fails transiently, and we must call it again to get to `update()`.

models/src/constants.rs
```diff
@@ -47,3 +47,6 @@ pub const CONTROLLER_DEPLOYMENT_NAME: &str = "brupop-controller-deployment";
 pub const CONTROLLER_SERVICE_NAME: &str = "brupop-controller-server"; // The name for the `svc` fronting the controller.
 pub const CONTROLLER_INTERNAL_PORT: i32 = 8080; // The internal port on which the the controller service is hosted.
 pub const CONTROLLER_SERVICE_PORT: i32 = 80; // The k8s service port hosting the controller service.
+pub const BRUPOP_CONTROLLER_PRIORITY_CLASS: &str = "brupop-controller-high-priority";
+pub const BRUPOP_CONTROLLER_PREEMPTION_POLICY: &str = "Never";
+pub const BRUPOP_CONTROLLER_PRIORITY_VALUE: i32 = 1000000; // The controller priority class value
```
cbgbt: How is this priority value determined?

controller/src/controller.rs
```diff
@@ -104,16 +106,16 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
...
     /// set during the next iteration of the controller's event loop.
     #[instrument(skip(self))]
-    async fn find_and_update_ready_node(&self) -> Option<BottlerocketShadow> {
-        let mut shadows: Vec<BottlerocketShadow> = self.all_nodes();
+    async fn find_and_update_ready_node(&self) -> Result<Option<BottlerocketShadow>> {
+        let shadows: Vec<BottlerocketShadow> = self.all_nodes();
 
-        shadows.sort_by(|a, b| a.compare_crash_count(b));
-        for brs in shadows {
+        sort_shadows(shadows.clone(), &get_associated_bottlerocketshadow_name()?)?;
```
cbgbt: This won't actually sort the vector, since we are cloning it as input rather than using a mutable reference.

controller/src/controller.rs
```diff
@@ -203,3 +205,138 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
...
+fn sort_shadows(
+    mut shadows: Vec<BottlerocketShadow>,
+    associated_brs_name: &str,
+) -> Result<Vec<BottlerocketShadow>> {
+    let mut associated_brs_node_position: usize = usize::MAX;
+    // sort shadows by crash count
+    shadows.sort_by(|a, b| a.compare_crash_count(b));
+
+    // move the shadow which associated bottlerocketshadow node hosts controller pod to the last
+    for (position, brs) in shadows.iter().enumerate() {
```
cbgbt: nit: Consider using `shadows.iter().position()`. This would also allow you to use an `Option` rather than having to use `usize::MAX` as a sentinel value.

controller/src/controller.rs
```diff
@@ -203,3 +205,138 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
...
+            associated_brs_node_position = position;
+            break;
+        }
+    }
+
+    // if it doesn't find the brs, it means those brss aren't ready and the program should skip sort.
+    if associated_brs_node_position != usize::MAX {
+        let last_brs = shadows[associated_brs_node_position].clone();
+        shadows.remove(associated_brs_node_position.clone());
+        shadows.push(last_brs.clone());
```
cbgbt: I don't think a clone here is necessary, since `last_brs` was cloned out of the vector and ownership is being pushed back into it here.

controller/src/controller.rs
gthao313: Yeah, I remember I changed it to `let sorted_shadows = sort_shadows(shadows.clone(), &get_associated_bottlerocketshadow_name()?)?;`. Maybe I was doing wrong with rebase. Thanks I'll fix it.

models/src/constants.rs
gthao313: The value can be set between 0 to 1 billion. The reasons are:
1.  I assume controller pod is needed to come back soon than other pods except critical pods like system related. I went through priorityclass document, and they usually set 1 million to present a high priority value but not critical.
2. We don't know what priority strategies the customers are using to set up their k8s resources. I don't want to make controller take precedence over customers' critical k8s resources. After looking a lot priorityclass github examples, they usually set their high priority and critical values larger than 10 million. So for more safety I downgrade the value to 1 million here.

agent/src/agentclient.rs
gthao313: I believe it's safe. I manually tested this behavior on a BR host and that showed pretty safe to run `prepare()` twice or more.
The test method: 
```
step1: run `apiclient raw -u /actions/refresh-updates -m POST` and `apiclient get /updates/status`
step2: run `apiclient raw -u /actions/prepare-update -m POST`
step1: run `apiclient raw -u /actions/refresh-updates -m POST` and `apiclient get /updates/status`
step4: run `apiclient raw -u /actions/prepare-update -m POST`
step4: run apiclient raw -u /actions/activate-update -m POST
```

controller/src/controller.rs
gthao313: Oh, i just remembered why I didn't use `shadows.iter().position()`. When I was trying to do like this, the position function showed it was unable to handle result value as input. Actually the position function can't handle ` result` which is `brs.metadata.name.as_ref().context(error::MissingBottlerocketShadowName)?`
```
let associated_brs_node_position = shadows.iter().position(|brs| brs.metadata.name.as_ref().context(error::MissingBottlerocketShadowName)? == associated_brs_name).context("")?
```

models/src/constants.rs
cbgbt: It might be good to elaborate a bit on the constant in a comment.

controller/src/controller.rs
cbgbt: Since you have `kube::ResourceExt` imported, you could just do `brs.name()` which technically panics if the name is missing. Not great, but there are likely bigger problems if the name is missing from objects.

Alternatively, if the name is missing you can probably skip the item rather than returning early:

```rust
let associated_brs_node_position = shadows
    .iter()
    .position(|brs| brs.metadata.name.as_ref() == Some(associated_brs_name))
```

I think it's best to lean on `Option` which the compiler can check rather than using a sentinel value, even though we can logically deduce that your sentinel would work.

controller/src/controller.rs
cbgbt: I think it would be more idiomatic here to use a mutable reference as input to the function. You can just return an empty `Result<()>` to indicate whether something or not went wrong.

controller/src/controller.rs
```diff
@@ -104,16 +106,16 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
...
     /// set during the next iteration of the controller's event loop.
     #[instrument(skip(self))]
-    async fn find_and_update_ready_node(&self) -> Option<BottlerocketShadow> {
+    async fn find_and_update_ready_node(&self) -> Result<Option<BottlerocketShadow>> {
         let mut shadows: Vec<BottlerocketShadow> = self.all_nodes();
 
-        shadows.sort_by(|a, b| a.compare_crash_count(b));
-        for brs in shadows {
+        sort_shadows(&mut shadows, &get_associated_bottlerocketshadow_name()?)?;
+        for brs in shadows.drain(..) {
```
gthao313, line 113: Update to use `drain`

controller/src/controller.rs
```diff
@@ -203,3 +205,188 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
...
+    );
+
+    Ok(associated_bottlerocketshadow_name)
+}
+
+/// sort shadows list which uses to determine the order of node update
+/// logic1: sort shadows by crash count
+/// logic2: move the shadow which associated bottlerocketshadow node hosts controller pod to the last
+#[instrument(skip())]
+fn sort_shadows(shadows: &mut Vec<BottlerocketShadow>, associated_brs_name: &str) -> Result<()> {
```
cbgbt: nit: I think this function cannot return an error value. You can probably remove the return type.

cbgbt (approved):

cbgbt (approved):

controller/src/error.rs
```diff
@@ -37,4 +37,7 @@ pub enum Error {
 
     #[snafu(display("Controller failed due to internal assertion issue: '{}'", source))]
     Assertion { source: serde_plain::Error },
+
+    #[snafu(display("Unable to get host controller pod node name: {}", source))]
+    GetNodeName { source: std::env::VarError },
```
somnusfish, line 42: Just a note here this `GetNodeName` didn't show up in the [adjust error module](https://github.com/bottlerocket-os/bottlerocket-update-operator/pull/223/files), you might want to reimplement this error when you merge two PRs.

gthao313: Yeah! I would like to merge this PR first and then solve the conflict on the PR you mentioned above. Thanks!

somnusfish (approved):

Issue #216, opened 2022-06-24
Brupop integ test unable to clean priorityclass 


gthao313:
Now we create more brupop resources when we install brupop like `priorityclass` and `cert`, and we should find a more generic method to clean up all created resources when running integration test. 

Idea:
run `kubectl delete -f bottlerocket-update-operator.yaml` instead of independently cleaning each resource.

Issue #208, opened 2022-06-13
Skip test pods creation if they already exit. 


gthao313:
When I create more than one nodegroups to test brupop, the error `Failed to create pods on eks cluster:` jumped. This would not break the integration test but we should fix. Skip the test pods creation if `services "nginx"`, statefulsets.apps "web-test",  deployments.apps "nginx-test", and poddisruptionbudgets.policy "pod-disruption-budget-test" already exists.
```
Error from server (AlreadyExists): error when creating "integ/src/pods-template.yaml": services "nginx" already exists
Error from server (AlreadyExists): error when creating "integ/src/pods-template.yaml": statefulsets.apps "web-test" already exists
Error from server (AlreadyExists): error when creating "integ/src/pods-template.yaml": deployments.apps "nginx-test" already exists
Error from server (AlreadyExists): error when creating "integ/src/pods-template.yaml": poddisruptionbudgets.policy "pod-disruption-budget-test" already exists
Failed to create pods on eks cluster: Failed to process pods test: "Create" pods
```

PR #217, opened 2022-06-28
Integ test - creation/deletion of brupop and test pods improvement

gthao313:
**Issue number:**
#216
#208

**Description of changes:**
```
Date:   Tue Jun 28 01:00:14 2022 +0000

    skip creating brupop resources if they already exist

    Integration test allows to launch multiple nodegroups on eks cluster,
    and it would be redundant to create brupop resources multiple times.
    Therefore this change make system to skip creating brupop resources
    if they already exist.

    Meanwhile, simplify the functionality of create and delete brupop
    resources.

commit 55587205f41633d823b3bef0ccaa91216f1b98cf
Date:   Tue Jun 28 00:33:44 2022 +0000

    skip creating test pods if they already exist

    Integration test allows to launch multiple nodegroups on eks cluster,
    and it would be redundant to create test pods multiple times. Therefore
    this change make system to skip creating test pods if they already
    exist.
```

**Testing done:**
Method: Launch multiple nodegroup and check if brupop resources and test pods are created only once (first nodegroup)
```
cargo run --bin integ integration-test --cluster-name brupop-ipv4 --region us-west-2 --bottlerocket-version 1.6.1  --arch x86_64 --nodegroup-name x86

cargo run --bin integ integration-test --cluster-name brupop-ipv4 --region us-west-2 --bottlerocket-version 1.6.1  --arch arm64 --nodegroup-name arm64
```

**Terms of contribution:**

### DIFF
--- a/integ/src/updater.rs
+++ b/integ/src/updater.rs
@@ -4,7 +4,6 @@
 !*/
 
 use lazy_static::lazy_static;
-
 use snafu::{ensure, ResultExt};
 use std::process::Command;
 
@@ -39,7 +38,7 @@ lazy_static! {
 
 #[derive(strum_macros::Display, Debug)]
 pub enum Action {
-    Create,
+    Apply,
     Delete,
 }
 
@@ -121,7 +120,9 @@ pub async fn delete_brupop_cluster_resources(kube_config_path: &str) -> UpdaterR
     Ok(())
 }
 
-// create/delete statefulset pods, stateless nginx pods, and pods with PDBs on EKS cluster
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^= Deletion and Creation of test pods  =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+// create or delete statefulset pods, stateless nginx pods, and pods with PDBs on EKS cluster
 pub async fn process_pods_test(action: Action, kube_config_path: &str) -> UpdaterResult<()> {
     let action_string: String = action.to_string();
 
@@ -144,7 +145,6 @@ pub async fn process_pods_test(action: Action, kube_config_path: &str) -> Update
             action: action_string
         }
     );
-
     Ok(())
 }
 
--- a/integ/src/updater.rs
+++ b/integ/src/updater.rs
@@ -2,39 +2,15 @@
   updater helps running brupop on existing EKS cluster and clean up all
   resources once completing integration test
 !*/
-
-use lazy_static::lazy_static;
 use snafu::{ensure, ResultExt};
 use std::process::Command;
 
 use k8s_openapi::api::core::v1::Node;
 use kube::api::{Api, ListParams};
 
-use models::constants::NAMESPACE;
-
 const CURRENT_LOCATION_PATH: &str = "integ/src";
 const PODS_TEMPLATE: &str = "pods-template.yaml";
 const KUBECTL_BINARY: &str = "kubectl";
-
-lazy_static! {
-    static ref BRUPOP_CLUSTER_ROLES: Vec<&'static str> = {
-        let mut m = Vec::new();
-        m.push("brupop-apiserver-role");
-        m.push("brupop-agent-role");
-        m.push("brupop-controller-role");
-        m
-    };
-}
-
-lazy_static! {
-    static ref BRUPOP_CLUSTER_ROLE_BINDINGS: Vec<&'static str> = {
-        let mut m = Vec::new();
-        m.push("brupop-apiserver-role-binding");
-        m.push("brupop-agent-role-binding");
-        m.push("brupop-controller-role-binding");
-        m
-    };
-}
 
 #[derive(strum_macros::Display, Debug)]
 pub enum Action {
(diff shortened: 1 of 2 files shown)


integ/src/updater.rs
```diff
@@ -3,149 +3,103 @@
...
+        return Ok(());
+    } else {
+        // while creating brupop resources it's possible that test pods already exist on the eks cluster;
+        //therefore, skips test pods creation if those pods already exist.
+        // This wouldn't affect pods deletion.
+        let error_content = String::from_utf8_lossy(&brupop_resource_output.stderr).to_string();
+        match error_content.contains("AlreadyExists") {
+            true => {
+                info!("brupop resources have been created, skip...");
+                return Ok(());
```
cbgbt: Is it possible for us to check that those resources are using the same code as what we're currently trying to test? That might be useful info to someone running tests.

integ/src/updater.rs
```diff
@@ -3,149 +3,103 @@
...
-            update_error::BrupopRun
-        );
+    if brupop_resource_output.status.success() {
+        return Ok(());
+    } else {
+        // while creating brupop resources it's possible that test pods already exist on the eks cluster;
+        //therefore, skips test pods creation if those pods already exist.
+        // This wouldn't affect pods deletion.
+        let error_content = String::from_utf8_lossy(&brupop_resource_output.stderr).to_string();
+        match error_content.contains("AlreadyExists") {
```
cbgbt: nit: IMO it's simpler to just use `if` instead of `match` on a boolean expression.

integ/src/updater.rs
gthao313: "the same code" here you mean is `bottlerocekt-update-operator.yaml`?  For example, developers are going to test some new changes on brupop resources, but due to the old brupop resources had been created, so when run a new integration test, it will skip the new brupop resources installing. Is my understand correct here or ? : )

cbgbt: Yes! That's what I was thinking of.

cbgbt: I chatted with Tianhao about the specifics of this: basically test resources should always be cleaned between runs, and shared resources are otherwise acceptable if testing for multiple architectures. I'm okay with proceeding with this implementation.

gthao313:
Ideally, we have to clean up integration test when we need test a code change. However, during brupop release process, we need test arm and amd at same time (run`cargo run --bin integ integration-test` to add two nodegroups to cluster), and this will ask integration test to create brupop resources and test pods twice. Therefore, these process would fail on `resources Already exist`. 

Solution: run `kubectl apply -f xxx` instead of `kubectl create -f xxx`.
`apply` intelligently skips the duplicate resources creation and only configure new or changed resources. 
```
service/nginx unchanged
statefulset.apps/web-test configured
deployment.apps/nginx-test unchanged
poddisruptionbudget.policy/pod-disruption-budget-test unchanged
customresourcedefinition.apiextensions.k8s.io/bottlerocketshadows.brupop.bottlerocket.aws configured
namespace/brupop-bottlerocket-aws unchanged
clusterissuer.cert-manager.io/selfsigned-issuer unchanged
issuer.cert-manager.io/my-ca-issuer unchanged
certificate.cert-manager.io/brupop-apiserver unchanged
serviceaccount/brupop-apiserver-service-account unchanged
clusterrole.rbac.authorization.k8s.io/brupop-apiserver-role unchanged
clusterrolebinding.rbac.authorization.k8s.io/brupop-apiserver-role-binding unchanged
clusterrolebinding.rbac.authorization.k8s.io/brupop-apiserver-auth-delegator-role-binding unchanged
deployment.apps/brupop-apiserver configured
service/brupop-apiserver unchanged
serviceaccount/brupop-agent-service-account unchanged
clusterrole.rbac.authorization.k8s.io/brupop-agent-role unchanged
clusterrolebinding.rbac.authorization.k8s.io/brupop-agent-role-binding unchanged
daemonset.apps/brupop-agent configured
serviceaccount/brupop-controller-service-account unchanged
clusterrole.rbac.authorization.k8s.io/brupop-controller-role unchanged
clusterrolebinding.rbac.authorization.k8s.io/brupop-controller-role-binding unchanged
deployment.apps/brupop-controller-deployment unchanged
service/brupop-controller-server unchanged
```

Hence, more than once integration test at same time will not fail on resource `alreadyexit`, BUT we don't suggest to test multiple brupop code changes simultaneously.

cbgbt (approved):

integ/src/updater.rs
```diff
@@ -2,52 +2,31 @@
...
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^= Deletion and Creation of brupop resources  =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+// install or destroy all brupop resources
+pub async fn process_brupop_resources(action: Action, kube_config_path: &str) -> UpdaterResult<()> {
+    let action_string: String = action.to_string();
+
     let brupop_resource_status = Command::new(KUBECTL_BINARY)
         .args([
-            "apply",
+            &action_string.to_lowercase(),
```
somnusfish, line 29: How do you plan to handle deletion failure? I know sometimes the deletion get stuck and need to manually patch the resource especially during the CRD update. Does it make sense to set up a timeout on deletion and print the manual patch command?

gthao313: Can you share what situations you had met?

somnusfish: We had a discussion offline and decided to open an new issue if the above situation happens.

somnusfish (approved):

PR #221, opened 2022-07-11
Add Webhooks logic for CRD migration

somnusfish:
**Issue number:**
#161 

**Description of changes:**
Add Webhooks in API Server and implemented the corresponding logic to convert different custom resources version.

Changes include:
* Update logic in CRD generation so the webhook config will be generated in yaml file.
* Implemented the BottlerocketShadow v1 to v2 convert in models.
* Defined `ConversionRequest` and `ConversionResponse` as the object webhook talk with Kubernetes API server
* Added chained convert logic and registered webhook on Brupop API Server.

**Testing done:**
* Unit test
* Provisioned a cluster with old Bottlerocket OS, deployed 0.2.1 on it, wait Brupop create CRDs with v1 and deploy the webhook to the cluster, monitor and wait the cluster complete upgrade to confirm webhook work as expected.

**Terms of contribution:**

### DIFF
--- a/apiserver/src/api/node.rs
+++ b/apiserver/src/api/node.rs
@@ -1,15 +1,18 @@
 use super::{APIServerSettings, ApiserverCommonHeaders};
 use crate::error::{self, Result};
+use crate::webhook::ConversionRequest;
+
 use models::node::{BottlerocketShadowClient, BottlerocketShadowStatus};
 
 use actix_web::{
     web::{self},
     HttpRequest, HttpResponse, Responder,
 };
+
 use serde_json::json;
 use snafu::ResultExt;
-
 use std::convert::TryFrom;
+use tracing::{event, Level};
 
 /// HTTP endpoint which creates BottlerocketShadow custom resources on behalf of the caller.
 pub(crate) async fn create_bottlerocket_shadow_resource<T: BottlerocketShadowClient>(
@@ -42,14 +45,31 @@ pub(crate) async fn update_bottlerocket_shadow_resource<T: BottlerocketShadowCli
     Ok(HttpResponse::Ok().body(format!("{}", json!(&node_status))))
 }
 
+pub(crate) async fn convert_bottlerocket_shadow_resource(
+    conversion_req: web::Json<ConversionRequest>,
+) -> Result<impl Responder> {
+    event!(Level::INFO, ?conversion_req, "Original conversion request");
+    let response = conversion_req.convert_resource();
+    let response_string = serde_json::to_string(&response).context(error::WebhookError)?;
+    event!(Level::INFO, ?response_string, "Converted response:");
+
+    // Webhook will always respond with 200.
+    // The actual status of the conversion will be returned in
+    // ConversionResponse.response.result.status
+    Ok(HttpResponse::Ok()
+        .content_type("application/json")
+        .body(response_string))
+}
+
 #[cfg(test)]
 mod tests {
     use super::super::tests::test_settings;
     use super::*;
     use crate::constants::{
-        HEADER_BRUPOP_K8S_AUTH_TOKEN, HEADER_BRUPOP_NODE_NAME, HEADER_BRUPOP_NODE_UID,
-        NODE_RESOURCE_ENDPOINT,
+        CRD_CONVERT_ENDPOINT, HEADER_BRUPOP_K8S_AUTH_TOKEN, HEADER_BRUPOP_NODE_NAME,
+        HEADER_BRUPOP_NODE_UID, NODE_RESOURCE_ENDPOINT,
     };
+    use crate::webhook::{ConversionRequest, ConversionResponse, Request};
     use models::node::{
         BottlerocketShadow, BottlerocketShadowSelector, BottlerocketShadowSpec,
         BottlerocketShadowState, MockBottlerocketShadowClient, Version,
@@ -178,6 +198,66 @@ mod tests {
             let return_status: BottlerocketShadowStatus =
                 serde_json::from_slice(&b).expect("Could not parse JSON response.");
             assert_eq!(return_status, node_status);
+        } else {
+            panic!("Response did not return a body.");
+        }
+    }
+
+    #[tokio::test]
+    async fn test_convert_crd() {
+        let conversion_req = ConversionRequest {
+            kind: "ConversionReview".to_string(),
+            api_version: "apiextensions.k8s.io/v1".to_string(),
+            request: Request {
+                uid: "5a6adc7e-c74b-43c0-9718-293de1b104cb".to_string(),
+                desired_api_version: "brupop.bottlerocket.aws/v2".to_string(),
+                objects: vec![json!({
+                    "apiVersion": "brupop.bottlerocket.aws/v1",
+                    "kind": "BottlerocketShadow",
+                    "metadata": {
+                        "name": "brs-ip-192-168-22-145.us-west-2.compute.internal",
+                        "namespace": "brupop-bottlerocket-aws",
+                        "uid": "3153df27-6619-4b6b-bc75-adbf92ef7266",
+                        "ownerReferences": [
+                            {
+                                "apiVersion": "v1",
+                                "kind": "Node",
+                                "name": "ip-192-168-22-145.us-west-2.compute.internal",
+                                "uid": "6b714046-3b20-4a79-aaa9-27cf626a2c12"
+                            }
+                        ]
+                    },
+                    "spec": {
+                        "state": "Idle",
+                    },
+                    "status": {
+                        "current_state": "Idle",
+                        "target_version": "1.8.0",
+                        "current_version": "1.8.0"
+                    }
+
+                })],
+            },
+        };
+        let req = test::TestRequest::put()
+            .uri(CRD_CONVERT_ENDPOINT)
+            .set_json(&conversion_req)
+            .to_request();
+
+        let app = test::init_service(App::new().route(
+            CRD_CONVERT_ENDPOINT,
+            web::put().to(convert_bottlerocket_shadow_resource),
+        ))
+        .await;
+
+        let resp = test::call_service(&app, req).await;
+
+        assert!(resp.status().is_success());
+        if let AnyBody::Bytes(b) = resp.into_body() {
+            // Only check the response body can be converted to ConversionResponse.
+            // Contents of the ConversionResponse should be tested in convert_resource method.
+            serde_json::from_slice::<ConversionResponse>(&b)
+                .expect("Could not parse JSON response.");
         } else {
             panic!("Response did not return a body.");
         }
--- a/apiserver/src/webhook/mod.rs
+++ b/apiserver/src/webhook/mod.rs
@@ -1,0 +1,416 @@
+mod request;
+mod response;
+
+pub use self::request::{ConversionRequest, Request};
+pub use self::response::{ConversionResponse, ConvertResult, Response};
+
+use models::node::v1::BottlerocketShadow as BottleRocketShadowV1;
+use models::node::v2::BottlerocketShadow as BottlerocketShadowV2;
+
+use snafu::{ResultExt, Snafu};
+use std::convert::TryFrom;
+use tracing::instrument;
+
+pub type Result<T> = std::result::Result<T, WebhookConvertError>;
+
+/// Convert k8s ConversionReview object from request to response
+/// by applying chained convert methods on its objects.
+///
+/// Sample request in yaml format:
+#[cfg_attr(doctest, doc = " ````no_test")]
+/// ```
+/// {
+///     "apiVersion": "apiextensions.k8s.io/v1",
+///     "kind": "ConversionReview",
+///     "request": {
+///         # Random uid uniquely identifying this conversion call
+///         "uid": "5a6adc7e-c74b-43c0-9718-293de1b104cb",
+///
+///         # The API group and version the objects should be converted to
+///         "desiredAPIVersion": "brupop.bottlerocket.aws/v2",
+///
+///         # The list of objects to convert.
+///         # May contain one or more objects, in one or more versions.
+///         "objects": [
+///             {
+///                 "kind": "BottlerocketShadow",
+///                 "apiVersion": "brupop.bottlerocket.aws/v1",
+///                 "metadata": {
+///                     "name": "brs-ip-192-168-22-145.us-west-2.compute.internal",
+///                     "namespace": "brupop-bottlerocket-aws",
+///                     "uid": "3153df27-6619-4b6b-bc75-adbf92ef7266"
+///                 },
+///                 "spec": {
+///                     "state": "Idle",
+///                 },
+///                 "status": {
+///                     "current_state": "Idle",
+///                     "target_version": "1.8.0",
+///                     "current_version": "1.8.0"
+///                 }
+///             }
+///         ]
+///
+///     }
+/// }
+/// ```
+/// Sample response in yaml format:
+#[cfg_attr(doctest, doc = " ````no_test")]
+/// ```
+/// {
+///     "apiVersion": "apiextensions.k8s.io/v1",
+///     "kind": "ConversionReview",
+///     "response": {
+///         # must match <request.uid>
+///         "uid": "5a6adc7e-c74b-43c0-9718-293de1b104cb",
+///
+///         "result": {
+///             "status": "Success"
+///         },
+///
+///         # Objects must match the order of request.objects, and have apiVersion set to <request.desiredAPIVersion>.
+///         # kind, metadata.uid, metadata.name, and metadata.namespace fields must not be changed by the webhook.
+///         # metadata.labels and metadata.annotations fields may be changed by the webhook.
+///         # All other changes to metadata fields by the webhook are ignored.
+///         "objects": [
+///             {
+///                 "kind": "BottlerocketShadow",
+///                 "apiVersion": "brupop.bottlerocket.aws/v2",
+///                 "metadata": {
+///                     "name": "brs-ip-192-168-22-145.us-west-2.compute.internal",
+///                     "namespace": "brupop-bottlerocket-aws",
+///                     "uid": "3153df27-6619-4b6b-bc75-adbf92ef7266"
+///                 },
+///                 "spec": {
+///                     "state": "Idle",
+///                 },
+///                 "status": {
+///                     "current_state": "Idle",
+///                     "target_version": "1.8.0",
+///                     "current_version": "1.8.0"
+///                 }
+///             }
+///         ]
+///
+///     }
+/// }
+/// ```
+pub fn convert_request_to_response(req: &ConversionRequest) -> ConversionResponse {
+    let request = &req.request;
+    let desired_version = request.desired_api_version.clone();
+
+    match convert_objects(desired_version, request.objects.clone()) {
+        Ok(new_objects) => {
+            let response = Response {
+                uid: request.uid.clone(),
+                result: ConvertResult::default(),
+                converted_objects: Some(new_objects),
+            };
+            ConversionResponse {
+                kind: req.kind.clone(),
+                api_version: req.api_version.clone(),
+                response: response,
+            }
+        }
+        Err(e) => {
+            let fail_result = ConvertResult::create_fail_result(e.to_string());
+            let response = Response {
+                uid: request.uid.clone(),
+                result: fail_result,
+                converted_objects: None,
+            };
+            ConversionResponse {
+                kind: req.kind.clone(),
+                api_version: req.api_version.clone(),
+                response: response,
+            }
+        }
+    }
+}
+
+#[instrument(err)]
+fn convert_objects(
+    desired_version: String,
+    objects: Vec<serde_json::Value>,
+) -> Result<Vec<serde_json::Value>> {
+    let mut new_objects = Vec::new();
+    for old_object in objects.into_iter() {
+        let old_brs_object = BRSObject { object: old_object };
+        let new_brs_object = old_brs_object.chained_convert_object(desired_version.clone())?;
+        new_objects.push(new_brs_object.object);
+    }
+    Ok(new_objects)
+}
+
+/// An abstraction over BottlerocketShadow's json value.
+/// Its implementation contains the logic to chain convert BottlerocketShadow
+/// to a different version.
+///
+/// To add a new version convert, first add a method build the logic
+/// to convert from previous version like:
+#[cfg_attr(doctest, doc = " ````no_test")]
+/// ```
+/// fn to_v2(source_obj: BRSObject) -> Result<BRSObject> {
+///     Self::try_from(BottlerocketShadowV2::from(BottleRocketShadowV1::try_from(
+///         source_obj,
+///     )?))
+/// }
+/// ```
+///
+/// Then update `convert_to_next_version` to map the
+/// BottlerocketShadow version to the above method.
+///
+struct BRSObject {
+    pub object: serde_json::Value,
+}
+
+impl BRSObject {
+    fn get_version(&self) -> Result<String> {
+        Ok(serde_json::from_value(self.object["apiVersion"].clone())
+            .context(SourceVersionNotExistInRequest)?)
+    }
+
+    fn to_v2(source_obj: BRSObject) -> Result<BRSObject> {
+        Self::try_from(BottlerocketShadowV2::from(BottleRocketShadowV1::try_from(
+            source_obj,
+        )?))
+    }
+
+    fn convert_to_next_version(self) -> Result<Self> {
+        let version = self.get_version()?;
+        match version.as_str() {
+            "brupop.bottlerocket.aws/v1" => BRSObject::to_v2(self),
+            _ => InvalidVersionError { version }.fail(),
+        }
+    }
+
+    #[instrument(skip(self), err)]
+    fn chained_convert_object(self, desired_version: String) -> Result<Self> {
+        let mut version = self.get_version()?;
+        let mut source_object = self;
+
+        while version != desired_version {
+            match source_object.convert_to_next_version() {
+                Ok(val) => source_object = val,
+                Err(_) => {
+                    return ChainedConvertError {
+                        src_version: version,
+                        dst_version: desired_version,
+                    }
+                    .fail()
+                }
+            }
+            version = source_object.get_version()?;
+        }
+
+        Ok(source_object)
+    }
+}
+
+impl TryFrom<BRSObject> for BottleRocketShadowV1 {
+    type Error = WebhookConvertError;
+
+    fn try_from(obj: BRSObject) -> Result<Self> {
+        serde_json::from_value(obj.object).context(JsonToBottlerocketShadowConvertError {
+            version: "v1".to_string(),
+        })
+    }
+}
+
+impl TryFrom<BRSObject> for BottlerocketShadowV2 {
+    type Error = WebhookConvertError;
+
+    fn try_from(obj: BRSObject) -> Result<Self> {
+        serde_json::from_value(obj.object).context(JsonToBottlerocketShadowConvertError {
+            version: "v2".to_string(),
+        })
+    }
+}
+
+impl TryFrom<BottlerocketShadowV2> for BRSObject {
+    type Error = WebhookConvertError;
+
+    fn try_from(shadow: BottlerocketShadowV2) -> Result<Self> {
+        Ok(BRSObject {
+            object: serde_json::to_value(shadow).context(BottlerocketShadowToJsonConvertError {
+                version: "v2".to_string(),
+            })?,
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{
+        convert_request_to_response, ConversionRequest, ConversionResponse, ConvertResult, Request,
+        Response,
+    };
+    use serde_json::json;
+
+    #[test]
+    fn test_convert_request_to_response_succeed() {
+        let conversion_req = ConversionRequest {
+            kind: "ConversionReview".to_string(),
+            api_version: "apiextensions.k8s.io/v1".to_string(),
+            request: Request {
+                uid: "5a6adc7e-c74b-43c0-9718-293de1b104cb".to_string(),
+                desired_api_version: "brupop.bottlerocket.aws/v2".to_string(),
+                objects: vec![json!({
+                    "apiVersion": "brupop.bottlerocket.aws/v1",
+                    "kind": "BottlerocketShadow",
+                    "metadata": {
+                        "name": "brs-ip-192-168-22-145.us-west-2.compute.internal",
+                        "namespace": "brupop-bottlerocket-aws",
+                        "uid": "3153df27-6619-4b6b-bc75-adbf92ef7266",
+                        "ownerReferences": [
+                            {
+                                "apiVersion": "v1",
+                                "kind": "Node",
+                                "name": "ip-192-168-22-145.us-west-2.compute.internal",
+                                "uid": "6b714046-3b20-4a79-aaa9-27cf626a2c12"
+                            }
+                        ]
+                    },
+                    "spec": {
+                        "state": "Idle",
+                    },
+                    "status": {
+                        "current_state": "Idle",
+                        "target_version": "1.8.0",
+                        "current_version": "1.8.0"
+                    }
+
+                })],
+            },
+        };
+
+        let expected_response = ConversionResponse {
+            kind: conversion_req.kind.clone(),
+            api_version: conversion_req.api_version.clone(),
+            response: Response {
+                uid: conversion_req.request.uid.clone(),
+                result: ConvertResult::default(),
+                converted_objects: Some(vec![json!({
+                    "apiVersion": "brupop.bottlerocket.aws/v2",
+                    "kind": "BottlerocketShadow",
+                    "metadata": {
+                        "name": "brs-ip-192-168-22-145.us-west-2.compute.internal",
+                        "namespace": "brupop-bottlerocket-aws",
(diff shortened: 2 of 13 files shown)


apiserver/src/webhook/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+    static ref TARGET_SET: HashSet<String> = {
+        let mut target_set = HashSet::new();
+        for value in CONVERTERS.values() {
+            target_set.insert(value.0.to_string());
+        }
+        target_set
+    };
+}
+
+pub fn convert_request_to_response(req: &ConversionRequest) -> ConversionResponse {
```
somnusfish, line 98: I would like hear some suggestions in this file to better struct the methods and some advice on the unit test.

models/src/node/crd/mod.rs
```diff
@@ -26,19 +26,39 @@
 /// };
 /// ```
 ///
+/// Add the BottlerocketShadow version to the end of BOTTLEROCKETSHADOW_CRD_VERSIONS
+/// So the webhook conversion could set up proper conversion_review_versions
+#[cfg_attr(doctest, doc = " ````no_test")]
+/// ```
+/// static ref BOTTLEROCKETSHADOW_CRD_VERSIONS: Vec<String> = {
```
gthao313: Should remove this?

models/src/node/crd/mod.rs
```diff
@@ -51,6 +72,13 @@ lazy_static! {
         crd_methods.push(v2::BottlerocketShadow::crd as fn()->CustomResourceDefinition);
         crd_methods
     };
+
```
cbgbt: Can we use the `vec!` macro for these?

apiserver/src/webhook/request.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub desired_api_version: String,
+    pub objects: Vec<serde_json::Value>,
+}
+
+impl ConversionRequest {
+    /// Wrap over convert_request_to_response so
+    /// actix_web::web::Json<ConversionRequest> could call the method
+    pub fn convert_to_response(&self) -> ConversionResponse {
+        convert_request_to_response(self)
+    }
```
cbgbt: I think `convert_resource()` may be a more precise name here.

models/src/node/crd/v2.rs
```diff
@@ -71,6 +76,18 @@ impl BottlerocketShadowState {
             Self::ErrorReset => ERROR_RESET_TIMEOUT,
         }
     }
+
+    fn from_v1_state(old_state: BottlerocketShadowState_v1) -> Self {
```
cbgbt: This is probably best expressed as

```rust
impl From<BottlerocketShadowState_v1> for BottlerocketShadowState {}
```

models/src/node/crd/v2.rs
```diff
@@ -184,6 +201,14 @@ impl BottlerocketShadowSpec {
         // We know this won't panic because we have a regex requirement on this attribute, which is enforced by the k8s schema.
         self.version.as_ref().map(|v| Version::from_str(v).unwrap())
     }
+
+    fn from_v1_spec(old_spec: BottlerocketShadowSpec_v1) -> Self {
```
cbgbt: Similarly probably best to use `From` here.

models/src/node/crd/mod.rs
```diff
@@ -121,11 +149,92 @@ fn combine_crd(
...
         .collect();
     let latest_crd = crds.pop().unwrap();
-    combine_crd(latest_crd, crds)
+    let combined_version_crds = combine_version_in_crds(latest_crd, crds);
+    add_webhook_setting(combined_version_crds)
+}
+
+pub type Result<T> = std::result::Result<T, CRDConvertError>;
+pub trait Convert {
+    fn convert_from_old_version(old_object: serde_json::Value) -> Result<serde_json::Value>;
```
cbgbt: nit:

```suggestion
    fn convert_from_previous_version(old_object: serde_json::Value) -> Result<serde_json::Value>;
```

models/src/node/crd/v2.rs
```diff
@@ -246,4 +271,209 @@ impl BottlerocketShadowStatus {
...
+                version: "v1".to_string(),
+            })?;
+        let old_metadata = old_shadow.metadata;
+        let old_spec = old_shadow.spec;
+        let old_status = old_shadow.status;
+
+        let status = match old_status {
+            None => None,
+            Some(old_status) => Some(BottlerocketShadowStatus::from_v1_status(old_status)),
+        };
```
cbgbt: If you end up implementing `From`, you can write this as something like:

```rust
let status = old_status.map(|s| s.into());
```

Rust should be able to deduce the type of `status` since it's used in the `BottlerocketShadow` creation below.

models/src/node/crd/mod.rs
```diff
@@ -121,11 +149,92 @@ fn combine_crd(
...
         .map(|crd_method| crd_method())
         .collect();
     let latest_crd = crds.pop().unwrap();
-    combine_crd(latest_crd, crds)
+    let combined_version_crds = combine_version_in_crds(latest_crd, crds);
+    add_webhook_setting(combined_version_crds)
+}
+
+pub type Result<T> = std::result::Result<T, CRDConvertError>;
+pub trait Convert {
```
cbgbt: Do we need backwards migrations as well? I'm imagining a case like:

* Controller is running version 1
* New agent comes up running version 2, creates V2 Custom Resource
* Controller cannot read newly created custom resource from new agent.

apiserver/src/webhook/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+use lazy_static::lazy_static;
+use models::node::Convert;
+use snafu::{ResultExt, Snafu};
+use std::collections::{HashMap, HashSet};
+use tracing::{event, Level};
+
+type ConvertMethod = fn(serde_json::Value) -> models::node::Result<serde_json::Value>;
+pub type Result<T> = std::result::Result<T, WebhookConvertError>;
+
+lazy_static! {
```
cbgbt: Some pointers on where these constants are used would be awesome.

apiserver/src/webhook/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+    let desired_version = request.desired_api_version.clone();
+
+    match convert_objects(desired_version, request.objects.clone()) {
+        Ok(new_objects) => {
+            let response = Response {
+                uid: request.uid.clone(),
+                result: ConvertResult::default(),
+                converted_objects: Some(new_objects),
+            };
+            return ConversionResponse {
```
cbgbt: nit: Can remove the `return`, since this `Response` will be the last expression in the function.

apiserver/src/webhook/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+            .context(SourceVersionNotExistInRequest)?;
+
+        let new_object =
+            chained_convert_objects(desired_version.clone(), original_version, old_object).unwrap();
+        new_objects.push(new_object);
+    }
+    Ok(new_objects)
+}
+
+fn chained_convert_objects(
```
cbgbt: You can [use `#[instrument(err)]` from `tracing` library](https://docs.rs/tracing/latest/tracing/attr.instrument.html) to automatically emit log messages on errors which get returned from functions, which is pretty great.

If you combine this [with `snafu::ensure!`](https://docs.rs/snafu/0.3.1/snafu/macro.ensure.html), you could probably simplify some of your log message + `return Err` patterns in here.

apiserver/src/webhook/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+        if !CONVERTERS.contains_key(source_version.as_str()) {
+            event!(
+                Level::DEBUG,
+                ?source_version,
+                "source version not in CONVERTERS, source_version:"
+            );
+            return Err(WebhookConvertError::SourceVersionNotExistInConverter { source_version });
+        }
+        source_object =
+            CONVERTERS[source_version.as_str()].1(source_object).context(ErrorDuringConvert)?;
```
cbgbt: nit: The tuple access syntax (`.0`, `.1`) makes this a bit hard to discern. It might help to turn that tuple into a struct with public named attributes.

apiserver/src/webhook/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+use snafu::{ResultExt, Snafu};
+use std::collections::{HashMap, HashSet};
+use tracing::{event, Level};
+
+type ConvertMethod = fn(serde_json::Value) -> models::node::Result<serde_json::Value>;
+pub type Result<T> = std::result::Result<T, WebhookConvertError>;
+
+lazy_static! {
+    static ref CONVERTERS: HashMap<String, (String, ConvertMethod)> = {
+        let mut converters = HashMap::new();
```
cbgbt: This might be easier to maintain if we use something like [`maplit!`](https://docs.rs/maplit/latest/maplit/)

apiserver/src/api/node.rs
```diff
@@ -42,6 +45,23 @@ pub(crate) async fn update_bottlerocket_shadow_resource<T: BottlerocketShadowCli
     Ok(HttpResponse::Ok().body(format!("{}", json!(&node_status))))
 }
 
+pub(crate) async fn convert_bottlerocket_shadow_resource(
```
cbgbt, line 48: We should probably add tests to this module which convert from V1 to V2 BRS via the API.

apiserver/src/webhook/mod.rs
cbgbt: Two ideas for testing this:
* Abstract the chained conversions into a `ChainConverter` struct which allows you to configure chains. Then test the converter using some arbitrary chains that you create. (More upfront cost now, but buys more confidence in the chained conversions going forward).
    * You can have a less "object-oriented" version of this that just makes the function take the converter chain as an argument.
* Add test cases for some explicit test chains that already exist. In this case, V1->V2, but this doesn't buy us much confidence that longer chains will work.

models/src/node/crd/mod.rs
somnusfish: It might be difficult. Right now the state change is 1 -> 1 map, but with the prevent controller from being unscheduled, the state from v1 to v2 would be 2 ->1, when we want to do the backward migration, it won't be clear which exact state it would be and can cause dead loop.

I was wondering when could the above situation happen in real life. theoretically the controller and agent would be deployed together, there might be very temporary slots that controller and agent in different version, but eventually controller and agent should be in the same version.

models/src/node/crd/mod.rs
somnusfish: Sorry I was confused what do you want me to remove?

models/src/node/crd/v2.rs
somnusfish: I used `BottlerocketShadow::from` at last, don't think there is much difference in using `from` and `into`, but let me know if there is any reason to choose one over the other.

apiserver/src/webhook/mod.rs
arnaldo2792: From offline discussion:

https://gist.github.com/arnaldo2792/6ba3bad8b3673402deb370ab28531aa4

In your current implementation, the shared "interface" for shadow objects is `serde_json::Value`. I think that it will be better if you have a struct that we control (like the one I called `Object`), that is just a wrapper around `serde_json::Value`, but that give us the flexibility to implement code within itself that knows how to translate this struct to the shadow objects that you will have, so the code can be more idiomatic.

The other structural change I made was removing the hashmap that keeps track of the converters, I haven't seen the usage of hasmaps for the use case that you have to "dynamically" call a function depending on a string, and instead what I've seen is the usage of `match` to specifically call a function. With the approach that I suggested it becomes more obvious what functions are called for each value that the `match` supports.

Lastly, I removed all the around the custom `Covert` bits, and I used Rust internals and `serde` to transform from one data structure to the other, so again, it is more obvious how to add more transformation, without having to add new converters and you just need to make sure you implement the correct traits for future shadow versions.

There are still improvements to what I did, for example, there could be a macro that automatically implements `TryFrom<ShadowObject> for Object` so that you don't repeat this implementation as I did:

```rust
impl TryFrom<ShadowObject> for Object {
    type Error = Error;
    fn try_from(shadow: ShadowObject) -> Result<Self, Self::Error> {
        Ok(Object {This is 
            object: serde_json::to_value(shadow).context(InvalidShadowObjectSnafu)?,
        })
    }
}
```

Since all the shadow objects that will be build from `Object` will have the same implementation.

apiserver/src/webhook/mod.rs
somnusfish: Will remove this part based on Arnaldo's suggestion.

apiserver/src/webhook/mod.rs
somnusfish: This part will be removed

models/src/node/crd/mod.rs
gthao313: Oh, I saw you commented out few lines. and I was wondering if you don't need them? 
like
```
/// static ref BOTTLEROCKETSHADOW_CRD_VERSIONS: Vec<String> = {
....
```

somnusfish: This is the doc for the file.

apiserver/src/api/node.rs
somnusfish: Added test

apiserver/src/webhook/mod.rs
somnusfish: Updated the code based on the comments

apiserver/src/webhook/mod.rs
somnusfish: This part get removed in the updated code.

models/src/node/crd/mod.rs
```diff
@@ -121,11 +148,77 @@ fn combine_crd(
...
+
+    let webhook_conversion = WebhookConversion {
+        client_config: Some(config),
+        conversion_review_versions: BOTTLEROCKETSHADOW_CRD_VERSIONS.to_vec(),
+    };
+
+    let crd_conversion = CustomResourceConversion {
+        strategy: "Webhook".to_string(),
+        webhook: Some(webhook_conversion),
+    };
```
gthao313: Nit (not a blocker): Might be better to construct them to single like we did to generate brupop resources yaml, Which looks like more concise.
```
let crd_conversion = CustomResourceConversion {
    strategy: "Webhook".to_string(),
    webhook: Some(
        WebhookConversion {
            client_config: Some(
                WebhookClientConfig {
                    service: Some(service),
                    ..Default::default()
                }
            ),
            conversion_review_versions: BOTTLEROCKETSHADOW_CRD_VERSIONS.to_vec(),
        }
    ),
};

```

apiserver/src/webhook/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+            _ => InvalidVersionError { version }.fail(),
+        }
+    }
+
+    #[instrument(skip(self), err)]
+    fn chained_convert_object(self, desired_version: String) -> Result<Self> {
+        let mut version = self.get_version()?;
+        let mut source_object = self;
+
+        while version != desired_version {
```
gthao313, line 192: If there is a situation that the `while` condition gets caught in an infinite loop because current version is unable to reach desired version but no error? 
If this situation exists, should we add a timeout and error out?

apiserver/src/webhook/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+        };
+
+        let converted_response = convert_request_to_response(&conversion_req);
+        assert_eq!(converted_response, expected_response);
+    }
+}
+#[derive(Debug, Snafu)]
+#[snafu(visibility = "pub")]
+pub enum WebhookConvertError {
+    #[snafu(display("Source version does not exist in ConversionRequest",))]
```
gthao313:
```suggestion
    #[snafu(display("Source version does not exist in ConversionRequest {}", source))]
```

apiserver/src/webhook/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+
+    #[snafu(display("Failed to convert BottlerocketShadow to json object {}", source))]
+    BottlerocketShadowToJsonConvertError {
+        version: String,
+        source: serde_json::error::Error,
+    },
+
+    #[snafu(display("Failed to convert json object to BottlerocketShadow {}", source))]
+    JsonToBottlerocketShadowConvertError {
+        version: String,
```
gthao313, line 404: Need version variable here?

apiserver/src/webhook/response.rs
```diff
@@ -0,0 +1,41 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct ConversionResponse {
+    pub kind: String,
+    pub api_version: String,
+    pub response: Response,
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+pub struct ConvertResult {
+    pub status: String,
```
gthao313: I think It's better to create an enum for `status` 
```
enum status {
    Success,
    Failed,
    ....
}
```

models/src/constants.rs
```diff
@@ -24,6 +24,8 @@ pub const CA_NAME: &str = "ca.crt";
 pub const PUBLIC_KEY_NAME: &str = "tls.crt";
 pub const PRIVATE_KEY_NAME: &str = "tls.key";
 pub const TLS_KEY_MOUNT_PATH: &str = "/etc/brupop-tls-keys";
+// Certificate object name
+pub const CERTIFICATE_NAME: &str = "brupop-apiserver-cerfiticate";
```
gthao313:
```suggestion
pub const CERTIFICATE_NAME: &str = "brupop-apiserver-certificate";
```

yamlgen/deploy/bottlerocket-update-operator.yaml
```diff
@@ -3,8 +3,22 @@
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: brupop-bottlerocket-aws/brupop-apiserver-cerfiticate
```
gthao313:
```suggestion
    cert-manager.io/inject-ca-from: brupop-bottlerocket-aws/brupop-apiserver-certificate
```

yamlgen/deploy/bottlerocket-update-operator.yaml
```diff
@@ -203,7 +217,7 @@ spec:
 apiVersion: cert-manager.io/v1
 kind: Certificate
 metadata:
-  name: brupop-apiserver
+  name: brupop-apiserver-cerfiticate
```
gthao313:
```suggestion
  name: brupop-apiserver-certificate
```

models/src/node/crd/v2.rs
```diff
@@ -73,6 +79,20 @@ impl BottlerocketShadowState {
     }
 }
 
+impl From<BottlerocketShadowState_v1> for BottlerocketShadowState {
+    fn from(previous_state: BottlerocketShadowState_v1) -> Self {
+        // TODO: Remap the state when merge PR with preventing controller from being unscheduled
```
gthao313, line 79: Yeah! Nice catch!

yamlgen/deploy/cert.yaml
```diff
@@ -18,7 +18,7 @@ spec:
 apiVersion: cert-manager.io/v1
 kind: Certificate
 metadata:
-  name: brupop-apiserver
+  name: brupop-apiserver-cerfiticate
```
gthao313:
```suggestion
  name: brupop-apiserver-certificate
```

apiserver/src/webhook/mod.rs
somnusfish: There won't be an infinite loop because it will reach the desired version or it will reach the place there is no version to convert hence error out.

models/src/node/crd/v2.rs
gthao313: I've merged controller update PR, so I think you can update here. : )

yamlgen/deploy/bottlerocket-update-operator.yaml
gthao313: Incorrect name still remain in Yaml file, maybe you need run `cargo build` to refresh the yaml file content. : )

somnusfish:
Addressed feedbacks and tested.

gthao313 (approved):
Looks good!

apiserver/src/api/node.rs
```diff
@@ -42,6 +43,14 @@ pub(crate) async fn update_bottlerocket_shadow_resource<T: BottlerocketShadowCli
...
+    http_req: HttpRequest,
+) -> Result<impl Responder> {
+    event!(Level::INFO, ?http_req, "Check the crd request.");
+    // TODO: implement the convert
+    Ok(HttpResponse::Ok())
```
arnaldo2792: Seems like this method is implemented in a different commit, so you can try to clean up this commit a little and have only one place where you change this method.

models/src/node/crd/mod.rs
```diff
@@ -26,19 +26,39 @@
 /// };
 /// ```
 ///
+/// Add the BottlerocketShadow version to the end of BOTTLEROCKETSHADOW_CRD_VERSIONS
+/// So the webhook conversion could set up proper conversion_review_versions
```
arnaldo2792: Nit: should be a lowercase 'S' at the beginning of this sentence.

models/src/node/crd/mod.rs
```diff
@@ -121,11 +148,70 @@ fn combine_crd(
     latest_crd
 }
 
+/// Generate webhook conversion from scratch since k8s_api didn't provide
+/// a decent way to set up CustomResourceConversion
```
arnaldo2792, line 140: Nit: kind of harsh on the k8s project :sweat_smile:, maybe saying "a way that works for us" could be better?

models/src/node/crd/mod.rs
```diff
@@ -121,11 +148,70 @@ fn combine_crd(
...
+/// Generate cert-manager annotations to help inject caBundle for webhook
+/// https://cert-manager.io/docs/concepts/ca-injector/#injecting-ca-data-from-a-certificate-resource
+fn generate_ca_annotations() -> BTreeMap<String, String> {
+    let mut cert_manager_annotations = BTreeMap::new();
+    cert_manager_annotations.insert(
+        "cert-manager.io/inject-ca-from".to_string(),
+        format!(
+            "{namespace}/{object}",
+            namespace = NAMESPACE,
+            object = CERTIFICATE_NAME
```
arnaldo2792, line 182: Do the string parameters for the interpolation need to be named? I haven't seen this in other Bottlerocket projects, so I'm wondering if we need to start asking for this?

models/src/node/crd/mod.rs
```diff
@@ -121,11 +148,70 @@ fn combine_crd(
...
+
 pub fn combined_crds() -> CustomResourceDefinition {
     let mut crds: Vec<CustomResourceDefinition> = BOTTLEROCKETSHADOW_CRD_METHODS
         .iter()
         .map(|crd_method| crd_method())
         .collect();
     let latest_crd = crds.pop().unwrap();
-    combine_crd(latest_crd, crds)
+    let combined_version_crds = combine_version_in_crds(latest_crd, crds);
+    add_webhook_setting(combined_version_crds)
```
arnaldo2792, line 204: Pedantic nit: Unless there are many places where `add_webkhook_settings` method is used or you envision there will be, I think it should be ok to do what the method does here (modify `combined_version_crds` and return an owned reference to it).

apiserver/src/webhook/request.rs
```diff
@@ -0,0 +1,28 @@
...
+    #[serde(rename = "desiredAPIVersion")]
+    pub desired_api_version: String,
+    pub objects: Vec<serde_json::Value>,
+}
+
+impl ConversionRequest {
+    /// Wrap over convert_request_to_response so
+    /// actix_web::web::Json<ConversionRequest> could call the method
+    pub fn convert_resource(&self) -> ConversionResponse {
+        convert_request_to_response(self)
```
arnaldo2792, line 26: This method is defined in a different commit, yet it is used here. Please make sure all your commits can be built individually.

models/src/node/crd/v2.rs
```diff
@@ -1,13 +1,19 @@
+use super::v1::BottlerocketShadow as BottleRocketShadow_v1;
```
arnaldo2792: Why `BottlerocketShadow_v1` and not `BottlerocketShadowV1`?

models/src/node/crd/v2.rs
```diff
@@ -181,6 +201,15 @@ impl BottlerocketShadowSpec {
     }
 }
 
+impl From<BottlerocketShadowSpec_v1> for BottlerocketShadowSpec {
+    fn from(previous_spec: BottlerocketShadowSpec_v1) -> Self {
+        Self::new(
+            BottlerocketShadowState::from(previous_spec.state),
+            previous_spec.state_timestamp().unwrap(),
```
arnaldo2792, line 208: Please avoid using `unwrap` since it could panic the stack

models/src/node/crd/v2.rs
```diff
@@ -242,3 +271,201 @@ impl BottlerocketShadowStatus {
...
+                uid: previous_metadata.uid,
+                owner_references: previous_metadata.owner_references,
+                ..Default::default()
+            },
+            spec,
+            status,
+        };
+        new_shadow
+    }
+}
```
arnaldo2792, line 313: Nit: new line missing.

models/src/node/crd/v2.rs
```diff
@@ -242,3 +271,201 @@ impl BottlerocketShadowStatus {
...
+                    "state_transition_timestamp": "2022-07-09T19:32:38.609610964+00:00",
+                    "version": "1.8.0"
+                }),
+            ),
+        ];
+
+        for (original, target) in original_target_spec.into_iter() {
+            let old_spec: BottlerocketShadowSpec_v1 = serde_json::from_value(original).unwrap();
+            let new_spec = BottlerocketShadowSpec::from(old_spec);
+            assert_eq!(serde_json::to_value(new_spec).unwrap(), target);
```
arnaldo2792, line 376: It will be nice probably in the future to implement the `Eq` traits for the shadow objects, only if you know of a use case that could benefit from this.

apiserver/src/api/node.rs
```diff
@@ -44,21 +46,30 @@ pub(crate) async fn update_bottlerocket_shadow_resource<T: BottlerocketShadowCli
...
 ) -> Result<impl Responder> {
-    event!(Level::INFO, ?http_req, "Check the crd request.");
-    // TODO: implement the convert
-    Ok(HttpResponse::Ok())
+    event!(Level::INFO, ?conversion_req, "Original conversion request");
+    let response = conversion_req.convert_resource();
+    let response_string = serde_json::to_string(&response).context(error::WebhookError)?;
+    event!(Level::INFO, ?response_string, "Converted response:");
+
+    // Webhook will always response with 200.
```
arnaldo2792: Nit: typo

```suggestion
    // Webhook will always respond with 200.
```

apiserver/src/api/node.rs
```diff
@@ -44,21 +46,30 @@ pub(crate) async fn update_bottlerocket_shadow_resource<T: BottlerocketShadowCli
...
-    http_req: HttpRequest,
+    conversion_req: web::Json<ConversionRequest>,
 ) -> Result<impl Responder> {
-    event!(Level::INFO, ?http_req, "Check the crd request.");
-    // TODO: implement the convert
-    Ok(HttpResponse::Ok())
+    event!(Level::INFO, ?conversion_req, "Original conversion request");
+    let response = conversion_req.convert_resource();
+    let response_string = serde_json::to_string(&response).context(error::WebhookError)?;
+    event!(Level::INFO, ?response_string, "Converted response:");
```
arnaldo2792, line 54: Will any of the converted objects include sensitive information that shouldn't be logged out?

apiserver/src/api/node.rs
```diff
@@ -191,4 +202,64 @@ mod tests {
...
+        .await;
+
+        let resp = test::call_service(&app, req).await;
+
+        assert!(resp.status().is_success());
+        if let AnyBody::Bytes(b) = resp.into_body() {
+            // Only check the response body can be converted to ConversionResponse.
+            // Contents of the ConversionResponse should be tested in convert_resource method.
+            let _return_response: ConversionResponse =
+                serde_json::from_slice(&b).expect("Could not parse JSON response.");
```
arnaldo2792: If you are not going to use the object, I think you can use `::from_slice<Type>()`

apiserver/src/webhook/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+use models::node::v1::BottlerocketShadow as BottleRocketShadow_v1;
+use models::node::v2::BottlerocketShadow as BottlerocketShadow_v2;
+
+use snafu::{ResultExt, Snafu};
+use std::convert::TryFrom;
+use tracing::instrument;
+
+pub type Result<T> = std::result::Result<T, WebhookConvertError>;
+
+pub fn convert_request_to_response(req: &ConversionRequest) -> ConversionResponse {
```
arnaldo2792, line 98: Maybe a comment here to explain others how this method works?

apiserver/src/webhook/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+    Ok(new_objects)
+}
+
+struct BRSObject {
+    pub object: serde_json::Value,
+}
+
+impl BRSObject {
+    fn get_version(&self) -> Result<String> {
+        Ok(serde_json::from_value(self.object["apiVersion"].clone())
```
arnaldo2792, line 169: Maybe `version` in `BRSObject` should be set once `get_version` is called to prevent double deserialization.

apiserver/src/webhook/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+                        src_version: version,
+                        dst_version: desired_version,
+                    }
+                    .fail()
+                }
+            }
+            version = source_object.get_version()?;
+        }
+
+        Ok(source_object)
```
arnaldo2792, line 206:
```suggestion
        let mut source_object = self;
        while version.get_version()? != desired_version {
            // ChainedConverError should just add more context to the error returned by "convert_to_next_version"
            source_object = source_object.convert_to_next_version().context(ChainedConvertErrorSnafu)?;
        }

        Ok(source_object)

```

apiserver/src/webhook/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+    let mut new_objects = Vec::new();
+    for old_object in objects.into_iter() {
+        let old_brs_object = BRSObject { object: old_object };
+        let new_brs_object = old_brs_object.chained_convert_object(desired_version.clone())?;
+        new_objects.push(new_brs_object.object);
+    }
+    Ok(new_objects)
+}
+
+struct BRSObject {
```
arnaldo2792, line 163: A comment in this struct will help others understand why was it needed

arnaldo2792 (changes requested):

apiserver/src/api/node.rs
somnusfish: Not sure what do you mean by clean but if I remove this method in this commit then this commit won't be as a whole

models/src/node/crd/mod.rs
somnusfish: Not necessary. I saw both named and unnamed use case. I just feel name it would make the map more clear.

apiserver/src/webhook/request.rs
somnusfish: I will combine the two commit into one since it is very hard to separate them now.

models/src/node/crd/v2.rs
somnusfish: I think according to https://rust-lang.github.io/api-guidelines/naming.html it might be better to use BottlerocketShadowV1, I will update it.

apiserver/src/webhook/mod.rs
somnusfish: I don't think there is much need to store version at this point because BRSObject will be created on every request, and during the convert, each object would only be used once.

apiserver/src/webhook/mod.rs
somnusfish: I don't think I am apply what you suggested here. 

First of all, `convert_to_next_version` returns error `InvalidVersionError` which indicate that the next version doesn't exist in the map, but it didn't have any context on the target version. So I need `ChainedConvertError` to give more information on during the chained convert, which version it is in and what is the target version.

I am not using context here because both `InvalidVersionError` and `ChainedConvertError` are defined in `WebhookConvertError` and I really couldn't find a way to put `InvalidVersionError` as source for `ChainedConvertError`. One way could be I put `InvalidVersionError` to another error type, but I don't know whether it would be worthy to do that.

apiserver/src/api/node.rs
somnusfish: There won't be any sensitive information on the request.

models/src/node/crd/mod.rs
somnusfish: As per discussion offline, I would like to keep the method be simple.

models/src/node/crd/v2.rs
somnusfish: I am using this because this allows me to create the object from json that I can define a list and compare the target and source inside.  To answer why am I not directly put the object in the list, this is because some of the entry for the struct is private and doesn't allow customized setting value.

models/src/node/crd/v2.rs
somnusfish: `state_transition_timestamp` is a private variable that is set during the creation, there is not other way to set it. Logically `state_timestamp` won't emit any error thus it is safe to use unwrap here.

arnaldo2792 (approved):

Issue #205, opened 2022-06-03
Field from CRD manifest gets removed after applying the manifest


michalschott:
operator version 0.2.1 / develop

```
➜  k version --short
Client Version: v1.22.9
Server Version: v1.22.6-eks-14c7a48
```

[This](https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/develop/yamlgen/deploy/bottlerocket-update-operator.yaml#L10) extra field in CRD definition gets removed after applying the manifest, thus tools like ArgoCD/FluxCD are constantly running tasks to update the in-cluster crd manifest.

somnusfish:
Hi @michalschott,

Thanks for reaching out. Would you mind elaborate more on the issue you just reported?

> operator version 0.2.1 / develop

The version that you mentioned here are actually two version, 0.2.1 and develop. which version did you use and what manifest did you apply? Note that the `bottlerocket-update-operator.yaml` for 0.2.1 is different than the one in develop.

And what is the error that you are facing?

Thanks

michalschott:
Apologies to not being clear enough at the beginning.

I've used [this manifest](https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/v0.2.1/yamlgen/deploy/bottlerocket-update-operator.yaml#L10), but also [this one](https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/develop/yamlgen/deploy/bottlerocket-update-operator.yaml#L10). Both of them have field `.spec.names.categories[]` defined which is being removed by kubernetes (in my case EKS 1.22) once manifest is applied. This results in a gap / difference between defined and in-live manifest, which causes tools like ArgoCD / FluxCD to constantly reapply defined manifests.

Solution would be to drop mentioned field from auto-generated manifest.

somnusfish:
The manifest get updated in our latest release 0.2.2, closing this issue for now and please feel free to reopen/create a new issue if you have further concerns.

PR #227, opened 2022-07-25
Remove empty categories in Custom Resouce spec

somnusfish:
**Issue number:**
#205 

**Description of changes:**
Remove empty categories list in CRD.

Reason:
`#[derive(CustomResource)]` by default set categories to [empty list](https://github.com/kube-rs/kube-rs/blob/master/kube-derive/src/custom_resource.rs#L25), which doesn't match the object in Kubernetes.
```
  Group:  brupop.bottlerocket.aws
  Names:
    Kind:       BottlerocketShadow
    List Kind:  BottlerocketShadowList
    Plural:     bottlerocketshadows
    Short Names:
      brs
    Singular:  bottlerocketshadow
  Scope:       Namespaced
```
Causing  ArgoCD / FluxCD  detect differences and reapply the manifest

**Testing done:**
cargo build

**Terms of contribution:**

### DIFF
--- a/models/src/node/crd/mod.rs
+++ b/models/src/node/crd/mod.rs
@@ -194,6 +194,14 @@ fn add_webhook_setting(
     combined_version_crds
 }
 
+/// `#[derive(CustomResource)]` set default categories to empty list
+/// causes mismatch in Kubernetes's object and YAML manifest file,
+/// futher causes ArgoCD/FluxCD constantly reapply defined manifest.
+fn remove_empty_categories(mut crds: CustomResourceDefinition) -> CustomResourceDefinition {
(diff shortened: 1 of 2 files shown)


gthao313:
The code looks like correct to me! Just curious if we should test the changes with  ArgoCD / FluxCD to confirm this solve the issue?

somnusfish:
> The code looks like correct to me! Just curious if we should test the changes with ArgoCD / FluxCD to confirm this solve the issue?

I don't think it's super necessary. I compared the manifest with the Kubernetes object, it's exactly the same now.

gthao313 (approved):

etungsten (approved):

Issue #156, opened 2022-02-14
Does this project deregister nodes from load balancers?


tjs-intel:
**What I'd like:**
I'm considering adding this to my EKS deployment, but I found a flaw in another project that drains nodes before termination (aws-node-termination-handler) where nodes weren't properly detached from load balancers before being terminated which resulted in failed requests. Does this project cover LB deregistration? I cant find the label `node.kubernetes.io/exclude-from-external-load-balancers` anywhere in this project.

**Any alternatives you've considered:**
I use a Cloudformation Stack for my ASGs, with a daemonset in the cluster that signals the stack. I might be able to get away with updating the Bottlerocket AMI every week and just let the aws-node-termination-handler deregister nodes (a recently added feature).

sarahhodne:
I'm curious about this too, I'm investigating to see if Bottlerocket would be a good fit for us, and load balancer deregistration is a recurring pain point in our cluster at the moment, we've found that even the `node.kubernetes.io/exclude-from-external-load-balancers` label isn't sufficient, since the pods drain from our nodes much faster than the instance can deregister from the load balancers (usually a minimum of 3 minutes, sometimes longer depending on configuration).

See also https://github.com/aws/aws-node-termination-handler/issues/316

cbgbt:
Thanks for raising this. You're right, we don't currently handle this case. But we should! We'll investigate how to best add this to the update operator.

For external load balancers with slower reaction times mentioned by @sarahhodne, I wonder if it would be appropriate for us to allow users to configure a check between exclusion from load balancers and draining to assert that the operator can proceed?

sarahhodne:
That should work in our case, I think.

If it's easier, even just having a sleep or some flag that ensures the drain takes at least x minutes should be sufficient. We can adjust how long the load balancers take to deregister with configuration, the problem is that the shortest duration we've been able to configure is 3 minutes, and since our pods can often terminate much faster than that, we need something that keeps the node online for a little while longer while the deregistration completes.

PR #231, opened 2022-08-06
Add option to deregister node before draining.

somnusfish:
**Issue number:**
#156 

**Description of changes:**
* Add `DEREGISTER_FROM_LB_WAIT_TIME_IN_SEC` in `.env` as environment variable.
* Deregister node from load balancer before draining
  when DEREGISTER_FROM_LB_WAIT_TIME_IN_SEC is larger than 0.
* Register node to load balancer after draining
  when DEREGISTER_FROM_LB_WAIT_TIME_IN_SEC is larger than 0.

**Testing done:**
Tested on cluster, monitored node get labeled and unlabeled.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,32 @@ This will create the required namespace, custom resource definition, roles, depl
 Note: The above link is set to use the configuration for the latest version of the Update Operator, `v0.2.1`.
 Be sure to use the git tag for the Update Operator release that you plan to deploy.
 
+#### Configuration
+
+##### Exclude Nodes from Load Balancers Before Draining
+This configuration uses Kuberenetes [ServiceNodeExclusion](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) feature.
+`EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC` can be used to enable the feature to exclude the node from load balancer before draining.
+When `EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC` is 0 (default), the feature is disabled.
+When `EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC` is set to a positive integer, bottlerocket update operater will exclude the node from 
+load balancer and then wait `EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC` seconds before draining the pods on the node.
+
+To enable this feature, go to `bottlerocket-update-operator.yaml`, change `EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC` to a positive integer value.
+For example:
+```yaml
+      ...
+      containers:
+        - command:
+            - "./agent"
+          env:
+            - name: MY_NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+            - name: EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC
+              value: "180"
+      ...
+```
+
 ### Label nodes
 
 By default, each Workload resource constrains scheduling of the update operator by limiting pods to Bottlerocket nodes based on their labels.
--- a/agent/src/agentclient.rs
+++ b/agent/src/agentclient.rs
@@ -1,7 +1,8 @@
 use crate::apiclient::{boot_update, get_chosen_update, get_os_info, prepare, update};
 use apiserver::{
     client::APIServerClient,
-    CordonAndDrainBottlerocketShadowRequest, UncordonBottlerocketShadowRequest,
+    CordonAndDrainBottlerocketShadowRequest, ExcludeNodeFromLoadBalancerRequest,
+    RemoveNodeExclusionFromLoadBalancerRequest, UncordonBottlerocketShadowRequest,
     {CreateBottlerocketShadowRequest, UpdateBottlerocketShadowRequest},
 };
 use models::{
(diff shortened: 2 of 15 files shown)


models/src/node/client.rs
```diff
@@ -47,6 +47,12 @@ pub trait BottlerocketShadowClient: Clone + Sized + Send + Sync {
     async fn drain_node(&self, selector: &BottlerocketShadowSelector) -> Result<()>;
     // Marks the given node as scheduleable, allowing Pods to be deployed onto it.
     async fn uncordon_node(&self, selector: &BottlerocketShadowSelector) -> Result<()>;
+    // Label the node with "node.kubernetes.io/exclude-from-external-load-balancers=True" to
+    // deregister the node from load balancer.
+    async fn deregister_node_from_lb(&self, selector: &BottlerocketShadowSelector) -> Result<()>;
+    // Remove "node.kubernetes.io/exclude-from-external-load-balancers" label from the node
+    // so that load balancer could register the node
+    async fn register_node_to_lb(&self, selector: &BottlerocketShadowSelector) -> Result<()>;
```
cbgbt: I feel like `register` implies more than what the actual behavior is, since what this function does is remove or undo an exclusion.

Should we just use the same terminology as the Kubernetes label? Maybe something like `exclude_node_from_lb` and `remove_node_exclusion_from_lb`?

agent/src/agentclient.rs
```diff
@@ -265,6 +267,44 @@ impl<T: APIServerClient> BrupopAgent<T> {
         Ok(())
     }
 
+    fn get_deregister_lb_time(&self) -> Result<u64> {
+        let wait_time: u64 = env::var("DEREGISTER_FROM_LB_WAIT_TIME_IN_SEC")
```
cbgbt: We should document this setting in the `README`.

agent/src/agentclient.rs
```diff
@@ -285,6 +325,7 @@ impl<T: APIServerClient> BrupopAgent<T> {
         if !shadow_exists {
             self.create_metadata_shadow().await?;
         }
+        self.register_node_to_lb().await?;
```
cbgbt: We may want to consider that Brupop users could be modifying this label outside of Brupop itself. I'm not totally convinced we should remove the exclusion label at the start, perhaps only if we know we have already added the label. What do you think?

somnusfish: This was my mistake, this is the part didn't get clean properly when I was testing the functionality initially. Yes, we need to remove this part out.

models/src/node/client.rs
somnusfish: I like the new name.

somnusfish:
Address the comments with name change and updated readme.md.

README.md
```diff
@@ -27,6 +27,17 @@ This will create the required namespace, custom resource definition, roles, depl
 Note: The above link is set to use the configuration for the latest version of the Update Operator, `v0.2.1`.
 Be sure to use the git tag for the Update Operator release that you plan to deploy.
 
+#### Features
```
cbgbt: nit: I might call this `Configuration` or `Additional Setup`

README.md
```diff
@@ -27,6 +27,17 @@ This will create the required namespace, custom resource definition, roles, depl
...
+
+##### exclude from load balancer before draining
+`EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC` can be used to enable the feature to exclude the node from load balancer before draining.
```
cbgbt, line 34: The Kubernetes documentation refers to this feature as `ServiceNodeExclusion`, although I can only really find reference to it in the Feature Gate section of the docs: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

We may want to refer to it by name here.

README.md
```diff
@@ -27,6 +27,17 @@ This will create the required namespace, custom resource definition, roles, depl
 Note: The above link is set to use the configuration for the latest version of the Update Operator, `v0.2.1`.
 Be sure to use the git tag for the Update Operator release that you plan to deploy.
 
+#### Features
+
+##### exclude from load balancer before draining
```
cbgbt: nit: capitalize this
```suggestion
##### Exclude Nodes from Load Balancers Before Draining
```

README.md
```diff
@@ -27,6 +27,17 @@ This will create the required namespace, custom resource definition, roles, depl
...
+`EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC` can be used to enable the feature to exclude the node from load balancer before draining.
+When `EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC` is 0 (default), the feature is disabled.
+When `EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC` is set to a positive integer, bottlerocket update operater will exclude the node from 
+load balancer, wait `EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC` seconds before draining the pods on the node.
```
cbgbt: nit
```suggestion
load balancer and then wait `EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC` seconds before draining the pods on the node.
```

README.md
```diff
@@ -27,6 +27,17 @@ This will create the required namespace, custom resource definition, roles, depl
...
+
+To enable this feature, go to `.env`, change `EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC` to a positive integer value,
+run `cargo build` to regenerate `bottlerocket-update-operator.yaml` and follow the normal installation guideline.
```
cbgbt: I think instead of directing folks to rebuild the project, we should have folks modify the `.yaml` document describing the pod deployment to modify the environment variable.

When we begin delivering this as a Helm chart, we can make that variable configurable via Helm.

somnusfish: Updated readme using .yaml for user. I would like to have a separate PR to update the development part.

cbgbt (approved):

gthao313 (approved):

Issue #107, opened 2021-11-02
0.2.0: Cleanup BRS when the operator is removed from a node


cbgbt:
**Issue or Feature Request:**
BRSs currently have `ownerReferences` to the k8s `Node` object that the BRS is associated with, meaning that if the Node is deleted, the BRS will be deleted with it. This is great, because it allows the controller to remove the BRS from its ActiveSet.

I foresee an issue though: If a customer removes the `brupop` label from their node in the middle of an update, the brupop `agent`/daemonset will be removed from the host, causing updates to cease; however, the BRS in question will not be deleted until the corresponding Node is deleted.

We should create a strategy by which BRSs are cleaned up when the daemonset is removed, or allow the controller to delete BRSs which timeout and seem to become endlessly stuck.

cbgbt:
I've added some text about this case to the README for now, so there is at least documentation for assisting customers to not become stuck by this. However, having an automated solution here would be idea.

cbgbt:
If the controller keeps a `reflector` of Nodes, it won't have to do an additional API call to check for label existance.

PR #235, opened 2022-08-11
cleanup brs when brupop is removed from the node.

gthao313:
**Issue number:**
#107 

**Description of changes:**
```
Date:   Thu Aug 11 19:32:12 2022 +0000

    cleanup brs when the brupop is removed from the node

    Currently we require customers to manually clean up the Custom Resources
    if they choose to delete only the updater-interface-version label. This
    change would help on automatically cleaning up BRS if customer removes
    brupop(label) from corresponding node.

commit b81eefc195f4669510146793acf64b3c5bc959e9
Date:   Thu Aug 11 18:36:49 2022 +0000

    Add deletion support to brs client

    According to brs cleanup strategy, we support brs client to be able to
    delete brs object.
```

**Testing done:**
Method: Run brupop integration test. Remove brupop label from nodes, and then confirm if associated brs has been removed. 

Test with three scenarios 

- [x] [Waiting-for-update] The node is waiting for update and customers remove the labels 
- [x] [In-progress] Customers remove the labels when brupop is preforming update on the node.
- [x] [Complete] The node has been updated, and customers remove the labels.

**Terms of contribution:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -840,6 +840,7 @@ dependencies = [
  "http",
  "k8s-openapi",
  "kube",
+ "maplit",
  "models",
  "opentelemetry",
  "opentelemetry-prometheus",
--- a/controller/src/controller.rs
+++ b/controller/src/controller.rs
@@ -2,12 +2,16 @@ use super::{
     metrics::{BrupopControllerMetrics, BrupopHostsData},
     statemachine::determine_next_node_spec,
 };
+use models::constants::{BRUPOP_INTERFACE_VERSION, LABEL_BRUPOP_INTERFACE_NAME, NAMESPACE};
 use models::node::{
     brs_name_from_node_name, BottlerocketShadow, BottlerocketShadowClient, BottlerocketShadowState,
     Selector,
 };
 
+use k8s_openapi::api::core::v1::Node;
+use kube::api::DeleteParams;
 use kube::runtime::reflector::Store;
+use kube::Api;
 use kube::ResourceExt;
 use opentelemetry::global;
 use snafu::ResultExt;
@@ -30,27 +34,45 @@ type Result<T> = std::result::Result<T, controllerclient_error::Error>;
 
 /// The BrupopController orchestrates updates across a cluster of Bottlerocket nodes.
 pub struct BrupopController<T: BottlerocketShadowClient> {
+    k8s_client: kube::client::Client,
     node_client: T,
     brs_reader: Store<BottlerocketShadow>,
+    node_reader: Store<Node>,
     metrics: BrupopControllerMetrics,
 }
 
 impl<T: BottlerocketShadowClient> BrupopController<T> {
-    pub fn new(node_client: T, brs_reader: Store<BottlerocketShadow>) -> Self {
+    pub fn new(
+        k8s_client: kube::client::Client,
+        node_client: T,
+        brs_reader: Store<BottlerocketShadow>,
+        node_reader: Store<Node>,
+    ) -> Self {
         // Creates brupop-controller meter via the configured
         // GlobalMeterProvider which is setup in PrometheusExporter
         let meter = global::meter("brupop-controller");
         let metrics = BrupopControllerMetrics::new(meter);
         BrupopController {
+            k8s_client,
             node_client,
             brs_reader,
+            node_reader,
             metrics,
         }
     }
 
-    /// Returns a list of all `BottlerocketShadow` objects in the cluster.
-    fn all_nodes(&self) -> Vec<BottlerocketShadow> {
+    /// Returns a list of all custom definition resource `BottlerocketShadow`/`brs` objects in the cluster.
+    fn all_brss(&self) -> Vec<BottlerocketShadow> {
         self.brs_reader
+            .state()
+            .iter()
+            .map(|arc_brs| (**arc_brs).clone())
+            .collect()
+    }
+
+    /// Returns a list of all bottlerocket nodes in the cluster.
+    fn all_nodes(&self) -> Vec<Node> {
+        self.node_reader
             .state()
             .iter()
             .map(|arc_brs| (**arc_brs).clone())
@@ -62,8 +84,8 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
     /// Nodes are being acted upon if they are not in the `WaitingForUpdate` state, or if their desired state does
     /// not match their current state.
     #[instrument(skip(self))]
-    fn active_node_set(&self) -> BTreeMap<String, BottlerocketShadow> {
-        self.all_nodes()
+    fn active_brs_set(&self) -> BTreeMap<String, BottlerocketShadow> {
+        self.all_brss()
             .iter()
             .filter(|brs| {
                 brs.status.as_ref().map_or(false, |brs_status| {
@@ -117,8 +139,8 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
     /// The state transition is then attempted. If successful, this node should be detected as part of the active
     /// set during the next iteration of the controller's event loop.
     #[instrument(skip(self))]
-    async fn find_and_update_ready_node(&self) -> Result<Option<BottlerocketShadow>> {
-        let mut shadows: Vec<BottlerocketShadow> = self.all_nodes();
+    async fn find_and_update_ready_brs(&self) -> Result<Option<BottlerocketShadow>> {
+        let mut shadows: Vec<BottlerocketShadow> = self.all_brss();
 
         sort_shadows(&mut shadows, &get_associated_bottlerocketshadow_name()?);
         for brs in shadows.drain(..) {
@@ -153,7 +175,7 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
         let mut hosts_version_count_map = HashMap::new();
         let mut hosts_state_count_map = HashMap::new();
 
-        for brs in self.all_nodes() {
+        for brs in self.all_brss() {
             if let Some(brs_status) = brs.status {
                 let current_version = brs_status.current_version().to_string();
                 let current_state = brs_status.current_state;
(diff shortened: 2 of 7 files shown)


gthao313:
Push above removing changes on apisever. Since controller has privilege to delete cluster resources, we don't need add deletion to apiserver.

controller/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-mod controller;
+pub mod controller;
```
somnusfish: I don't find `pub` is necessary here, any reason this has to be `pub`?

gthao313: Since we make create multiple error classifications, I have an error module `controllerclient_error` under controller.rs. Then my new changes on `mian.rs` need use `controllerclient_error` error type as resource, so I have to make it pub.

```
pub mod controller_error {
    use controller::controller::controllerclient_error;
    use snafu::Snafu;
```

somnusfish: Can we do the following in `lib.rs`
```
pub use crate::controller::BrupopController;
```
And do in `main.rs`?
```
use controller::controllerclient_error;
```

controller/src/main.rs
```diff
@@ -70,16 +82,23 @@ async fn main() -> Result<()> {
...
     tokio::select! {
-        _ = drainer => {
-            event!(Level::ERROR, "reflector drained");
+        _ = brs_drainer => {
+            event!(Level::ERROR, "brs reflector drained");
+            return Err(controller_error::Error::KubernetesWatcherFailed {object: "brs".to_string()});
+        },
+        _ = node_drainer => {
+            event!(Level::ERROR, "node reflector drained");
+            return Err(controller_error::Error::KubernetesWatcherFailed {object: "node".to_string()});
```
somnusfish: Any reason returning Err here? The Result<()> in main is actually not parsed by any other handler. If this is for log purpose, does it make sense to log in the `event!` rather than return Err?

controller/src/controller.rs
```diff
@@ -2,12 +2,16 @@ use super::{
     metrics::{BrupopControllerMetrics, BrupopHostsData},
     statemachine::determine_next_node_spec,
 };
+use k8s_openapi::api::core::v1::Node;
```
somnusfish: nit: Would you mind sort the external import in alphabetical order?

controller/src/controller.rs
```diff
@@ -24,40 +28,58 @@ type Result<T> = std::result::Result<T, controllerclient_error::Error>;
...
+            .map(|arc_brs| (**arc_brs).clone())
+            .collect()
+    }
+
     /// Returns the set of BottlerocketShadow objects which is currently being acted upon.
     ///
     /// Nodes are being acted upon if they are not in the `WaitingForUpdate` state, or if their desired state does
     /// not match their current state.
     #[instrument(skip(self))]
     fn active_node_set(&self) -> BTreeMap<String, BottlerocketShadow> {
```
somnusfish: Maybe `active_brs_set` would be more clear?

controller/src/controller.rs
```diff
@@ -111,7 +133,7 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
     /// set during the next iteration of the controller's event loop.
     #[instrument(skip(self))]
     async fn find_and_update_ready_node(&self) -> Result<Option<BottlerocketShadow>> {
```
somnusfish: Same here, may be `find_and_update_ready_brs`?

controller/src/controller.rs
```diff
@@ -205,6 +264,11 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
                 }
             }
 
+            // Cleanup BRS when the operator is removed from a node
```
somnusfish, line 275: I was wondering if it would be better to add more explanation here to illustrate that even the above steps moved a invalid Brs to active, the clean up would catch later here.

controller/src/controller.rs
```diff
@@ -261,10 +325,32 @@ fn sort_shadows(shadows: &mut Vec<BottlerocketShadow>, associated_brs_name: &str
...
+        if !check_brupop_label(node.clone()) {
+            unlabeled_nodes.push(node);
+        }
+    }
+
+    unlabeled_nodes
+}
+
+#[instrument(skip(node))]
+fn check_brupop_label(node: Node) -> bool {
```
somnusfish: Does it make sense to use reference here?
```
fn check_brupop_label(node: &Node) -> bool 

if !check_brupop_label(&node) {
```

controller/src/controller.rs
```diff
@@ -168,6 +190,43 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
...
+    async fn bottlerocketshadows_cleanup(
+        &self,
+        nodes: Vec<Node>,
+        brss: Vec<BottlerocketShadow>,
+    ) -> Result<()> {
+        let mut unlabeled_nodes = find_unlabeled_nodes(nodes);
+
+        for unlabeled_node in unlabeled_nodes.drain(..) {
+            let associated_bottlerocketshadow = brs_name_from_node_name(&unlabeled_node.name());
+            if brss
```
somnusfish: I don't see any use for `BottlerocketShadow` when deleting the shadow. Does it make sense to get the BottlerocketShadow's name in `HashSet` to avoid iterating on all of the `brss` for each unlabeld_node?

controller/src/controller.rs
```diff
@@ -168,6 +190,43 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
...
 
+    #[instrument(skip(self, nodes, brss))]
+    async fn bottlerocketshadows_cleanup(
+        &self,
+        nodes: Vec<Node>,
+        brss: Vec<BottlerocketShadow>,
+    ) -> Result<()> {
+        let mut unlabeled_nodes = find_unlabeled_nodes(nodes);
+
+        for unlabeled_node in unlabeled_nodes.drain(..) {
```
somnusfish: I am not sure if I understand correctly, the `drain()` actually take the ownership for the `unlabeld_nodes`, but in this use case we only need the reference for the `unlabled_node.name()`, does it make sense to use reference for `unlabeled_nodes`? Or is there any reason to use `drain` here?

controller/src/controller.rs
```diff
@@ -24,40 +28,58 @@ type Result<T> = std::result::Result<T, controllerclient_error::Error>;
...
             node_client,
             brs_reader,
+            node_reader,
             metrics,
         }
     }
 
     /// Returns a list of all `BottlerocketShadow` objects in the cluster.
-    fn all_nodes(&self) -> Vec<BottlerocketShadow> {
+    fn all_brss(&self) -> Vec<BottlerocketShadow> {
```
arnaldo2792, line 65: This is getting confusing, could you please add a more descriptive name for the method?

Cargo.lock
```diff
@@ -840,6 +840,7 @@ dependencies = [
  "http",
  "k8s-openapi",
  "kube",
+ "maplit",
```
arnaldo2792, line 843: For the lack of a better place for this comment: In the commit message replace any `customer` occurrence with `user`.

controller/src/controller.rs
```diff
@@ -261,10 +325,32 @@ fn sort_shadows(shadows: &mut Vec<BottlerocketShadow>, associated_brs_name: &str
...
+        if !check_brupop_label(&node.clone()) {
+            unlabeled_nodes.push(node.name());
+        }
+    }
+
+    unlabeled_nodes
+}
+
+#[instrument(skip(node))]
+fn check_brupop_label(node: &Node) -> bool {
```
arnaldo2792: A more descriptive name for the function should be better :+1:, something like ` node_has_label` or `is_node_labeled_with_<your_label_name_here>`

controller/src/controller.rs
```diff
@@ -418,5 +532,8 @@ pub mod controllerclient_error {
 
         #[snafu(display("Could not determine selector for node: '{}'", source))]
         NodeSelectorCreation { source: BottlerocketShadowError },
+
+        #[snafu(display("Unable to get Node uid because of missing Node `uid` value"))]
```
arnaldo2792: Nit: `s/uid/UID`

controller/src/controller.rs
```diff
@@ -168,6 +190,43 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
...
+                bottlerocket_shadows
+                    .delete(
+                        associated_bottlerocketshadow.as_str(),
+                        &DeleteParams::default(),
+                    )
+                    .await
+                    .context(controllerclient_error::DeleteNode)?
+                    .map_right(|_| None)
+                    .map_left(Some)
+                    .into_inner();
```
arnaldo2792: For my own understanding, why do you need this?

somnusfish (approved):

controller/src/controller.rs
```diff
@@ -30,40 +34,58 @@ type Result<T> = std::result::Result<T, controllerclient_error::Error>;
...
+    /// Returns a list of all custom definition resource `BottlerocketShadow`/`brs` objects in the cluster.
+    fn all_brss(&self) -> Vec<BottlerocketShadow> {
         self.brs_reader
             .state()
             .iter()
             .map(|arc_brs| (**arc_brs).clone())
             .collect()
     }
 
+    /// Returns a list of all bottlerocket nodes in the cluster.
```
arnaldo2792, line 73: Nit: `s/bottlerocket/Bottlerocket`

controller/src/controller.rs
```diff
@@ -175,6 +197,40 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
...
+                let bottlerocket_shadows: Api<BottlerocketShadow> =
+                    Api::namespaced(self.k8s_client.clone(), NAMESPACE);
+
+                bottlerocket_shadows
+                    .delete(
+                        associated_bottlerocketshadow.as_str(),
+                        &DeleteParams::default(),
+                    )
+                    .await
+                    .context(controllerclient_error::DeleteNode)?;
```
arnaldo2792, line 228: Do you need to log an event after the node is deleted?

arnaldo2792 (approved):

PR #238, opened 2022-08-17
Add support for concurrent update

somnusfish:
**Issue number:**
#163 

**Description of changes:**
Add support for concurrent update

**Testing done:**
Tested on my cluster on number value and unlimited value.
Making sure Brupop run update parallel based on the setting value.

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #241, opened 2022-09-12
Provide a mechanism to allow updates in certain time window

### LINKED ISSUE #67: Provide a mechanism to prevent updates in certain time window
**Description**
  
Customers could have workload that cannot be interrupted during some periods of the day. Therefore, a mechanism to prevent Bottlerocket nodes update by the [bottlerocket-update-operator](https://github.com/bottlerocket-os/bottlerocket-update-operator) in some time window is useful.

**Workarounds**

1. Since, the Bottlerocket update operator expects Kubernetes label `updater-interface-version` on nodes for it to start agent *DaemonSet*. Customer can have an external mechanism to control nodes label `updater-interface-version`.
2. Lock the version of all Bottlerocket hosts to desired version by setting settings.updates.version-lock ([details](https://github.com/bottlerocket-os/bottlerocket/blob/develop/README.md#updates-settings)) and have a mechanism that changes the settings on nodes to latest whenever you would like to update your cluster. 

**Issue or Feature Request:**
Feature Request

gthao313:
users could have workload that cannot be interrupted during some periods of the day, so we provide a mechanism to prevent Bottlerocket nodes update on user customized time window.

**Issue number:**

Closes: #67

**Description of changes:**
```

Provide a mechanism to allow updates in certain update time window

Users could have workload that cannot be interrupted during some periods
of the day, so we provide a mechanism to allow Bottlerocket nodes updates
on user customized update time window. Beyond update time window,
bottlerocker update operator will prevent any updates.
```

**Testing done:**
Set up update time window and then confirm if brupop stops updating nodes except in-process node.

```
UPDATE_WINDOW_START=23:55:0
UPDATE_WINDOW_STOP=0:10:0
```

```
UPDATE_WINDOW_START=0:15:0
UPDATE_WINDOW_STOP=0:24:0
```

**Terms of contribution:**

### DIFF
--- a/.env
+++ b/.env
@@ -25,3 +25,9 @@ EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC=0
 # Note: the "unlimited" option does not work well with `EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC`
 # option, which could potential exclude all nodes from load balancer at the same time.
 MAX_CONCURRENT_UPDATE=1
+
+# Update time window setting
+# Brupop will operate node updates within update time window.
+# when you set up time window start and stop time, you should use UTC (24-hour time notation).
+UPDATE_WINDOW_START=0:0:0
+UPDATE_WINDOW_STOP=0:0:0
--- a/README.md
+++ b/README.md
@@ -79,6 +79,31 @@ For example:
               value: "1"
 ```
 
+##### Set Up Update Time Window
+`UPDATE_WINDOW_START` and `UPDATE_WINDOW_STOP` can be used to specify the time window in which updates are permitted.
+When `UPDATE_WINDOW_START` and `UPDATE_WINDOW_STOP` is 0:0:0 (default), the feature is disabled.
+
+To enable this feature, go to `bottlerocket-update-operator.yaml`, change `UPDATE_WINDOW_START` and `UPDATE_WINDOW__STOP` to a `hour:minute:second` formate value (UTC (24-hour time notation)). Note that `UPDATE_WINDOW_START` is inclusive and `UPDATE_WINDOW_STOP` is exclusive.
+
+To avoid stopping an in-process node update when the update window stops, Bottlerocket update operator reserves 6 mins before `UPDATE_WINDOW_STOP` to finish remaining updates.
+
+Note: brupop uses UTC (24-hour time notation), please convert your local time to UTC.
+For example:
+```yaml
+      containers:
+        - command:
+            - "./controller"
+          env:
+            - name: MY_NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+            - name: UPDATE_WINDOW_START
+              value: "9:0:0"
+            - name: UPDATE_WINDOW_STOP
+              value: "21:0:0"
+```
+
 ### Label nodes
 
 By default, each Workload resource constrains scheduling of the update operator by limiting pods to Bottlerocket nodes based on their labels.
--- a/controller/src/controller.rs
+++ b/controller/src/controller.rs
@@ -8,6 +8,7 @@ use models::node::{
     Selector,
 };
 
+use chrono::{Duration as chrono_duration, NaiveTime, Utc};
 use k8s_openapi::api::core::v1::Node;
 use kube::api::DeleteParams;
 use kube::runtime::reflector::Store;
@@ -25,6 +26,11 @@ const ACTION_INTERVAL: Duration = Duration::from_secs(2);
 
 // Defines environment variable name used to fetch max concurrent update number.
 const MAX_CONCURRENT_UPDATE_ENV_VAR: &str = "MAX_CONCURRENT_UPDATE";
+
+// Defines the update time window related env variable names
+const UPDATE_WINDOW_START_ENV_VAR: &str = "UPDATE_WINDOW_START";
+const UPDATE_WINDOW_STOP_ENV_VAR: &str = "UPDATE_WINDOW_STOP";
+const UPDATE_WINDOW_BUFFER: i64 = 6;
 
 /// The module-wide result type.
 type Result<T> = std::result::Result<T, controllerclient_error::Error>;
@@ -228,6 +234,24 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
         Ok(())
     }
 
+    #[instrument(skip(self), err)]
+    async fn progress_active_set(
+        &self,
+        active_set: BTreeMap<String, BottlerocketShadow>,
+    ) -> Result<()> {
+        // Try to push forward all active nodes, gathering results along the way.
+        let mut nodes: Vec<BottlerocketShadow> = active_set.into_values().collect();
+
+        for brs in nodes.drain(..) {
+            // Timeouts and errors are logged by instrumentation in `progress_node()`.
+            #[allow(unused_must_use)]
+            {
+                self.progress_node(brs).await;
+            }
+        }
+        Ok(())
+    }
+
     /// Runs the event loop for the Brupop controller.
     ///
     /// Because the controller wants to gate the number of simultaneously updating nodes, we can't allow the update state machine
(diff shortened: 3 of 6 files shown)


gthao313:
This PR has conflicts with #238. I'll solve the conflicts when #238 is merged.

controller/src/controller.rs
```diff
@@ -194,25 +196,52 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
...
-                    // Timeouts and errors are logged by instrumentation in `progress_node()`.
-                    #[allow(unused_must_use)]
-                    {
-                        self.progress_node(brs).await;
+            // time window: users can specify a time window to prevent Bottlerocket nodes update. If current time is within the time window,
+            // controller shouldn't have any action on it.
+            match within_time_window()? {
+                // If controller has already started a node update but the time window just start,
+                // controller should respect it, continue to complete the update, and stop actions.
+                true => {
```
somnusfish: I don't think it's good idea to push the active node to complete. Any remaining update circle is likely end up reboot the node  which is what we tried to prevent in the time window.

controller/src/controller.rs
```diff
@@ -296,6 +327,35 @@ fn is_initial_state(brs: &BottlerocketShadow) -> bool {
...
+            date: time_window_start_env,
+        },
+    )?;
+    let time_window_end = NaiveTime::parse_from_str(&time_window_end_env, "%H:%M:%S").context(
+        controllerclient_error::ConvertToNativeTime {
+            date: time_window_end_env,
+        },
+    )?;
+
+    return Ok((current_time > time_window_start) && (current_time < time_window_end));
```
somnusfish: I have two concerns here:
1. What if user accidentally put start time later than end time? 
2. What if user want to block 11pm - 2 am (next day) slot?

gthao313: That's correct concern! I realized it as well and I'm working on the solution.

If users accidentally put start time later than end time, I'll assume this situation as `start time to end time (next day)`. For example, as you mentioned start time `11pm` is later than `2 am`, so brupop will recognize it as 11pm - 2 am (next day) slot. 

I can't recognize `if user accidentally put start time later than end time` as an invalid input since we have valid input like cross a day.

controller/src/controller.rs
gthao313: I think when brupop starts to update nodes, stopping the update immediately due to coming prevent update time window will increase the risk of failure. Although we have mechanism  to handle failure, reducing risk of failure should be the priority.
Meanwhile, the whole update process on a node would be 3 ~ 5mins, I think the short time will not affect anything.

.env
```diff
@@ -25,3 +25,9 @@ EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC=0
 # Note: the "unlimited" option does not work well with `EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC`
```
arnaldo2792, line 25: Nit on commit message:

- Body should begin with a capital letter.
- `Bottlerocket nodes updates`

.env
```diff
@@ -25,3 +25,9 @@ EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC=0
...
 # option, which could potential exclude all nodes from load balancer at the same time.
 MAX_CONCURRENT_UPDATE=1
+
+# Prevent update time window setting
+# Brupop will prevent node updates within time window
+# when you set up time window start and end time, you should use UTC (24-hour time notation).
+PREVENT_UPDATE_TIME_WINDOW_START= 0:0:0
+PREVENT_UPDATE_TIME_WINDOW_END= 0:0:0
```
arnaldo2792: Probably?

```suggestion
PREVENT_UPDATE_SINCE= 0:0:0
PREVENT_UPDATE_UNTIL= 0:0:0
```

gthao313:
Push above: 
1. Add a buffer 6 mins before time window coming - this can make sure that brupop completes in-process updates and doesn't have any actions when time window starts.
2.Add a strategy to deal with if window start time later than end time - will recognize it as cross day period. like 11pm to 2am.
3. rename time window related env variable.
4. rebase

README.md
```diff
@@ -79,6 +79,30 @@ For example:
               value: "1"
 ```
 
+##### Set Up Time Window
+`PREVENT_UPDATE_SINCE` and `PREVENT_UPDATE_UNTIL` can be used to specify the prevent update time window during updating.
+When `PREVENT_UPDATE_SINCE` and `PREVENT_UPDATE_UNTIL` is 0:0:0 (default), the feature is disabled.
+
+
```
jpmcb, line 87: Nit: extra line here

README.md
```diff
@@ -79,6 +79,30 @@ For example:
...
+To enable this feature, go to `bottlerocket-update-operator.yaml`, change `PREVENT_UPDATE_SINCE` and `PREVENT_UPDATE_UNTIL` to a `hour:minute:second` formate value (UTC (24-hour time notation)).
+
+Note: brupop uses UTC (24-hour time notation), please convert your local time to UTC.
```
jpmcb, line 90: One thing reading this doc that's not immediately clear to me is if this is _inclusive_ or _exclusive_ (although this is a bit of a minor concern).

If I set `PREVENT_UPDATE_SINCE` to `19:00` and `PREVENT_UPDATE_UNTIL` to `20:00`, what happens when the clock strikes `20:00` exactly?

it might be worth calling out if this is inclusive of the provided times or exclusive to eliminate any confusion.

controller/src/controller.rs
```diff
@@ -379,6 +413,66 @@ fn node_has_label(node: &Node) -> bool {
...
+        .context(controllerclient_error::ConvertToNativeTimeSnafu {
+            date: prevent_update_since_env,
+        })?;
+    let prevent_update_until = NaiveTime::parse_from_str(&prevent_update_until_env, "%H:%M:%S")
+        .context(controllerclient_error::ConvertToNativeTimeSnafu {
+            date: prevent_update_until_env,
+        })?;
+
+    // Due to the situation that controller has already started a node update but the time window is coming,
+    // we design controller to stop updating any new nodes 6 mins (the time brupop spends to update a node)
```
jpmcb, line 444: Should add this behavior to the docs so users aren't surprised when there's a potential 6 minute buffer between their window and an occurring update.

jpmcb (approved):
This looks good to me! Just a few doc nits

---

Testing:

Build the image from this pull request and deployed to a cluster with `v1.9.1` nodes. I set the following parameters for the time window:
```
            - name: PREVENT_UPDATE_SINCE
              value: "19:0:0"
            - name: PREVENT_UPDATE_UNTIL
              value: "20:0:0"
```
Where the current utc time was `19:45` - I saw that the bottlerocketshadows were idel until just after 20:00 where the update occurred

---

On the discussion if pushing a currently updating BR node to completion if the time window is reached is a good idea or not, I don't have enough context to have a strong opinion. But, I think this behavior is fine for now and if we get user feedback that it's not what they want or expected, we can re-visit it 👍🏼

.env
jpmcb: I think I actually prefer using "start" and "end" as the in name. "Since" and "until" are a bit confusing because "since" could _by itself_ indicate a time period.

But whatever @gthao313 thinks is best works for me!

gthao313: Yeah! I would like use `start` and `end`

controller/src/controller.rs
gthao313: Sure! I'll add it.

.env
```diff
@@ -25,3 +25,9 @@ EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC=0
 # Note: the "unlimited" option does not work well with `EXCLUDE_FROM_LB_WAIT_TIME_IN_SEC`
 # option, which could potential exclude all nodes from load balancer at the same time.
 MAX_CONCURRENT_UPDATE=1
+
+# Prevent update time window setting
+# Brupop will prevent node updates within time window
+# when you set up time window start and end time, you should use UTC (24-hour time notation).
+PREVENT_UPDATE_SINCE=0:0:0
+PREVENT_UPDATE_UNTIL=0:0:0
```
webern: Flipping this around might be easier to understand:

```
UPDATE_WINDOW_BEGIN
UPDATE_WINDOW_END
```

Or

```
UPDATE_WINDOW_START
UPDATE_WINDOW_STOP
```

gthao313: @webern you think it's better to rename or you think use update window logic instead of prevent window logic?

webern: It seems more cognitively intuitive to me since I am familiar with the idea of a "maintenance window", but not a "no maintenance window".

gthao313: I see, that's a good point! 
I'd like to keep "no maintenance window" because actually this feature was asked by customers when we was implementing brupop 0.1.0. Customer wanted us to provide a feature which can prevent update on current time. Therefore, I implemented on this way.

I like maintenance window this idea, if you think it worths to do in this way, I can update it! thanks : )

jpmcb: Hmmm Matt brings up a good point: I'd be surprised if update windows were extremely wide leading to the need for "preventative, no maintenance" windows. My _assumption_ is that most people running kubernetes clusters will have small windows of time when updates _should_ be allowed.

So, that's a +1 from me for `UPDATE_WINDOW_START` and `UPDATE_WINDOW_END`

gthao313:
Push above changes logic from `prevent update  time window` to `update time window` - "maintenance window".

controller/src/controller.rs
```diff
@@ -247,25 +253,54 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
...
+                        let mut nodes: Vec<BottlerocketShadow> = active_set.into_values().collect();
+
+                        for brs in nodes.drain(..) {
+                            // Timeouts and errors are logged by instrumentation in `progress_node()`.
+                            #[allow(unused_must_use)]
+                            {
+                                self.progress_node(brs).await;
+                            }
+                        }
+                    }
```
jpmcb: I see this chunk is pretty similar to the `false` block above. Would it make sense to refactor this into it's own function since they are so similar?

jpmcb: Also, I think I like the variable name `nodes` better than `brss` which I'm guessing stands for "bottle rocket shadows"?

jpmcb (approved):
Deployed a nodegroup with bottlerocket version `1.9.0` and applied the update operator yamls with
```
            - name: UPDATE_WINDOW_START
              value: "17:0:0"
            - name: UPDATE_WINDOW_STOP
              value: "18:0:0"
```
where the _current_ UTC was ~`16:45`. I watched the nodes in the background and saw them begin to update at `17:00`. Great work!!!!! 👏🏼

jpmcb (approved):

README.md
```diff
@@ -79,6 +79,31 @@ For example:
               value: "1"
 ```
 
+##### Set Up Update Time Window
+`UPDATE_WINDOW_START` and `UPDATE_WINDOW_STOP` can be used to specify the update time window during updating.
```
webern:
```suggestion
`UPDATE_WINDOW_START` and `UPDATE_WINDOW_STOP` can be used to specify the time window in which updates are permitted.
```

README.md
```diff
@@ -79,6 +79,31 @@ For example:
...
+When `UPDATE_WINDOW_START` and `UPDATE_WINDOW_STOP` is 0:0:0 (default), the feature is disabled.
+
+To enable this feature, go to `bottlerocket-update-operator.yaml`, change `UPDATE_WINDOW_START` and `UPDATE_WINDOW__STOP` to a `hour:minute:second` formate value (UTC (24-hour time notation)). Note that `UPDATE_WINDOW_START` is inclusive and `UPDATE_WINDOW_STOP` is exclusive.
+
+For avoiding stopping in-process node update when update window stops, Bottlerocket update operator reserves 6 mins before `UPDATE_WINDOW_STOP` to finish remaining update circle.
```
webern:
```suggestion
To avoid stopping an in-process node update when the update window stops, Bottlerocket update operator reserves 6 mins before `UPDATE_WINDOW_STOP` to finish remaining updates.
```

webern (approved):

gthao313:
Push above Rebase and fix README according to Matt's comments.

PR #249, opened 2022-09-22
Use ca.crt in ssl

somnusfish:
**Issue number:**
#232 

**Description of changes:**
Use CA.crt in the SSL configurations.

**Testing done:**
Integration test done.

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #260, opened 2022-09-27
Use ca.crt in ssl - rebased

### LINKED ISSUE #232: Potentially odd TLS certificate setup in v0.2.2
**Image I'm using:**
public.ecr.aws/bottlerocket/bottlerocket-update-operator:v0.2.2

**Issue or Feature Request:**
The way the TLS certificate is configured with the APIServer and agents seems unusual to me. However this maybe a misunderstanding or lack of knowledge on my part.

The `bottlerocket-update-operator.yaml` deployment manifests creates a `Certificate` resource to be self-signed by cert-manager, this certificate is a CA but ALSO has DNS names. Generally speaking CA certs don't have DNS names and are only used to verify and certify child certificates. This has the advantage that the CA private key can be held separate from applications relying on certificates signed by the CA and if one of them is compromised the attacker cannot sign new certificates. Having the CA separate from the client certs also makes rotation easier as you can give a longer life to the CA and a short life to client cert and then rotate the client cert without needing to get all users of the CA to reload the CA in tandem. 

As far as I can tell the Bottlerocket updater agent only uses the TLS certificate to verify the certificate of the APIServer and ignores the extra `ca.crt` file injected in by cert-manager. This means the Client cert always has to be the root CA otherwise communication between the agent and APIServer fails with: `error trying to connect: error:0A000086:SSL routines:tls_post_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:1883: (unable to get local issuer certificate)'''"}`. Also the APIServer and the Agent don't reload the certificate if it changes, as a result when cert-manager rotates the certificate `brupop-apiserver-certificate` (which by default will be after 60 days) a cluster admin would need to bounce all the agent and APIServer pods otherwise communication will fail, including with the Kubernetes API.

I think ideally you would create a longer lived (say 10 years) CA without any DNS Names. Then sign a dedicated cert for the APIServer with a short TTL of maybe a day. The APIServer would automatically reload the cert if it changed (or have a sidecar to trigger a SIGUP). The agent could then use the CA to verify the connection to the APIServer and wouldn't need to worry about the client certificate rotating (for a while).

Also, in `bottlerocket-update-operator.yaml` there also appears to be an unused `Issuer` called `my-ca-issuer`. But there might be a reason for this?

jpmcb:
```
Co-authored-by: Lu Ma <lummam@amazon.com>
Co-authored-by: John McBride <jpmmcb@amazon.com>
```

This PR succeeds and closes #249  (note: it was required to re-create this PR since Lu no longer works on this project)

**Issue number:**

Closes: https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/232

**Description of changes:**

Uses a CA cert in the SSL configuration

**Testing done:**

Running integration tests now, but should be fine since #249 worked with it's integration tests

**Terms of contribution:**

### DIFF
--- a/apiserver/src/api/mod.rs
+++ b/apiserver/src/api/mod.rs
@@ -15,8 +15,8 @@ use crate::{
     telemetry,
 };
 use models::constants::{
-    AGENT, APISERVER_HEALTH_CHECK_ROUTE, APISERVER_SERVICE_NAME, LABEL_COMPONENT, NAMESPACE,
-    PRIVATE_KEY_NAME, PUBLIC_KEY_NAME, TLS_KEY_MOUNT_PATH,
+    AGENT, APISERVER_HEALTH_CHECK_ROUTE, APISERVER_SERVICE_NAME, CA_NAME, LABEL_COMPONENT,
+    NAMESPACE, PRIVATE_KEY_NAME, PUBLIC_KEY_NAME, TLS_KEY_MOUNT_PATH,
 };
 use models::node::{BottlerocketShadowClient, BottlerocketShadowSelector};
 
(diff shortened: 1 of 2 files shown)


gthao313 (approved):

stmcginnis (approved):
Looks good.

jpmcb:
Thanks Sean! 👏🏼

Issue #121, opened 2021-11-16
0.2.0: Do away with crate-wide error modules


cbgbt:
Error submodules are preferred for error types, and it is preferable to create multiple error classifications rather than bubbling all errors up to a single error type.

cbgbt:
Might be worth dividing this up into tasks for agent/apiserver/controller/etc to make it more tractable.

gthao313:
We have `apiserver` and `model` left.

PR #279, opened 2022-10-07
Refactors `apiserver` errors to be module specific

jpmcb:
**Issue number:**

Related: https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/121

**Description of changes:**

Refactors the `apiserver` error handling to be less generic and more module spcific.

**Testing done:**

_Coming soon_

**Terms of contribution:**

### DIFF
--- a/apiserver/src/api/drain.rs
+++ b/apiserver/src/api/drain.rs
@@ -1,5 +1,6 @@
+use super::error;
+use super::Result;
 use super::{APIServerSettings, ApiserverCommonHeaders};
-use crate::error::{self, Result};
 use models::node::BottlerocketShadowClient;
 
 use actix_web::{
--- a/apiserver/src/api/error.rs
+++ b/apiserver/src/api/error.rs
@@ -3,16 +3,9 @@ use models::node::BottlerocketShadowError;
 use actix_web::error::ResponseError;
 use snafu::Snafu;
 
-/// The crate-wide result type.
-pub type Result<T> = std::result::Result<T, Error>;
-
-/// The crate-wide error type.
 #[derive(Debug, Snafu)]
 #[snafu(visibility(pub))]
 pub enum Error {
-    #[snafu(display("Unable to create client: '{}'", source))]
-    ClientCreate { source: kube::Error },
-
     #[snafu(display("Unable to parse HTTP header. Missing '{}'", missing_header))]
     HTTPHeaderParse { missing_header: &'static str },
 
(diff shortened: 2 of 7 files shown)


apiserver/src/api/error.rs
```diff
@@ -49,4 +37,4 @@ pub enum Error {
     WebhookError { source: serde_json::error::Error },
```
gthao313, line 37: I _think_ brupop doesn't use this error anymore. Maybe better to remove it. : )

apiserver/src/api/error.rs
```diff
@@ -3,16 +3,9 @@ use models::node::BottlerocketShadowError;
...
-pub type Result<T> = std::result::Result<T, Error>;
-
-/// The crate-wide error type.
 #[derive(Debug, Snafu)]
 #[snafu(visibility(pub))]
-pub enum Error {
-    #[snafu(display("Unable to create client: '{}'", source))]
-    ClientCreate { source: kube::Error },
-
+pub enum ApiError {
```
gthao313: Do you have preferences and any thoughts that make you keep this `error.rs`? I think might be better to delete this file and move the ApiError module to `mod.rs`? I'm not 100% confident that my way is better than yours. Love to hear your thoughts! thanks

apiserver/src/api/drain.rs
```diff
@@ -1,7 +1,9 @@
 use super::{APIServerSettings, ApiserverCommonHeaders};
-use crate::error::{self, Result};
 use models::node::BottlerocketShadowClient;
```
webern: I wouldn't add a new line here. no newlines in 'use' statements and let rustfmt do its thing.

```suggestion
```

apiserver/src/api/error.rs
webern: Throughout our codebase (including other projects) we tend to use `Error` for all error enums instead of giving them bespoke names.

webern (approved):

apiserver/src/api/error.rs
jpmcb: This made sense to live in it's own file since it denotes errors for the entire API module. I _believe_ putting it in `mod.rs` would essentially accomplish the same thing; make these errors accessible to entire module (but not necessarily the crate). Still new to the rust paradigms, so up to what y'all think!

jpmcb: Sounds good! I was following some of the other examples found in brupop:

https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/7f8ac76f338af68902d9559ab9d5ede874fe472e/apiserver/src/auth/error.rs#L7

Maybe these are due for a small refactor?

jpmcb:
Force pushed:
- Resolved formatting with `cargo fmt`
- `ApiError` --> `Error`

webern (approved):

webern (approved):

gthao313 (approved):

PR #280, opened 2022-10-10
Reload certificate periodically

### LINKED ISSUE #233: Reload certificate periodically
**Image I'm using:**

**Issue or Feature Request:**
Related issue #232, the APIServer and the Agent don't reload the certificate if it changes, as a result when cert-manager rotates the certificate brupop-apiserver-certificate (which by default will be after 60 days) a cluster admin would need to bounce all the agent and APIServer pods otherwise communication will fail, including with the Kubernetes API.

gthao313:
when cert-manager rotates the certificate brupop-apiserver-certificate (which by default will be after 60 days), brupop apiserver needs to reload certificate to bounce all the agent and APIServer pods.

**Issue number:**
Closes: #233 

**Description of changes:**
```
Date:   Mon Oct 10 17:46:26 2022 +0000

    Reload certificate periodically

    when cert-manager rotates the certificate brupop-apiserver-certificate
    (which by default will be after 60 days), brupop apiserver needs to reload
    certificate to bounce all the agent and APIServer pods.
```

Add a mechanism to monitor if certificate has been renewed. If yes, stop servers and apiserver will be restarted to reload new certificate.  Agent doesn't need do anything since the logic in brupop that agent will always use current certificate on volume to talk with apiserver.  

**Testing done:**
1. Set Certificate to be expired on 7 mins 
```
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: brupop-apiserver-certificate
  namespace: brupop-bottlerocket-aws
spec:
  isCA: true
  commonName: my-selfsigned-ca
  secretName: brupop-tls
  renewBefore: 2159h53m
```

2. Run integration test to monitor if all nodes still are updated to latest version.

**Terms of contribution:**

### DIFF
--- a/apiserver/src/api/mod.rs
+++ b/apiserver/src/api/mod.rs
@@ -18,10 +18,10 @@ use models::constants::{
     AGENT, APISERVER_HEALTH_CHECK_ROUTE, APISERVER_SERVICE_NAME, CA_NAME, LABEL_COMPONENT,
     NAMESPACE, PRIVATE_KEY_NAME, PUBLIC_KEY_NAME, TLS_KEY_MOUNT_PATH,
 };
-use models::node::{BottlerocketShadowClient, BottlerocketShadowSelector};
+use models::node::{read_certificate, BottlerocketShadowClient, BottlerocketShadowSelector};
 
 use actix_web::{
-    dev::ServiceRequest,
+    dev::{Server, ServiceRequest},
     http::{self, HeaderMap},
     web::{self, Data},
     App, HttpServer,
@@ -38,6 +38,7 @@ use openssl::ssl::{SslAcceptor, SslFiletype, SslMethod};
 use opentelemetry::global::meter;
 use snafu::{OptionExt, ResultExt};
 use std::env;
+use tokio::time::{sleep, Duration};
 use tracing::{event, Level};
 use tracing_actix_web::TracingLogger;
 
@@ -45,6 +46,8 @@ use std::convert::TryFrom;
 
 // The set of API endpoints for which `tracing::Span`s will not be recorded.
 pub const NO_TELEMETRY_ENDPOINTS: &[&str] = &[APISERVER_HEALTH_CHECK_ROUTE];
+
+const CERTIFICATE_DETECTOR_SLEEP_DURATION: Duration = Duration::from_secs(60);
 
 /// The API module-wide result type.
 type Result<T> = std::result::Result<T, error::Error>;
@@ -101,6 +104,10 @@ pub async fn run_server<T: 'static + BottlerocketShadowClient>(
     k8s_client: kube::Client,
     prometheus_exporter: Option<opentelemetry_prometheus::PrometheusExporter>,
 ) -> Result<()> {
+    let public_key_path = format!("{}/{}", TLS_KEY_MOUNT_PATH, PUBLIC_KEY_NAME);
+    let certificate_cache =
+        read_certificate(&public_key_path).context(error::ReadCertificateFailedSnafu)?;
+
     let server_port = settings.server_port;
 
     // Set up a reflector to watch all kubernetes pods in the namespace.
@@ -223,6 +230,10 @@ pub async fn run_server<T: 'static + BottlerocketShadowClient>(
             event!(Level::ERROR, "reflector drained");
             return Err(error::Error::KubernetesWatcherFailed {});
         },
+        _ = reload_certificate(server.clone(), &public_key_path, certificate_cache)=> {
+            event!(Level::ERROR, "certificate refreshed");
+            return Err(error::Error::ReloadCertificateFailed {});
+        },
         res = server => {
             event!(Level::ERROR, "server exited");
             res.context(error::HttpServerSnafu)?;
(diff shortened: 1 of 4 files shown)


apiserver/src/api/mod.rs
```diff
@@ -229,6 +239,30 @@ pub async fn run_server<T: 'static + BottlerocketShadowClient>(
...
+    loop {
+        let current_certificate =
+            read_certificate(public_key_path).context(error::ReadCertificateFailedSnafu)?;
+        if current_certificate != certificate_cache {
+            event!(
+                Level::INFO,
+                "Certificate has been renewed, restarting server to reload new certificate"
+            );
+            server.stop(true).await;
+            std::process::exit(0)
```
webern: I don't like to see a process exit buried so deeply in a program. Ideally you would return a bool (or enum) that would say whether or not an exit is needed. Bubble that all the way up to main if possible. Ideally only main should have process exits IMO.

webern (approved):

apiserver/src/api/mod.rs
gthao313: After testing, I think we don't need `process exit` since `server.stop(true).await;` will restart apiserver. I'll just remove this `std::process::exit(0)`. thanks

gthao313:
rebase

models/src/node/mod.rs
```diff
@@ -29,3 +34,30 @@ pub const K8S_NODE_KIND: &str = "BottlerocketShadow";
...
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub))]
+    pub enum Error {
+        #[snafu(display(
+            "IO error occurred while attempting to use APIServerClient: '{}'",
+            source
+        ))]
+        IOError { source: Box<dyn std::error::Error> },
+    }
+}
```
jpmcb: Would this error be better placed in the `node/error.rs` file with the rest of this modules errors? (also see comment above where we seem to already have a Result type for this module

models/src/node/mod.rs
```diff
@@ -9,6 +9,11 @@ pub use self::error::Error as BottlerocketShadowError;
 
 use lazy_static::lazy_static;
 pub use semver::Version;
+use snafu::ResultExt;
+use std::io::Read;
+
+/// The module-wide result type.
+type Result<T> = std::result::Result<T, mod_error::Error>;
```
jpmcb: Isn't this already defined here?

https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/72bf1b36c003bc270daee59464d68a3fee2eaf1f/models/src/node/error.rs#L5

apiserver/src/api/mod.rs
```diff
@@ -101,6 +104,9 @@ pub async fn run_server<T: 'static + BottlerocketShadowClient>(
     k8s_client: kube::Client,
     prometheus_exporter: Option<opentelemetry_prometheus::PrometheusExporter>,
 ) -> Result<()> {
+    let public_key_path = format!("{}/{}", TLS_KEY_MOUNT_PATH, PUBLIC_KEY_NAME);
+    let certificate_cache = read_certificate(&public_key_path).unwrap();
```
jpmcb: Let's replace this `unwrap` with a structured error in the API module (reference: https://github.com/bottlerocket-os/bottlerocket/commit/5a58d4b2b98dbed06a103d866b4b753f580f0e32)

apiserver/src/api/mod.rs
```diff
@@ -232,6 +242,29 @@ pub async fn run_server<T: 'static + BottlerocketShadowClient>(
     Ok(())
 }
 
+// Certificate is refreshed periodically (default 60 days). Once certificate is renewed, apiserver
+// needs to stop the server for reloading new certificate.
+// We cache the certificate initially when brupop starts server, and compare to update-to-date certificate periodically.
+// If they aren't match, we recognize it as new certificate is coming out, so the server needs to be restarted.
```
jpmcb: A few nits here:

```suggestion
// The certificate is refreshed periodically (default 60 days). Once the certificate is renewed, the apiserver
// needs to stop in order to reload the new certificate.
// We cache the certificate initially when brupop starts the server, and compare it to the update-to-date certificate periodically.
// If they don't match, we recognize it as a new certificate, so the server needs to be restarted.
```

jpmcb (commented):
Awesome! 🕺🏼

gthao313:
push above address @jpmcb comments

jpmcb (approved):
🥇

apiserver/src/api/mod.rs
```diff
@@ -232,6 +243,29 @@ pub async fn run_server<T: 'static + BottlerocketShadowClient>(
     Ok(())
 }
 
+// The certificate is refreshed periodically (default 60 days). Once the certificate is renewed, the apiserver
+// needs to stop in order to reload the new certificate.
+// We cache the certificate initially when brupop starts the server, and compare it to the update-to-date certificate periodically.
+// If they don't match, we recognize it as a new certificate, so the server needs to be restarted.
+async fn reload_certificate(
+    server: Server,
```
cbgbt, line 251: Sorry to comment on an old PR, but I was catching up on some of the stuff that's changed in brupop!

I wonder if it makes more sense to make this a mutable reference? It seems unusual to clone it. Although, I suppose if it works, Actix must have done some work to make sure that the interior state of the Server is thread-safe.

somnusfish: I actually find out that using https://github.com/notify-rs/notify to watch the fs change and reload the certificate is what other people like to do.

jpmcb: @cbgbt - This has actually changed recently: https://github.com/bottlerocket-os/bottlerocket-update-operator/commit/561e6c0c1251cee2c9c6694232fae52564ce6419#diff-5cad36d03cb8802a9c5a0a20ae2e841d683e1cd564146c441e2940fb1ecf652eR255

A few APIs changed in Actix so cloning the server and using it directly wasn't a great option. Instead, now, we can use a `actix_web::dev::ServerHandle` as a "server handler" to call into it and stop it like:
```rust
 server_handler.stop(true).await;
```

PR #327, opened 2022-10-28
Add resources requests and limits for all containers

### LINKED ISSUE #240: How many cpu/memory does brupop use
Asked by customer. Need investigate on how many cpu/memory resources does brupop use.

jpmcb:
### Issue number:

Closes #240

### Description of changes:

This patch provides the expected resource (cpu and memory) requests for each container along with some reasonable limits. These numbers were derived from several experiments (noted in #240 and in the process of creating this patch): the `10m` CPU and `50Mi` memory limit is due to components having spikes in resource requirements when nodes rotate and pods are re-scheduled (during upgrade scenarios where the agent, api-server, and/or controller exist on the upgrading node). The `request` delimiter allows for more resources to be consumed (if they are available on the node) and should be safe since we drain the nodes before doing an upgrade. Having _too tight_ of resource limits will cause the system's kernel to kill the process, so these are intentionally forgiving.

**_TODO_**:

- [x] documentation for consumers of brupop

### Testing done:

Several integration tests were run with the `metric-server` and kubernetes `dashboard` also installed to monitor resource utilization. 

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,8 +77,25 @@ The default values are generated from the [`.env`](https://github.com/bottlerock
   It is used by the node agents to access the API server.
   If this environment variable is _not_ found, the Brupop agents will fail to start.
 
-#### Exclude Nodes from Load Balancers Before Draining
-This configuration uses the Kuberenetes [ServiceNodeExclusion](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) feature.
+##### Resource Requests & Limits
+
+The `bottlerocket-update-operator.yaml` manifest makes several default recommendations for
+Kubernetes resource requests and limits. In general, the update operator and it's components are lite-weight
+and shouldn't consume more than 10m CPU (which is roughly equivalent to 1/100th of a CPU core)
+and 50Mi (which is roughly equivalent to 0.05 GB of memory).
+If this limit is breached, the Kubernetes API will restart the faulting container.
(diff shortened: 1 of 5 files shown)


jpmcb:
Force pushed to add documentation ✏️ 

This PR is ready to go, but leaving in draft for now so we don't get in the way of the upcoming v1.0.0 release

jpmcb:
Marking ready for review now that 1.0.0 is out and force pushed to address a merge conflict in the `README.md` documentation!

jpmcb:
Final verification experiment:

1. Installed [the kubernetes metrics server](https://github.com/kubernetes-sigs/metrics-server)
2. Installed [the kubernetes dashboard](https://github.com/kubernetes/dashboard) with an admin user
3. Ran integration tests off this patch using the modified yaml manifest and the default `1.0.0`image.

### Results

Screen Shot 2022-11-03 at 2 52 05 PM

#### API server
Memory below 3Mi
CPU at 1.00 M

#### Controller
Memory stable ~2.75 Mi with small spikes below 3 Mi
CPU at 1.00 M or lower

#### Agents
Memory below 7 Mi
CPU at 1.00 M with small spikes up to ~3 M during node updates

webern (approved):

gthao313 (approved):

PR #340, opened 2022-11-08
TLS: Utilize leaf certs from root CA for brupop API server and agent

### LINKED ISSUE #232: Potentially odd TLS certificate setup in v0.2.2

jpmcb:
### Issue number:

Closes: #232

### Description of changes:

```
- Ensures that the agent does not get the entire cert bundle from the
  brupop api-server. Instead, get's it's own cert that is signed by the
  root cert.
- Renames variable to make more sense for leaf certs on both apiserver
  and client.

```

This change fixes our usage of a root CA as _both_ server and client certificates. The original circular chain of trust meant that the agent always had root certificate privileges and would have the entire server's cert chain mounted.

Instead, this patch creates the following chain of trust:
1. Initialize a self signed "bootstrap" issuer. 
2. The bootstrap issuer is utilized to create the root CA (which is effectively also self signed)
3. Using the root CA, initialize a "Brupop root certificate" issuer that can generate leaf certificates
4. Generate leaf certificates for both the api server and the agent utilizing the brupop root certificate issuer
    -  Set the `dnsName` for the apisever certificate to cover the brupop namespace: `*.brupop-bottlerocket-aws.svc.` addresses
    - Set the `dnsName` for the agent certificate to cover the brupop namespace: `*.brupop-bottlerocket-aws.svc.` addresses
6. The brupop API server mounts and uses the entire chain (it's CA, it's private key, and it's public key) and sends these certs to be validated by the agent.
7. The brupop agent mounts and uses it's cert build a chain of trust it can use to validate the server

### Testing done:

1. Built an image, pushed to ECR and updated the `image` tags in the yaml manifest to point to my testing image
2. Ran the integration tests
3. All looks good 👍🏼 - upgraded to newer version of bottlerocket and no logs indicating a problem
```
❯ k get brs -A
brupop-bottlerocket-aws   brs-ip-192-168-135-135.us-west-2.compute.internal   Idle    1.10.1    Idle                            0
brupop-bottlerocket-aws   brs-ip-192-168-150-220.us-west-2.compute.internal   Idle    1.10.1    Idle                            0
brupop-bottlerocket-aws   brs-ip-192-168-158-91.us-west-2.compute.internal    Idle    1.10.1    Idle                            0
```

**Terms of contribution:**

### DIFF
--- a/apiserver/src/client/webclient.rs
+++ b/apiserver/src/client/webclient.rs
@@ -117,17 +117,17 @@ impl K8SAPIServerClient {
 
     /// Returns the https client configured to use self-signed certificate
     fn https_client() -> Result<reqwest::Client> {
-        let mut buf = Vec::new();
+        let mut cert_buf = Vec::new();
 
-        let ca_path = format!("{}/{}", TLS_KEY_MOUNT_PATH, CA_NAME);
-        std::fs::File::open(ca_path)
+        let leaf_cert_path = format!("{}/{}", TLS_KEY_MOUNT_PATH, CA_NAME);
+        std::fs::File::open(leaf_cert_path)
             .map_err(|err| Box::new(err) as Box<dyn std::error::Error>)
             .context(error::IOSnafu)?
-            .read_to_end(&mut buf)
+            .read_to_end(&mut cert_buf)
             .map_err(|err| Box::new(err) as Box<dyn std::error::Error>)
             .context(error::IOSnafu)?;
 
-        let cert = reqwest::Certificate::from_pem(&buf).context(error::CreateClientSnafu)?;
+        let cert = reqwest::Certificate::from_pem(&cert_buf).context(error::CreateClientSnafu)?;
 
         let client = reqwest::Client::builder()
             .add_root_certificate(cert)
--- a/yamlgen/deploy/cert.yaml
+++ b/yamlgen/deploy/cert.yaml
@@ -1,19 +1,36 @@
 ---
 apiVersion: cert-manager.io/v1
-kind: ClusterIssuer
+kind: Issuer
 metadata:
   name: selfsigned-issuer
+  namespace: brupop-bottlerocket-aws
 spec:
   selfSigned: {}
 ---
 apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: brupop-selfsigned-ca
+  namespace: brupop-bottlerocket-aws
+spec:
+  isCA: true
+  commonName: brupop-selfsigned-ca
+  secretName: brupop-root-ca-secret
+  privateKey:
+    algorithm: RSA
+    encoding: PKCS8
+  issuerRef:
+    name: selfsigned-issuer
+    kind: Issuer
+---
+apiVersion: cert-manager.io/v1
 kind: Issuer
 metadata:
-  name: my-ca-issuer
+  name: brupop-root-certificate-issuer
   namespace: brupop-bottlerocket-aws
 spec:
   ca:
-    secretName: brupop-tls
+    secretName: brupop-root-ca-secret
 ---
 apiVersion: cert-manager.io/v1
 kind: Certificate
@@ -21,16 +38,38 @@ metadata:
   name: brupop-apiserver-certificate
   namespace: brupop-bottlerocket-aws
 spec:
-  isCA: true
-  commonName: my-selfsigned-ca
-  secretName: brupop-tls
+  secretName: brupop-apiserver-certificate
   privateKey:
     algorithm: RSA
     encoding: PKCS8
   dnsNames:
-    - brupop-apiserver.brupop-bottlerocket-aws.svc.cluster.local
-    - brupop-apiserver.brupop-bottlerocket-aws.svc
+    - "*.brupop-bottlerocket-aws.svc.cluster.local"
+    - "*.brupop-bottlerocket-aws.svc"
+  usages:
+    - server auth
+    - key encipherment
+    - digital signature
   issuerRef:
-    name: selfsigned-issuer
-    kind: ClusterIssuer
-    group: cert-manager.io
+    name: brupop-root-certificate-issuer
+    kind: Issuer
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: brupop-apiserver-client-certificate
+  namespace: brupop-bottlerocket-aws
+spec:
+  secretName: brupop-apiserver-client-certificate
+  privateKey:
+    algorithm: RSA
+    encoding: PKCS8
+  dnsNames:
+    - "*.brupop-bottlerocket-aws.svc.cluster.local"
+    - "*.brupop-bottlerocket-aws.svc"
+  usages:
+    - client auth
+    - key encipherment
+    - digital signature
+  issuerRef:
+    name: brupop-root-certificate-issuer
+    kind: Issuer
--- a/models/src/agent.rs
+++ b/models/src/agent.rs
@@ -1,7 +1,7 @@
 use crate::brupop_labels;
 use crate::constants::{
     AGENT, AGENT_NAME, APISERVER_SERVICE_NAME, APP_COMPONENT, APP_MANAGED_BY, APP_PART_OF, BRUPOP,
-    BRUPOP_INTERFACE_VERSION, LABEL_BRUPOP_INTERFACE_NAME, LABEL_COMPONENT, NAMESPACE, SECRET_NAME,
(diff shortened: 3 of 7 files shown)


jpmcb:
Force pushed to address `cargo fmt`.

Marking as ready for review!! 👏🏼

jpmcb:
One thing we'll want to think about here (and would be great to get @gthao313 eyes on) is how this may affect the feature of periodically rotating our certificates (per https://github.com/bottlerocket-os/bottlerocket-update-operator/pull/280)

I don't believe this will have any impact since the agent will always use whatever certificate is mounted when it generates a client to the apiserver and when the api-server restarts, it'll pick up the new certificate when it generates it's server.

In essence, the logic still stands but the underlying certificates themselves are different which doesn't change the functionality.

apiserver/src/client/webclient.rs
```diff
@@ -130,6 +130,7 @@ impl K8SAPIServerClient {
         let cert = reqwest::Certificate::from_pem(&buf).context(error::CreateClientSnafu)?;
 
         let client = reqwest::Client::builder()
+            .danger_accept_invalid_certs(true)
```
jpmcb: I actually think that `.danger_accept_invalid_certs(true)` is not what we want here. This will accept any cert, from any website, even those that are expired.

Instead, we should extend the chain of trust to know about our [self-signed cert via `.add_root_certificate`](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.add_root_certificate)

Thanks @yeazelm for the tip!!

webern (approved):
Not an expert on this, but looks good.

webern (approved):

yamlgen/deploy/cert.yaml
```diff
@@ -1,36 +1,75 @@
...
+spec:
+  secretName: brupop-apiserver-client-certificate
+  privateKey:
+    algorithm: RSA
+    encoding: PKCS8
+  dnsNames:
+    - "*.brupop-bottlerocket-aws.svc.cluster.local"
+    - "*.brupop-bottlerocket-aws.svc"
+  usages:
+    - server auth
```
cbgbt: Should this be `client auth` instead? (assuming we would ultimately use this keypair to accomplish mTLS)

apiserver/src/client/webclient.rs
```diff
@@ -119,8 +119,8 @@ impl K8SAPIServerClient {
     fn https_client() -> Result<reqwest::Client> {
         let mut buf = Vec::new();
 
-        let ca_path = format!("{}/{}", TLS_KEY_MOUNT_PATH, CA_NAME);
-        std::fs::File::open(ca_path)
+        let leaf_cert_path = format!("{}/{}", TLS_KEY_MOUNT_PATH, PUBLIC_KEY_NAME);
+        std::fs::File::open(leaf_cert_path)
```
cbgbt: I'm learning new things reviewing this PR, so grain of salt,

but don't we still want this to be the `ca.crt` file, rather than the client's public key? My understanding is that the client is using this to validate that the server's public key is signed by our CA.

yamlgen/deploy/cert.yaml
cbgbt: We could then do something like [this](https://github.com/camelop/rust-mtls-example/blob/main/src/main.rs#L34) for the other bit -- using the client's keypair to establish mutual trust via the `identity()` method.

jpmcb: Good call - yes that makes much more sense and will set us up well for mTLS

apiserver/src/client/webclient.rs
jpmcb: True - that does make more sense! Made that change!

jpmcb:
Force pushed to address a few comments

The client now uses it's CA cert which is the leaf from the root CA. Also adjusted the client cert and it's `Usage` to have `client auth` instead of `server auth`

Ran through a few rounds of integration tests and 👍🏼

cbgbt:
This LGTM; however, if we're only a few lines away from mTLS here should we just do it?

jpmcb:
> if we're only a few lines away from mTLS here should we just do it?

Might as well! Thanks for all the feedback @cbgbt !!! 👏🏼 hopefully will have that unwound in a bit

jpmcb:
~Force pushed to implement mTLS - updated top level PR description.~

~Now, the agent is also building it's `identity` with it's private key and public key! I had considered using [an experimental feature for cert-manager](https://cert-manager.io/docs/usage/certificate/#additional-certificate-output-formats) that would _also_ output the `<private key PEM + "\n" + public key PEM>`, but we can't be sure if that will become a stable feature or if users will have that feature gate turned on. Hence, the extra code to build the PEM identity chain file ourselves.~

Edit: See force push and comment below

jpmcb:
Force pushed to revert the mTLS change.

Had several internal discussions (including with Amazon security) and the team decided that introducing full mTLS for pod to pod communication is not quite what we want to do here:
- We'd be introducing a possible additional attack vector on the brupop api-server since _any_ client could potentially connect and send a vulnerable certificate that could result in remote code execution (like the CVE we saw last month with OpenSSL)
  - We already have TLS one way where the agent is validating the server's certs.
  - We also already have validation of the agent from the server side with the [Kubernetes service account token](https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/d1c3ff409eca3e8587f9cb99b9d22a30d9ecd91c/apiserver/src/auth/authorizor.rs#L50). This token is a secret who's chain of trust is based on values passed to `--service-account-issuer`, `--service-account-key-file`, and `--service-account-signing-key-file` when the `kube-apiserver` starts. And since we are using [token volume projection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#serviceaccount-token-volume-projection), we can be assured that this token will rotate frequently (at least every 24 hours).
- Therefore, introducing mTLS doesn't get us much in mutual verification while potentially introducing another vector of attack on the brupop api server.

Validated this with another round of end to end tests and I see an update from 1.9.0 to 1.11.1 👍🏼 
```
❯ k get brs -A
brupop-bottlerocket-aws   brs-ip-192-168-153-127.us-west-2.compute.internal   Idle    1.11.1    Idle                            0
brupop-bottlerocket-aws   brs-ip-192-168-155-145.us-west-2.compute.internal   Idle    1.11.1    Idle                            0
brupop-bottlerocket-aws   brs-ip-192-168-155-42.us-west-2.compute.internal    Idle    1.11.1    Idle                            0
```

cbgbt (approved):
Thanks for the tenacity here, nice work! :rocket:

webern (approved):

PR #342, opened 2022-11-17
Expose failure output for apiclient

### LINKED ISSUE #335: Expose failure output for apiclient  🔄  agent interactions
Right now, we make many assumptions about the failure state when calling into `apiclient` on the bottlerocket hosts (primarily from the agent). For example:

https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/eebfbcde67cd6606125c23ed815d0db53f407ff3/agent/src/apiclient.rs#L176-L185

The agent calls to the bottlerocket API for a refresh but this could fail for any number of reasons. We are currently returning the status code, but we should return the entire output to the user in the error for easier debugging

gthao313:
The agent calls to the bottlerocket API and it could fail for any number of reasons. Currently we only return the status code, so we improve update API invoke functionality to return entire error output for easier debugging.

**Issue number:**
Closes: #335 

**Description of changes:**
```
Date:   Thu Nov 17 19:23:43 2022 +0000

    Expose failure output for apiclient

    The agent calls to the bottlerocket API and it could fail for any number
    of reasons. Currently we only return the status code, so we improve
    update API invoke functionality to return entire error output for easier
    debugging.
```

**Testing done:**
:white_check_mark: Integration test

**Terms of contribution:**

### DIFF
--- a/agent/src/apiclient.rs
+++ b/agent/src/apiclient.rs
@@ -150,6 +150,8 @@ async fn invoke_apiclient(args: Vec<String>) -> Result<Output> {
                             _ => {
                                 // API response was a non-transient error, bail out
                                 return apiclient_error::BadHttpResponseSnafu {
+                                    args: args,
+                                    error_content: &error_content,
                                     statuscode: error_statuscode,
                                 }
                                 .fail();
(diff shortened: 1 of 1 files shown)


gthao313:
@jpmcb Since you create the issue, Can you take a look at this to verify if this change meet your expectation? Thanks! : )

agent/src/apiclient.rs
```diff
@@ -150,7 +150,8 @@ async fn invoke_apiclient(args: Vec<String>) -> Result<Output> {
                             _ => {
                                 // API response was a non-transient error, bail out
                                 return apiclient_error::BadHttpResponseSnafu {
-                                    statuscode: error_statuscode,
+                                    args: args,
```
cbgbt: Might be good to leave the status code there, which would make it easier to write code to handle the error elsewhere.

jpmcb: +1 Agree with Sean. I think it'd make sense to have all 3:
- the command /args that were invoked
- the error status code
- the error contents

jpmcb (approved):

webern (approved):

PR #350, opened 2022-12-01
Remove `yamlgen`. Refactor deployment and yaml generation to use Helm templating

### LINKED ISSUE #126: 0.2.0: Create Brupop Helm Chart
**Issue or Feature Request:**
Brupop should be installable via Helm. This would provide customers an easy route to install dependencies, like [cert-manager](https://cert-manager.io/) for managing SSL certificates used for communication with the apiserver, or the OpenTelemetry Collector, should we integrate with that rather than direct Prometheus integration.

jpmcb:
**Issue number:**

Closes #126 

**Description of changes:**

Refactors deployment and yaml templating to use Helm. Still a work in progress. Using this draft PR as a staging area for early feedback

**Testing done:**

Able to:
- `make` images
- `helm install brupop-crd deploy/charts/bottlerocket-shadow`
- `helm install brupop deploy/charts/bottlerocket-update-operator`

**Terms of contribution:**


jpmcb:
Force pushed to update certificate resources that changed with #340 - Ran through successful updates using:

```
$ helm install brupop-crd --create-namespace deploy/charts/bottlerocket-shadow
$ helm install brupop deploy/charts/bottlerocket-update-operator
```

jpmcb:
Force pushed:
- fixed build errors
- resolved merge conflict (with `Cargo.lock` update)
- Updated READMEs with documentation on how to install brupop via helm
- added `make manifest` target that can be used to generate manifest at root of repository
- Removed `.env` file (in favor of helm values yaml file)
- Added logic to create / delete brupop namespace in integration tests

Marking this as ready for review.

There are no big things left to do here and I'd love to start getting some feedback from the community. I'm sure there are little things left here since this is a bigger refactor, but we can continue to tackle those in the coming weeks!

We'll also want to think about doing a one off change to the new release workflow to ensure that works for releases. Aka, we can change that to trigger on each push to `develop` and ensure it's working properly. That workflow is designed to push a helm repo to a github page hosted in this repository. We'll also need to ensure we have the right access right associated with that so it can be created on release automatically.

deploy/charts/bottlerocket-shadow/templates/NOTES.txt
```diff
@@ -0,0 +1,6 @@
+Thank you for installing the Bottlerocket Shadow CRD!
+
+This CRD is meant to be used with the bottlerocket-update-operator (affectionatly called "brupop")
```
stmcginnis: s/affectionatly/affectionately/

README.md
```diff
@@ -20,17 +20,52 @@ kubectl apply -f \
...
 
 ```sh
-kubectl apply -f ./bottlerocket-update-operator.yaml
+# Add the cert-manager helm chart
+helm repo add jetstack https://charts.jetstack.io
+
+# Update your local chart cache with the latest
+helm repo update
+
+# Install the cert-manager (including it's CRDs)
```
stmcginnis: nit: "it's" means "it is", "its" is possessive.

stmcginnis (commented):
Looks good to me. There are a lot of changes here, but with the noted testing done I feel comfortable with the change.

Looks like there is a merge conflict to be resolved now. Since an update is needed anyway, a couple small things pointed out.

README.md
```diff
@@ -41,6 +76,55 @@ kubectl label node MY_NODE_NAME bottlerocket.aws/updater-interface-version=2.0.0
...
+#  --docker-server 109276217309.dkr.ecr.us-west-2.amazonaws.com \
+#  --docker-username=AWS \
+#  --docker-password=$(aws --region us-west-2 ecr get-login-password) \
+#  --namespace=brupop-bottlerocket-aws
+#image_pull_secrets: |-
+#  - name: "brupop"
+
+# External load balancer setting.
+# When `exclude_from_lb_wait_time_in_sec` is set to positive value
+# brupop will exclude the node from load balancer and
```
cbgbt: I might clean up some of the language on these comments since they are lifted from what was previously a dev-level config file, but e.g. here we should probably clarify that this uses `"node.kubernetes.io/exclude-from-external-load-balancers"` labels under the hood.

deploy/charts/bottlerocket-update-operator/templates/NOTES.txt
```diff
@@ -0,0 +1,11 @@
+Thank you for installing Brupop!
+
+The bottlerocket-update-operator will now begin querying the updater API for updates.
```
cbgbt: nit: clean up wording to clarify which API is being discussed.

```suggestion
The bottlerocket-update-operator will now begin querying Bottlerocket's update API for updates.
```

README.md
```diff
@@ -20,17 +20,52 @@ kubectl apply -f \
...
+helm install \
+  brupop-crd bottlerocket-update-operator/bottlerocket-shadow \
+  --create-namespace \
+  --version v1.0.0 \
+
+# Install the brupop operator
+helm install \
+  brupop bottlerocket-update-operator/bottlerocket-shadow \
+  --version v1.0.0
+```
```
cbgbt: Why separate the CRD and the operator installations?

deploy/build.rs
```diff
@@ -0,0 +1,31 @@
...
+
+use models::node::combined_crds;
+use std::env;
+use std::fs::File;
+use std::path::PathBuf;
+
+const DEPLOY_DIR: &str = env!("CARGO_MANIFEST_DIR");
+
+fn main() {
+    dotenv::dotenv().ok();
```
cbgbt: Since we're still using dotenv as a development tool, should we leave the `.env` file around with some comments for what it can do?

deploy/charts/bottlerocket-shadow/README.md
```diff
@@ -0,0 +1,27 @@
+# Bottlerocket Shadow
+
+Bottlerocket shadows are the "reflections" of the bottlerocket nodes themselves.
+
+They are used by the Bottlerocket update operator to perform update operations through the
+host's `apiclient update` interface.
+
+### Local chart development
```
cbgbt, line 8: May want to additionally leave a note here about labeling hosts so that they receive a Shadow.

deploy/charts/bottlerocket-shadow/templates/custom-resource-definition.yaml
```diff
@@ -0,0 +1,187 @@
+apiVersion: apiextensions.k8s.io/v1
```
cbgbt, line 1: Possibly I'll answer this myself as I keep reading, but in the pre-helm setup, this was autogenerated based on our Rust models. Do we manually keep the two in sync now, or is it still possible to generate this document?

deploy/charts/bottlerocket-update-operator/README.md
```diff
@@ -0,0 +1,99 @@
+# Bottlerocket Update Operator Helm Chart
+
+This helmchart carries the templates and resource definitions for
+the bottlerocket-update-operator.
+
+It depends on the `bottlerocket-shadow` CRD chart and should be installed
```
cbgbt, line 6: Is it possible to model this dependency using Helm itself?

integ/src/main.rs
```diff
@@ -197,6 +198,10 @@ async fn run() -> Result<()> {
                 .context(error::RunBrupopSnafu)?;
             tokio::time::sleep(WAIT_CERT_MANAGER_COMPLETE).await;
             info!("Running brupop on existing EKS cluster ...");
+            if let Err(error) = process_brupop_namespace(Action::Create, &kube_config_path).await {
+                warn!("Error creating brupop namespace: {:?}", error)
+            };
```
cbgbt: Should we exit here? I could see this being non-fatal if you get this error when such a namespace already exists, but otherwise it seems like we should return an error here.

deploy/charts/bottlerocket-shadow/templates/custom-resource-definition.yaml
cbgbt: Just to flesh this feedback out, a lot of these values only make sense in the context of the Kubernetes YAML model and have no bearing on the `brupop` sources, seems totally fine to remove the Rust models.

What I'm mostly interested in is the auto-generated enums and validations based on our Rust models, which was pretty handy. It would be a lot easier to have a mismatch or mistake if we untie them. Reading more carefully, it seems like the shadow CRD is still autogenerated (probably for this reason!). Nice!

cbgbt:
This is awesome, nice work @jpmcb! 🚀

cbgbt (commented):
Whoops, forgot to hit send on these.

README.md
jpmcb: Good question!

This is a common Helm convention: the CRD chart is separate from the consumer of the CRD. This is because helm will reconcile resources from a operator chart in the most efficient way possible:

> Helm is optimized to load as many resources into Kubernetes as fast as possible. By design, Kubernetes can take an entire set of manifests and bring them all online (this is called the reconciliation loop).

> But there's a difference with CRDs.

> For a CRD, the declaration must be registered before any resources of that CRDs kind(s) can be used. And the registration process sometimes takes a few seconds.

[Helm Docs - Install a CRD Declaration Before Using the Resource](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#install-a-crd-declaration-before-using-the-resource)

There _is an option_ in Helm to let the `helm install` handle this for you. However, this has the follow caveats:

> There is no support at this time for upgrading or deleting CRDs using Helm. This was an explicit decision after much community discussion due to the danger for unintentional data loss. Furthermore, there is currently no community consensus around how to handle CRDs and their lifecycle. As this evolves, Helm will add support for those use cases.

[Helm docs - Some caveats (and explanations)](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations)

Instead, I seperated the two charts to make handling the bottlerocket shadows explicit:
- Cluster operators can can manage the `brs` CRD out of band from the operator itself
  - We avoid the problem of not being able to upgrade/downgrade CRD versions (even though brs v2 is relatively stable)

There is also an argument that has been made in the community that cluster custom resource definitions should only be installable by a cluster admin (where strong RBAC rules are often enforced) but other users may have access to view and edit the `brs` resources and namespace.

In short, this was a deliberate decision to enforce stronger RBAC practices for our users and gets around some issues with versioning CRDs in helm.

deploy/build.rs
jpmcb: Hmmmm that's a good point - I think this means we'll have to yes. Let me circle back and look at that.

deploy/charts/bottlerocket-shadow/templates/custom-resource-definition.yaml
jpmcb: > What I'm mostly interested in is the auto-generated enums and validations based on our Rust models, which was pretty handy. It would be a lot easier to have a mismatch or mistake if we untie them. Reading more carefully, it seems like the shadow CRD is still autogenerated (probably for this reason!). Nice!

Right - I couldn't 100% decouple the kubernetes resources from the rust code. However, I think I struck a balance between what our rust code needs in it (like the enums, etc.) and yaml kubernetes resources

deploy/charts/bottlerocket-shadow/templates/custom-resource-definition.yaml
```diff
@@ -0,0 +1,187 @@
...
+spec:
+  conversion:
+    strategy: Webhook
+    webhook:
+      clientConfig:
+        service:
+          name: brupop-apiserver
+          namespace: brupop-bottlerocket-aws
+          path: /crdconvert
+          port: 443
```
jpmcb: This needs to be templated with the values defined in the `values.yaml` for this chart 👀

deploy/charts/bottlerocket-update-operator/README.md
jpmcb: > Is it possible to model this dependency using Helm itself?

Good question.

I think [this doc from the helm maintainers on the difficulty handling CRDs ](https://github.com/helm/community/blob/f9e06c16d89ccea1bea77c01a6a96ae3b309f823/architecture/crds.md)is really good prior-art to this PR.

Generally, I believe our posture should be to make few assumptions about the shape of our user's clusters and how they are administered. 

I'll annotate [the following section from the doc](https://github.com/helm/community/blob/f9e06c16d89ccea1bea77c01a6a96ae3b309f823/architecture/crds.md#installing-crds) to give some insight into what I believe makes sense for brupop:

> There are a number of well-described issues with installing CRDs. Users must have broad permissions. A CRD must be fully installed before a manifest that references that CRD can be accepted by the Kubernetes API server. (However, a CR can be accepted before there is anything available to handle the CR event.) It is entirely possible for two different applications to require different versions of the same CRD, but no clear way to support or declare that need within Kubernetes. This is exacerbated by the fact that while a CRD is a global object, the controllers that listen for CRD events may be namespace-bound. This means that two instances of the same controller (in different namespaces) can use the same CRD definition. There is no way to query Kubernetes to discover this fact.

Brupop _does_ have namespaced resources (its API server, controller, daemonsets, etc.) that rely on the global `brs` CRD to be present before install. We shouldn't make an assumption on the elevated permission of the users deploying or accessing the brupop namespaced resources (for example, creating namespaces may be a permission many cluster users don't have).

And while I'd be surprised if anyone was attempting to use `brs` as a _dependent_ resource in their own custom kubernetes controller, it is ultimately possible. And I believe the really key part of this is "There is no way to query Kubernetes to discover this fact." We can't make assumptions _from the brupop perspective or install_ on if some other namespaced resources are using the `brs` custom resources. We don't want to couple the install of the CRD and the controller too tightly since we do provide multiple versions of the CRD. Therefore, the CRD should be it's own, separate chart.

> All of these things present challenges for installing CRDs. Essentially, any installation process must have some kind of logic like "If the CRD exists, check to make sure that it is compatible with the version of the CRD that I need, and then proceed on my installation. Otherwise, attempt to install the CRD, and handle the case that I might not have permission to do so."

> After looking at several ways of making it easy for chart developers to do this, we alighted on a fairly easy (though inelegant) solution: Helm, as it stands today, provides a separate location in a chart to add CRDs. These will be uploaded before the chart's templates are even rendered (thus avoiding verification issues with the discovery API). If the CRD exists or if the permissions do not allow it, the chart will fail before the application is deployed.

> Of course, users are unhappy with this for a host of reasons. They want CRDs templated (without understanding the race conditions). They want stronger version controls. They don't like having a separate directory for CRDs. We sympathize... but we currently have not devised a better solution for the aforementioned problems.

And unfortunately, this is exactly where we find ourselves; CRD within a separate chart. Our CRD _does_ need to be templated (in order to derive the namespace and port of the conversion webhook) and we can't rely on pre-rendered templating from the controller chart.

Now to your question:

---

> model this dependency using Helm itself?

Modeling this dependency in Helm would require we make the CRD apart of the update operator's chart itself. This is the last point in the above deposition `Helm provides a separate location in a chart to add CRDs` where a CRD is installed before helm renders and installs the remaining resources.

But, as laid out, I don't believe coupling the CRD and the controller install is the way we want to go.

The alternative is a fast fail when attempting to `helm install` the brupop controller indicating the CRD is not present on the cluster.

jpmcb:
Force pushed to correct for `Cargo.lock` merge conflict.

Going to do another round of iteration here based on comments and a few `TODO` lines left.

integ/src/main.rs
jpmcb: > I could see this being non-fatal if you get this error when such a namespace already exists

Exactly. I ran into a few cases where the integration tests didn't completely remove the namespace (or there were hanging resources preventing the namespace from being purged, my temporary cluster creds timed-out, etc. etc.) and it was just easier to warn instead of completely bail out of creation.

But if we think this is too loose, I can also see exiting early here and requiring the operator to fix the problem.

jpmcb:
Force pushed to add the following:
- Snapshot testing of our generated yaml CRDs. This way, if something changes under the hood in our rust CRD definitions, we can have a signal to accept/reject those changes. This mechanism can also be used as an indicator to update the CRD templates in the helm chart.
  - `cargo test` to run snapshot tests via `insta`. And `cargo insta review` to accept new snapshots.
  - `make check-crd-golden-diff` to compare the rust generated CRD yaml to the helm chart template.
- Remove `const` declaration of namespace. With the helm chart, removes the requirement that users deploy into the `bottlerocket-aws` namespace we had hard coded. Now, we infer the current _in cluster_ namespace from configs.
- Update docs based on feedback

---

Integration tests still looking 👍🏼 

```
❯ k get bottlerocketshadows -A
brupop-bottlerocket-aws   brs-ip-192-168-142-119.us-west-2.compute.internal   Idle    1.11.1    Idle                            0
brupop-bottlerocket-aws   brs-ip-192-168-144-207.us-west-2.compute.internal   Idle    1.11.1    Idle                            0
brupop-bottlerocket-aws   brs-ip-192-168-156-62.us-west-2.compute.internal    Idle    1.11.1    Idle                            0
```

integ/src/main.rs
cbgbt: It might be worth leaving a comment or somehow otherwise making it clear in the code that this is OK in the case that the namespace exists.

deploy/charts/bottlerocket-shadow/templates/custom-resource-definition.yaml
jpmcb: Yes - let's move these togeher

jpmcb:
Force pushed to template a few remaining references to `brupop-bottlerocket-aws` - getting close!!

.github/workflows/deploy-helm-charts.yaml
```diff
@@ -0,0 +1,34 @@
+name: Release Helm Charts
+
+on:
+  release:
+    # Only cut new helm charts when a new release is published
+    types: [published]
```
jpmcb, line 9: Maybe for this PR, we can change this [to include `workflow_dispatch`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch). That way, the maintainers with access rights can test this workflow with `gh` on the command line without having to cut arbitrary releases:

```
gh workflow run deploy-helm-chart.yml
```

cbgbt: I am okay with including this in a separate PR, we just want it before we cut this into a release.

README.md
cbgbt: This makes sense, thanks for elaborating.

agent/src/agentclient.rs
```diff
@@ -92,8 +89,12 @@ impl<T: APIServerClient> BrupopAgent<T> {
         if !associated_bottlerocketshadow.is_empty() {
             Ok(true)
         } else {
+            // Build an incluster config used to infer the current namespace
+            let incluster_config =
+                kube::Config::incluster_dns().context(agentclient_error::K8sConfigCreateSnafu)?;
```
cbgbt: nit: Does it make sense to have a `infer_namespace()` function Rather than repeating this?

apiserver/src/client/webclient.rs
```diff
@@ -45,20 +45,32 @@ pub trait APIServerClient {
     async fn create_bottlerocket_shadow(
         &self,
         req: CreateBottlerocketShadowRequest,
+        namespace: String,
```
cbgbt: These should be `&str` since this function does not need to take ownership of the string/force re-allocation.

Note: You can also do `T: AsRef<str>` if you want to accept both `&str` and `String`, but many APIs just take `&str` unless they need ownership.

apiserver/src/client/webclient.rs
```diff
@@ -45,20 +45,32 @@ pub trait APIServerClient {
...
     ) -> Result<BottlerocketShadow>;
     async fn update_bottlerocket_shadow(
         &self,
         req: UpdateBottlerocketShadowRequest,
+        namespace: String,
```
cbgbt: Rather than add the namespace to the trait, I think we should just take it as an argument on the constructor for `K8SAPIServerClient` and store it -- it's not generically necessary for all implementations.

controller/src/main.rs
```diff
@@ -35,12 +35,15 @@ async fn main() -> Result<()> {
     let incluster_config =
         kube::Config::incluster_dns().context(controller_error::ConfigCreateSnafu)?;
 
+    // Use the incluster config to infer the namespace
+    let namespace = incluster_config.default_namespace.to_string();
+
```
cbgbt, line 44: Maybe throwing this under something like `models/util.rs` or `models/k8s_utils.rs` could help us code share.

deploy/charts/bottlerocket-shadow/templates/custom-resource-definition.yaml
cbgbt: I could be missing where this is already done while writing this comment, so apologies if so, but while it's on my mind: I like the `insta` approach to managing these files. Would be good to document that, possibly in the Development section of the README.

deploy/charts/bottlerocket-update-operator/README.md
cbgbt: Thanks for digging in here. The relationships are a bit awkward, but it does seem like the best setup we can muster using Helm.

deploy/src/test.rs
```diff
@@ -0,0 +1,32 @@
+#[cfg(test)]
+use std::path::PathBuf;
+
+// The following test uses insta snapshots of the
```
cbgbt: Ah, here's the docs I was asking about. Nice. You'll see this when the test fails and know what to do next.

cbgbt: nit: You should be able to move this under `deploy/tests/generated_crds.rs` or something similar. See the Rust docs about [integration tests](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html).

cbgbt (commented):
Looks good! Just a few minor changes that I'd like to see and a few nits that I could take or leave.

agent/src/agentclient.rs
jpmcb: We do this elsewhere where we explicitly create an incluster DNS client config:
https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/f69a1708c3fb095b160665ca162fe4d2f602d6bc/apiserver/src/main.rs#L48-L49

And this is a workaround until we can get the next version of `kube-rs` with the `default_namespace()` function:
https://github.com/kube-rs/kube/pull/1123

jpmcb:
Force pushed to address merge conflicts - addressing comments next!

apiserver/src/client/webclient.rs
jpmcb: > it's not generically necessary for all implementations

Won't all implementations of `update_bottlerocket_shadow` require the namespace since the `bottlerocketshadow` is a namespaced CRD?

We got away with this before because the controller _only_ knew how to operate on `brs` in the hard coded `brupop-bottlerocket-aws` namespace. If a cluster operator tried to deploy `brs` resources to any other namespace, the entire stack breaks.

In my mind, because the CRD is namespaced, `brs` resources are tightly coupled to the namespace they are deployed in and the API should reflect that as well

agent/src/agentclient.rs
jpmcb: Thinking alittle more on this: we'll still need to get an incluster config so I wonder if a subsequent PR (once `default_namespace()` is available in `kube-rs`) would be to refactor these calls into a common shared library

deploy/charts/bottlerocket-shadow/README.md
jpmcb: Done!!

deploy/build.rs
jpmcb: The new workflow should become:

1. Create a `values.yaml` file (maybe we can create a top level one for development purposes and git ignore it)
2. Populate it with necessary values. For example:
```yaml
namespace: "my-custom-ns"
```
3. Pass the values yaml file to be used during install either via:
- `helm install -f values.yaml ...` (to install to a cluster)
- `helm template -f values.yaml ... > file.yaml` (to template into the yaml)

I think there's some better dev UX to tease out here, especially with how the integration tests work into this flow: currently, the integration tests will just read from `bottlerocket-update-operator.yaml` and apply that to the cluster but there's opportunity here to make this more seamless.

integ/src/main.rs
jpmcb: Circling back on this - ended up removing `process_brupop_namespace` since the final generated yaml manifest has the namespace resource 👍🏼

deploy/charts/bottlerocket-shadow/templates/custom-resource-definition.yaml
jpmcb: Done! ☑️

apiserver/src/client/webclient.rs
jpmcb: Done!!! ✅

deploy/src/test.rs
jpmcb: Done!! ✅

apiserver/src/client/webclient.rs
cbgbt: The way I'm thinking about it is that the namespace here identifies the location of the APIServer itself, so it's more of a client configuration detail. It's contrived for sure, but the test implementation is an example of one that doesn't need the configuration, but would now be required to ignore `namespace` or assert that the value corresponds to an environment-derived value.

agent/src/agentclient.rs
cbgbt: I would prefer if we clean this up now given how many times we repeat this pattern. It would make the move to `default_namespace()` a smaller change as well.

jpmcb:
Force pushed to address above comments

apiserver/src/client/webclient.rs
cbgbt: > In my mind, because the CRD is namespaced, brs resources are tightly coupled to the namespace they are deployed in and the API should reflect that as well

Isn't this also reflected if we make it a member of the client constructor? It's not possible (per our rbac permissions) for the API server to modify shadows outside of its namespace, I don't think there's ever a case where you pass a different namespace in this value that succeeds.

agent/src/agentclient.rs
jpmcb: Sounds good!! 😄

apiserver/src/client/webclient.rs
jpmcb: >  the test implementation is an example of one that doesn't need the configuration, but would now be required to ignore namespace or assert that the value corresponds to an environment-derived value.

That's an excellent point!!! Yes, this can be refactored out to the constructor!

jpmcb: Ended up doing abit of a refactor. Now, when the various client configs are created, we pass down the namespace into subsequent constructors that require namespaced resources. It didn't make sense to have to create a new client config with each query to namespaced resources. That would be a costly operation. Instead, we can now access the `namespace` data member where it makes sense.

jpmcb:
Force push to address Sean's comments around infering the namespace throughout. Now, when the various client configs are created, we pass down the namespace into subsequent constructors that require namespaced resources. It didn't make sense to have to create a new client config with each query to namespaced resources. That would be a frequent, costly operation. Instead, we can now access the namespace data member built from `new`.

cbgbt (approved):
Super clean! Nice work John! 🚀 🎉

Makefile
```diff
@@ -74,3 +77,19 @@ dist: brupop-image
...
+	# from the rust definitions. This is useful to ensure there are no hanging changes
+	# that need to be made to the template.
+	diff --color \
+		deploy/charts/bottlerocket-shadow/templates/custom-resource-definition.yaml \
+		deploy/src/golden/custom-resource-definition.yaml || return 0
+
+manifest:
+	echo --- > bottlerocket-update-operator.yaml && \
+	kubectl create namespace brupop-bottlerocket-aws \
+		--dry-run=client \
```
webern, line 98: Does this require a kubeconfig or Kubernetes cluster? If not, this is cool...

integ/src/main.rs
```diff
@@ -217,11 +217,14 @@ async fn run() -> Result<()> {
             )
             .await
             .context(error::LoadKubeConfigSnafu)?;
+
+            // infer the default namespace from the loaded kubeconfig
+            let namespace = config.default_namespace.to_string();
```
webern, line 222: If the user can configure the namespace of their choice, how is it that you get that from the kubeconfig?

models/src/constants.rs
```diff
@@ -1,3 +1,5 @@
+// TODO - need to refactor constants
```
webern: Is this TODO intentional?

deploy/tests/snapshots/generated_crds__generated_crds.snap
```diff
@@ -0,0 +1,192 @@
+---
```
webern: Is this filename as intended? `generated_crds__generated_crds.snap`

webern (approved):
About 1/3 coverage from me. A few questions of ignorance on my part. Looks good.

Makefile
jpmcb: Right, no kubeconfig/kubernetes cluster needed. It just generates the yaml for us. For example, here, I unset my context (so I'm not targeting any cluster) and I can still generate the yaml:

```
❯ k config unset current-context
Property "current-context" unset.

❯ kubectl create namespace test-namespace --dry-run=client -o yaml
apiVersion: v1
kind: Namespace
metadata:
  creationTimestamp: null
  name: test-namespace
spec: {}
status: {}
```

integ/src/main.rs
jpmcb: Good question: the `config.default_namespace` property is the namespace defined by the loaded client config. This is typically the namespace of where the execution is currently happening.

So, in the case of the controller, it reads the in-cluster config and infers that it is in the "bottlerocket-aws" (or whatever the namespace was set during deploy).

This case in the integration tests is abit different in that we aren't inferring a in-cluster client config, we load one explicitly from eksctl. But the principle is the same - it infers the default namespace to deploy some stuff to (and in this case, it is actually "default")

It's abit confusing because the name "default" is overloaded. Most of the time, from the controllers perspective, we are simply attempting to infer what namespace the current context is executing from.

models/src/constants.rs
jpmcb: Probably not - I need to go back again after the cron PR merges to do another round of refactoring / rebase to get those into the helm chart

deploy/tests/snapshots/generated_crds__generated_crds.snap
jpmcb: Yes. `insta` first uses of the name of the Rust file where the test originated from and the name of the test itself. I'll change the name of that file so it's alittle more clear

jpmcb: The name of this file should now be `insta_test__generated_crds.snap` which is the `<test-filename>__<test-name>.snap`

jpmcb:
Force pushed to rename the `insta` Rust test file in order to be abit more clear what's happening. The name of this file should now be `insta_test__generated_crds.snap` which is the `<test-filename>__<test-name>.snap`

FYI - this will also need a hefty rebase and abit of additonal legwork to get in once the cron PR merges

nalshamaajc:
I hope this helm chart gets ready soon! 🙏 😄

lsobowalejc:
Yes! this is definitely a needed feature. Can't wait!

maxres-ch:
Hi there. Just wanted to check if there's any insight on when this might be merged?

jpmcb:
Force pushed with large rebase (which included integrating the work for the scheduler cron) and some docs updates. The biggest thing you'll notice is different is the templates now include the `scheduler_cron_expression` template value.

```diff
diff --git a/bottlerocket-update-operator.yaml b/bottlerocket-update-operator.yaml
index cf982ec..a96c65a 100644
--- a/bottlerocket-update-operator.yaml
+++ b/bottlerocket-update-operator.yaml
@@ -711,10 +711,8 @@ spec:
                   fieldPath: spec.nodeName
             - name: MAX_CONCURRENT_UPDATE
               value: "1"
-            - name: UPDATE_WINDOW_START
-              value: "0:0:0"
-            - name: UPDATE_WINDOW_STOP
-              value: "0:0:0"
+            - name: SCHEDULER_CRON_EXPRESSION
+              value: "* * * * * * *"
           image: public.ecr.aws/bottlerocket/bottlerocket-update-operator:v1.0.0
           name: brupop
       priorityClassName: brupop-controller-high-priority

```

### Testing

1. Deployed new nodegroup to a k8s cluster

2. Installed shadow CRD successfully:
```
$ helm install brupop-crd deploy/charts/bottlerocket-shadow
...

$ kubectl get crds
NAME                                          CREATED AT
bottlerocketshadows.brupop.bottlerocket.aws   2023-06-01T18:24:49Z
...
```

3. Installed the controller stack:
```
$ kubectl create namespace brupop-bottlerocket-aws
$ helm install brupop deploy/charts/bottlerocket-update-operator
...
```

4. Nodes update as expected.

This should be ready to go in now

gthao313:
Follow the instruction above, it works perfectly on my side! nice work!!!

gthao313 (approved):

.github/workflows/deploy-helm-charts.yaml
```diff
@@ -0,0 +1,37 @@
...
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Install Helm
+        uses: azure/setup-helm@v3
+        with:
+          version: v3.10.0
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.4.1
```
stmcginnis, line 33: Is the plan to enable GitHub Pages to host the published charts?

deploy/charts/bottlerocket-update-operator/README.md
```diff
@@ -0,0 +1,138 @@
+# Bottlerocket Update Operator Helm Chart
+
+This helmchart carries the templates and resource definitions for
```
stmcginnis: not: s/helmchart/helm chart/

.github/workflows/deploy-helm-charts.yaml
jpmcb: Exactly!

jpmcb: It'll require some iteration to get it right before running the next release:
- I'm not sure if pages for this repo are turned on. But we can use the `workflow-dispatch` method to launch this workflow out of band of a release.
- We'll need to update release processes to capture updating the helm chart metadata (version etc.) alongside other things (like updating the readme).

But really, in theory, it should _just work_.

stmcginnis (approved):
Overall looks good, and I think it would be great to get this in and available to get more usages/feedback.

stmcginnis (approved):
Thanks! Looks like the checks might just need a re-run to get by a transient error.

jpmcb:
Force pushed to rebase on `develop` which now has fixes for problems caused by #461

stmcginnis (approved):
Let's gooooo!! 🥳

PR #362, opened 2022-12-15
Integration tests: upgrade `aws` dependencies

jpmcb:
**Issue number:**

Closes: #361 
Closes: #360 
Closes: #359
Closes: #358 
Closes: #357

**Description of changes:**

Upgrades our dependency on `aws` dependencies in our integration tests.
- Includes changes to code that expect tuple syntax (instead of the struct)
- Exemption for webpki-root MPL-2.0 license which is pulled in through new versions of aws smithy rust client.

**Testing done:**

Ran through integration tests.

**Terms of contribution:**

### DIFF
--- a/integ/src/nodegroup_provider.rs
+++ b/integ/src/nodegroup_provider.rs
@@ -11,7 +11,7 @@ use std::time::Duration;
 use aws_sdk_eks::model::IpFamily;
 
 use aws_config::meta::region::RegionProviderChain;
-use aws_sdk_ec2::error::{DescribeLaunchTemplatesError, DescribeLaunchTemplatesErrorKind};
+use aws_sdk_ec2::error::DescribeLaunchTemplatesError;
 use aws_sdk_ec2::model::{
     ArchitectureValues, InstanceType, LaunchTemplateTagSpecificationRequest,
     RequestLaunchTemplateData, ResourceType, Tag,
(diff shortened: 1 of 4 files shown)


webern (approved):

jpmcb:
Force pushed to address:

- breaking changes in the way errors work in the AWS rust sdk: https://github.com/awslabs/aws-sdk-rust/issues/657 - Builds should be passing now.

Went through integration tests and all looks 👍🏼 marking ready for review now.

stmcginnis (approved):

integ/src/nodegroup_provider.rs
```diff
@@ -545,26 +545,25 @@ async fn confirm_nodegroup_deleted(
...
     result: Result<GetInstanceProfileOutput, SdkError<GetInstanceProfileError>>,
 ) -> bool {
-    if let Err(SdkError::ServiceError { err, raw: _ }) = result {
-        if matches!(
-            &err.kind,
-            GetInstanceProfileErrorKind::NoSuchEntityException(_)
-        ) {
-            return false;
-        }
+    match result {
```
cbgbt, line 557: Just a tip, no need to actually change the code to this, but matches on `Result` and `Option` types are often simplified or expressed in a more "functional" style by using some of the methods on those types, e.g.

```rust
result
    .map(|_| true)
    .unwrap_or_else(|e| !matches!(
        e.into_service_error(),
        GetInstanceProfileError {
            kind: GetInstanceProfileErrorKind::NoSuchEntityException(_),
            ..
        }
    ))


```

cbgbt (approved):

integ/src/nodegroup_provider.rs
jpmcb: That would work - I was mostly following [the advice laid out in the aws-sdk's guidance on error handling](https://github.com/awslabs/aws-sdk-rust/issues/657) and used a `match`. Thanks for the tip!!

PR #401, opened 2023-01-27
Remove `openssl` in favor of `rustls-tls`

jpmcb:
**Issue number:**

N/a

**Description of changes:**

This PR removes our dependency on OpenSSL in favor of using `rustls-tls`

**Testing done:**

Built and pushed a custom image with these changes. Modified our `bottlerocket-update-operator.yaml` to use that `image`.
Integration tests pass!

Putting this patch up for now to get some early feedback and iteration, but could use additional validation for:
- [x] IPv6 cluster upgrades work
- [x] IPv4 cluster upgrades work

**Terms of contribution:**

### DIFF
--- a/agent/Cargo.toml
+++ b/agent/Cargo.toml
@@ -18,13 +18,12 @@ tracing-opentelemetry = "0.18"
 
 # k8s-openapi must match the version required by kube and enable a k8s version feature
 k8s-openapi = { version = "0.17.0", default-features = false, features = ["v1_20"] }
-kube = { version = "0.78.0", default-features = true, features = [ "derive", "runtime", "rustls-tls" ] }
+kube = { version = "0.78.0", default-features = false, features = [ "derive", "runtime", "rustls-tls" ] }
 
 semver = { version = "1.0", features = [ "serde" ] }
 serde = { version = "1", features = [ "derive" ] }
 serde_json = "1"
 snafu = "0.7"
 tokio = { version = "1", features = ["macros", "rt-multi-thread", "time"] }
-reqwest = { version = "0.11", features =  [ "json" ] }
 chrono = { version = "0.4.23", features = [ "serde" ] }
 tokio-retry = "0.3"
--- a/apiserver/src/api/mod.rs
+++ b/apiserver/src/api/mod.rs
@@ -34,10 +34,14 @@ use kube::{
     runtime::{reflector, watcher::watcher, WatchStreamExt},
     ResourceExt,
 };
-use openssl::ssl::{SslAcceptor, SslFiletype, SslMethod};
 use opentelemetry::global::meter;
+use rustls::{
+    server::AllowAnyAnonymousOrAuthenticatedClient, Certificate, PrivateKey, RootCertStore,
+    ServerConfig,
+};
+use rustls_pemfile::{certs, pkcs8_private_keys};
 use snafu::{OptionExt, ResultExt};
-use std::env;
+use std::{env, fs::File, io::BufReader};
 use tokio::time::{sleep, Duration};
 use tracing::{event, Level};
 use tracing_actix_web::TracingLogger;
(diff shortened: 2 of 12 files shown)


jpmcb:
I know what you're thinking.

_"But John, I thought [`rustls-tls` in `kube` didn't work with IPv6 clusters](https://github.com/kube-rs/kube/issues/153) since [webpki can't resolve hostnames via IP addresses](https://github.com/briansmith/webpki/issues/54)?"_

Thankfully, in https://github.com/bottlerocket-os/bottlerocket-update-operator/commit/d2b5d788228dbd3913dce285cc866d0637f49b10 we switched to using the suggested work-around of reaching the kubernetes API server in cluster through kubernetes DNS. We now build our k8s clients by defining that they should use the `incluster_dns` in their config:

```rust
    let incluster_config =
        kube::Config::incluster_dns().context(controller_error::ConfigCreateSnafu)?;

    let k8s_client = kube::client::Client::try_from(incluster_config)
        .context(controller_error::ClientCreateSnafu)?;
```

So nothing right now is necessarily preventing us from using the `rustls-tls` feature in kube.

And it also looks like [the next version of `rustls-tls` will have a fix for this upstream](https://github.com/rustls/rustls/commit/d312391239aa15eaceaf8d32ed9997cb747e54d8) (which was fixed in their fork of webpki). This should flow downstream to `kube` and we can get the benefit of this if we need to create outside cluster clients in the future. Exciting stuff!

webern (approved):

jpmcb:
Testing in both IPv4 and IPv6 clusters looks good! Marking ready for review

agent/Cargo.toml
```diff
@@ -18,13 +18,13 @@ tracing-opentelemetry = "0.18"
...
 k8s-openapi = { version = "0.17.0", default-features = false, features = ["v1_20"] }
-kube = { version = "0.78.0", default-features = true, features = [ "derive", "runtime", "rustls-tls" ] }
+kube = { version = "0.78.0", default-features = false, features = [ "derive", "runtime", "rustls-tls" ] }
 
 semver = { version = "1.0", features = [ "serde" ] }
 serde = { version = "1", features = [ "derive" ] }
 serde_json = "1"
 snafu = "0.7"
 tokio = { version = "1", features = ["macros", "rt-multi-thread", "time"] }
-reqwest = { version = "0.11", features =  [ "json" ] }
```
cbgbt, line 28: I think we have to enable feature `rustls-tls` here? Or are we just using `reqwest` to talk to the local Bottlerocket API socket?

apiserver/src/api/mod.rs
```diff
@@ -158,23 +162,66 @@ pub async fn run_server<T: 'static + BottlerocketShadowClient>(
...
-    builder
-        .set_private_key_file(
-            format!("{}/{}", TLS_KEY_MOUNT_PATH, PRIVATE_KEY_NAME),
-            SslFiletype::PEM,
-        )
-        .context(error::SSLSnafu)?;
+    // Server public certificate file
+    let cert_file_path = format!("{}/{}", TLS_KEY_MOUNT_PATH, PUBLIC_KEY_NAME);
+    let cert_file = &mut BufReader::new(
+        //File::open(cert_file_path).context(error::FileOpenSnafu { path: cert_file_path })?,
```
cbgbt: nit: remove comment

apiserver/src/api/mod.rs
```diff
@@ -158,23 +162,66 @@ pub async fn run_server<T: 'static + BottlerocketShadowClient>(
...
+        .context(error::CertExtractSnafu {
+            path: ca_file_path.to_string(),
+        })?
+        .into_iter()
+        .map(Certificate)
+        .collect();
+
+    let mut cert_store = RootCertStore::empty();
+    for ca in cas {
+        _ = cert_store.add(&ca).context(error::CertStoreSnafu);
```
cbgbt: Instead of tossing the return value here, should we be using `?`?

agent/Cargo.toml
jpmcb: Hmmm good question.

I originally removed `default-features = false` to ensure that the `default-tls` feature (enabled by default) in this package wouldn't try to pull in openssl dependencies.

But now that I look at it, _I don't see `reqwest` used at all in the `agent`_.

```
❯ pwd
/home/ubuntu/workspace/bottlerocket-os/bottlerocket-update-operator/agent
❯ rg "reqwest"
Cargo.toml
28:#reqwest = { version = "0.11", default-features = false, features =  [ "json" ] }
```

> Or are we just using reqwest to talk to the local Bottlerocket API socket?

It looks like in https://github.com/bottlerocket-os/bottlerocket-update-operator/commit/0ea5f104ab459b7611a364b85a8ae23ad70cc481 the manual `reqwest` builder was abstracted into the `apiserver`'s client.

So I believe this line can be outright removed.

apiserver/src/api/mod.rs
jpmcb: Whoops! Good catch!

jpmcb:
Force pushed to address comments above and remove unused `reqwest` dependency in `agent`. Running through another quick integration test validation ✅ 

Further validation:
```
❯ rg -i openssl
CHANGELOG.md
20:* Upgraded Bottlerocket SDK to consume fix for OpenSSl CVE-2022-3602 and CVE-2022-3786 ([#331])

deny.toml
20:    "OpenSSL",
58:expression = "MIT AND ISC AND OpenSSL"

Cargo.lock
2187:name = "openssl-probe"
2633: "openssl-probe",
```

Looks like `openssl-probe` is a dependency of `rustls-native-certs` in the `rustls` dependency chain 👍🏼 no other cases of openssl in the cargo toml

jpmcb:
Force pushed to rebase on `develop` which had removed `webpki-roots` in `deny.toml`. Added that exception back since the rustls dependency chain uses it.

---

Edit: integration test passed with the latest force push. All nodes upgrade from 1.9.0 to the latest bottlerocket

```
❯ k get brs -A
brupop-bottlerocket-aws   brs-ip-192-168-130-3.us-west-2.compute.internal     Idle    1.12.0    Idle                            0
brupop-bottlerocket-aws   brs-ip-192-168-134-58.us-west-2.compute.internal    Idle    1.12.0    Idle                            0
brupop-bottlerocket-aws   brs-ip-192-168-154-241.us-west-2.compute.internal   Idle    1.12.0    Idle                            0
```

deny.toml
```diff
@@ -23,6 +23,8 @@ allow = [
 ]
 
 exceptions = [
+    # Explicitly allows MPL-2 being pulled in through reqwest's and actix's rustls dependency chain (which uses webpki)
+    { name = "webpki-roots", allow = ["MPL-2.0"], version = "*" },
```
stmcginnis, line 27: Sorry about the extra work. :/

stmcginnis (approved):
Looks good to me and all tests are passing.

deny.toml
jpmcb: Oh no worries! We had no way of knowing this was going to end up being a dependency again since it was removed by the AWS smithy libraries 😄

agent/Cargo.toml
webern: Nice, there's a cargo plugin `cargo unused` that can sometimes help find these.

cbgbt (approved):
🚀

PR #402, opened 2023-01-27
Remove webpki-root from license check deny.yaml

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This crate is a transitive dependency, but it is not directly included in our Cargo.toml, resulting in cargo-deny complaining about an "unmatched license exception" in deny.yaml.

```txt
warning[L005]: license exception was not encountered
   ┌─ /src/deny.toml:27:14
   │
27 │     { name = "webpki-roots", allow = ["MPL-2.0"], version = "*" },
   │              ^^^^^^^^^^^^^^ unmatched license exception
```

**Testing done:**

Ran `make check-licenses` and verified the warning goes away. Performed `make build` to verify there were no other side effects.

**Terms of contribution:**

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -23,8 +23,6 @@ allow = [
 ]
 
 exceptions = [
-    # Explicitly allows MPL-2 being pulled in through newer versions of awslabs/smithy-rs (which uses webpki)
-    { name = "webpki-roots", allow = ["MPL-2.0"], version = "*" },
     { name = "unicode-ident", version = "1.0.2", allow = ["MIT", "Apache-2.0", "Unicode-DFS-2016"] },
 ]
 


jpmcb (approved):
Good catch! 

> This crate is a transitive dependency, but it is not directly included in our Cargo.toml

Looks like [aws-smithy actually _removed_ the `webpki-roots` feature](https://github.com/awslabs/smithy-rs/releases/tag/release-2023-01-11) from their `rustls-tls` dependency last week when we updated it in the integration tests here: https://github.com/bottlerocket-os/bottlerocket-update-operator/commit/6890462c99ff06c8300353bead537427cf0e7e0d

I'm surprised we didn't catch this earlier, but looks like we don't fail when running the `make check-license` target: 
```
❯ make
cargo fetch --locked
docker run --rm \
        --network none \
        --user "1000:1000" \
        --security-opt label=disable \
        --env CARGO_HOME="/src/.cargo" \
        --volume "/home/ubuntu/workspace/bottlerocket-os/bottlerocket-update-operator/:/src" \
        --workdir "/src/" \
        "public.ecr.aws/bottlerocket/bottlerocket-sdk-x86_64:v0.28.0" \
        bash -c "cargo deny --all-features check --disable-fetch licenses bans sources"
warning[L005]: license exception was not encountered
   ┌─ /src/deny.toml:27:14
   │
27 │     { name = "webpki-roots", allow = ["MPL-2.0"], version = "*" },
   │              ^^^^^^^^^^^^^^ unmatched license exception

bans ok, licenses ok, sources ok
cargo fmt -- --check
cargo build --locked
```

We'll want to update that make target to also bail out if there are license problems.

webern (approved):

cbgbt (approved):

stmcginnis:
> We'll want to update that make target to also bail out if there are license problems.

I took a quick look at cargo-deny configuration to see if there is a way we could trigger this. It [doesn't appear there is a setting to treat these as errors](https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html). Since cargo-deny doesn't have an exit code for this, we can't get any info out of the `docker run` call either.

We could try capturing and parsing the stdout from the call to look for the warning message, but that feels a little fragile and probably not worth the extra work.

Another option could be to file an enhancement request upstream to have something like a `--fail-on-warning` flag added, or a `[license.exceptions.error-on-unmatched]` configuration.

Not sure how critical this is though. If we have something in there that isn't matched, there really is no harm done. It won't lead to us missing any license violations. At worst, it's just a case like this where we start to get some warning output and someone just needs to do a little clean up to tidy up our CI output.

But let me know if you feel more strongly about this - happy to follow up with an upstream enhancement request if you think we really need this.

PR #403, opened 2023-01-27
Update deprecated method calls

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

Our builds have been emitting a few warnings for use of deprecated calls. This updates these instances to use their replacement calls.

**Testing done:**

For the change in `base64::encode`, [tested with Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7720ee1d0651009436de97923d55ddfc) to verify the output matches with the new call.

There is a loss of millisecond precision with the `Utc.with_ymd_and_hms()`, but verified with `cargo test` that we weren't expecting that for the `fake_shadow` results.

**Terms of contribution:**

### DIFF
--- a/integ/src/nodegroup_provider.rs
+++ b/integ/src/nodegroup_provider.rs
@@ -3,6 +3,8 @@
   Meanwhile, terminating all created ec2 instances when integration test is running 'clean' subcommand.
 !*/
 
+use base64::engine::general_purpose::STANDARD as base64_engine;
+use base64::Engine as _;
 use serde::{Deserialize, Serialize};
 use std::fmt::Debug;
 use std::process::Command;
@@ -465,7 +467,7 @@ fn userdata(
         _ => return Err(ProviderError::new_with_context("Invalid dns ip")),
     };
 
-    Ok(base64::encode(format!(
+    Ok(base64_engine.encode(format!(
         r#"[settings.updates]
         ignore-waves = true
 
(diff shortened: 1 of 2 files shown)


jpmcb (approved):

webern (approved):

integ/src/nodegroup_provider.rs
```diff
@@ -465,7 +466,7 @@ fn userdata(
         _ => return Err(ProviderError::new_with_context("Invalid dns ip")),
     };
 
-    Ok(base64::encode(format!(
+    Ok(general_purpose::STANDARD.encode(format!(
```
cbgbt: nit: It would be nice to `use base64::engine::general_purpose::STANDARD as base64_` or similar to clarify this line.

cbgbt (approved):

integ/src/nodegroup_provider.rs
stmcginnis: Yeah, good call. I don't love the API on this, so at least aliasing that would make it a little more clear what is being done there.

Tested the change with the Rust playground again to make sure there are no differences.

Since there are a few approvals here, going to merge once tests pass to avoid taking everyone's time on something minor. But if anyone has any concerns about this last update, please let me know and I will be happy to revert or update to address those concerns.

PR #418, opened 2023-02-22
Makefile: refactors `brupop-image` target

jpmcb:
Ensures that building the brupop image works without needing to have the rust and cargo toolchain installed. This is useful for release and CI/CD where an environment may not have these tools installed.

**Issue number:**

N/a - related to useful changes for releasing brupop

**Description of changes:**

```
Ensures that building the brupop image works without needing to have the
rust and cargo toolchain installed. This is useful for release and CI/CD
where an environment may not have these tools installed.

```

See comments below for further explanation 

**Testing done:**

Able to run:
- `make brupop-image` without cargo/rust toolchain installed
- `make`, the default target, which builds the image but also checks licenses, fetches sources locally, etc
- `make build` which fetches sources, checks licenses, but builds binaries locally (not in the docker environment or as part of an image)

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,13 @@ USER root
 # Required to build in --offline mode
 ENV CARGO_HOME=/src/.cargo
 
+# Add brupop source
 ADD ./ /src/
+
+# Ensure cargo dependencies are fetched and available in the Docker context
+RUN cargo fetch --locked --manifest-path /src/Cargo.toml
+
+# Builds brupop binaries
 RUN cargo install --offline --locked --target ${UNAME_ARCH}-bottlerocket-linux-musl --path /src/agent --root /src/agent && \
     cargo install --offline --locked --target ${UNAME_ARCH}-bottlerocket-linux-musl --path /src/apiserver --root /src/apiserver && \
     cargo install --offline --locked --target ${UNAME_ARCH}-bottlerocket-linux-musl --path /src/controller --root /src/controller
--- a/Makefile
+++ b/Makefile
@@ -36,12 +36,13 @@ BUILDER_IMAGE = public.ecr.aws/bottlerocket/bottlerocket-sdk-$(BOTTLEROCKET_SDK_
 
 export CARGO_HOME = $(TOP)/.cargo
 
-image: build brupop-image
+image: check-licenses brupop-image
 
 # Fetches crates from upstream
 fetch:
 	cargo fetch --locked
 
+# Checks allowed/denied upstream licenses against the deny.toml
 check-licenses: fetch
 	docker run --rm \
 		--network none \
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -7,7 +7,13 @@ USER root
 # Required to build in --offline mode
 ENV CARGO_HOME=/src/.cargo
 
+# Add brupop source
 ADD ./ /src/
+
+# Ensure cargo dependencies are fetched and available in the Docker context
+RUN cd /src/ && cargo fetch --locked
```
jpmcb, line 14: Fetches sources in docker image to ensure we have necessary upstream code to build locally. This is still dependent on the Cargo.lock, so nothing new should be pulled in here.

Makefile
```diff
@@ -36,12 +36,13 @@ BUILDER_IMAGE = public.ecr.aws/bottlerocket/bottlerocket-sdk-$(BOTTLEROCKET_SDK_
 
 export CARGO_HOME = $(TOP)/.cargo
 
-image: build brupop-image
+image: check-licenses brupop-image
```
jpmcb, line 39: Removes the duplicate build step here. No reason to build binaries locally when building the image does the same thing. Calls into the `check-license` path to ensure licenses are good. This is the default target.

Makefile
```diff
@@ -53,21 +54,22 @@ check-licenses: fetch
 		"$(BUILDER_IMAGE)" \
 		bash -c "cargo deny --all-features check --disable-fetch licenses bans sources"
 
-# Builds, Lints and Tests the Rust workspace
-build: fetch check-licenses
+# Builds, Lints, and Tests the Rust workspace locally
+build: check-licenses
```
jpmcb, line 58: `check-licenses` already does fetch. So removed this duplicate work.

Makefile
```diff
@@ -53,21 +54,22 @@ check-licenses: fetch
...
 	cargo fmt -- --check
-	cargo build --locked
 	cargo test --locked
+	cargo build --locked
```
jpmcb, line 61: Moved building to run _after_ tests have passed.

Makefile
```diff
@@ -53,21 +54,22 @@ check-licenses: fetch
...
 
-brupop-image: fetch check-licenses
+# Builds only the brupop image. Useful target for CI/CD, releasing, etc.
+brupop-image:
```
jpmcb, line 64: `brupop-image` no longer has a dependency on fetch or check-licenses since both depend on the cargo / rust toolchain being available. This ensures that we can build the brupop image in an environment that may not have those tools.

Makefile
```diff
@@ -53,21 +54,22 @@ check-licenses: fetch
...
 	docker build $(DOCKER_BUILD_FLAGS) \
 		--build-arg UNAME_ARCH="$(UNAME_ARCH)" \
 		--build-arg BUILDER_IMAGE="$(BUILDER_IMAGE)" \
 		--tag "$(IMAGE_NAME)" \
 		--network none \
 		-f Dockerfile .
 
-dist: brupop-image
+dist: check-licenses brupop-image
```
jpmcb, line 71: This target should utilize the check-license / fetch path since it's creating a distribution (and we should ensure we have checked the licenses locally)

Dockerfile
```diff
@@ -7,7 +7,13 @@ USER root
 # Required to build in --offline mode
 ENV CARGO_HOME=/src/.cargo
 
+# Add brupop source
 ADD ./ /src/
+
+# Ensure cargo dependencies are fetched and available in the Docker context
+RUN cargo fetch --locked --manifest-path /src/Cargo.toml
```
jpmcb, line 14: Fetches sources in docker image to ensure we have necessary upstream code to build locally. This is still dependent on the Cargo.lock, so nothing new should be pulled in here. Note: this requires network access

Makefile
```diff
@@ -53,21 +54,21 @@ check-licenses: fetch
...
+	cargo build --locked
 
-brupop-image: fetch check-licenses
+# Builds only the brupop image. Useful target for CI/CD, releasing, etc.
+brupop-image:
 	docker build $(DOCKER_BUILD_FLAGS) \
 		--build-arg UNAME_ARCH="$(UNAME_ARCH)" \
 		--build-arg BUILDER_IMAGE="$(BUILDER_IMAGE)" \
 		--tag "$(IMAGE_NAME)" \
-		--network none \
```
jpmcb, line 67: Note: in order to fetch cargo sources from crates.io upstream _in the image_, we need to allow network access. But since this using the lockfile, this should be ok since typically, on the local machine, the sources would be fetched and mounted into the container.

jpculp (approved):

cbgbt (approved):

gthao313 (approved):

Issue #428, opened 2023-03-09
Error when `SCHEDULER_CRON_EXPRESSION` is set without `UPDATE_WINDOW_START` & `UPDATE_WINDOW_STOP`


prashant-prodigal:
We are installing bottlerocket update operator in an EKS with no internet access. But the operator deployment starts failing, its giving this error:

  2023-03-09T13:12:28.369373Z  INFO actix_server::builder: starting 2 workers
    at /src/.cargo/registry/src/github.com-1ecc6299db9ec823/actix-server-2.2.0/src/builder.rs:200

  2023-03-09T13:12:28.369436Z ERROR controller: controller exited
    at controller/src/main.rs:110

I am using Latest image as per the docs. Could you pls point me to what could be wrong here?

jpmcb:
Hi @prashant-prodigal - the error you are seeing at `controller/src/main.rs:110` is related to the controller's metric server attempting to bind to the local loopback network and start serving metrics.

> We are installing bottlerocket update operator in an EKS with no internet access.

Note that the Bottlerocket update operator requires network access to `updates.bottlerocket.aws`: this is how update operator system queries for new OS updates. Read more about it here: https://github.com/bottlerocket-os/bottlerocket-update-operator#why-are-my-bottlerocket-nodes-egressing-to-httpsupdatesbottlerocketaws

Does your node have some network attached? In order for the prometheus server to come up, it'll at least need to be able to bind on `0.0.0.0` for IPv4 clusters or `[::]` for IPv6 clusters.

Can you provide the full logs from the failed controller deployment?

```
 kubectl logs -n brupop-bottlerocket-aws pod/brupop-controller-deployment-{YOUR-DEPLOYMENT}
```

prashant-prodigal:
Hello, I have allowed the URL https://updates.bottlerocket.aws still we are getting below error from command
 kubectl logs -n brupop-bottlerocket-aws pod/brupop-controller-deployment-{YOUR-DEPLOYMENT}

  2023-03-10T04:36:28.369259Z  INFO actix_server::builder: starting 2 workers
    at /src/.cargo/registry/src/github.com-1ecc6299db9ec823/actix-server-2.2.0/src/builder.rs:200

  2023-03-10T04:36:28.369368Z ERROR controller: controller exited
    at controller/src/main.rs:110

jpmcb:
What's the shape of your network? Are there any other logs in from the other update operator components?

[...]

PR #453, opened 2023-05-16
Give cron expressions precedence over time windows when both are present

stmcginnis:
**Issue number:**

Related: #428

**Description of changes:**

This adds a minor behavior change. In the existing `from_environment` call we would error out if both the new cron based expressions and the old time window settings were both provided. Rather than erroring out here, it's probably a better user experience to have a warning, but use the non-deprecated setting to run.

This also updates the documentation to focus more on cron settings and to make an explicit statement of what the expected behavior will be if both the new and old are present.

**Testing done:**

Stepped through new unit tests to make sure values are parsed correctly and the expected logic path is taken.

This testing doesn't explain [the behavior users have seen](https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/428#issuecomment-1546446212) with actually needing both set, but it may make it more obvious what has happened to help gather more data on that.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,33 +145,6 @@ For example:
               value: "1"
 ```
 
-#### Set an Update Time Window - DEPRECATED
-
-Note!! that these settings are deprecated and will be removed in a future release.
-You should use the Scheduler settings below.
-If you still decide to use these settings, please use "hour:00:00" format only instead of "HH:MM:SS".
-
-`UPDATE_WINDOW_START` and `UPDATE_WINDOW_STOP` can be used to specify the time window in which updates are permitted.
-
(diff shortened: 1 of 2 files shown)


stmcginnis:
Timing/concurrency issue with the way the new unit tests work. Update coming.

jpmcb (approved):
Would also be good to get @gthao313 to check this out - thanks Sean!

gthao313:
Hi I think the errors that customers faced when using `cron scheduler`  might not related to this PR. I have replied on the issue https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/428#issuecomment-1550368459.

Ideally, when we ship `cron scheduler` on next release, We should highly recommend users to use `cron expression` instead of `time window`. Therefore, after discussing with @cbgbt, we think this should be error out if users provide `cron expression` and `time window`.

stmcginnis:
Yeah, this is definitely not what folks are running in to with #428. This was just an improvement I noticed while taking a look at that.

gthao313 (approved):

Issue #461, opened 2023-06-06
Don't consume newer version of Rust via `rustup update stable`


jpmcb:
In the following GitHub workflow for CI builds:

https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/047357262b02b462007451922844fb604c1566d4/.github/workflows/ci.yml#L30

we are consuming the newest version of rust "locally" which pulls down dependencies that get fed into the docker container that uses the bottlerocket sdk. This leads to the possibility of the "local" version of rust fast-forwarding from the version in the sdk. For example, at the time of this writing, the sdk is on 1.69, and in these breaking CI builds, rustup consumes 1.70:

https://github.com/bottlerocket-os/bottlerocket-update-operator/actions/runs/5189657965/jobs/9355968232?pr=350

We may be able to forgo doing the whole operation of getting rust "locally" and instead remove the `--network=none` line in the docker container so it can just do it all from within the build.

PR #462, opened 2023-06-06
Cargo: use env vars when calling cargo

jpmcb:
**Issue number:**

#461 - stopgap

**Description of changes:**

This is a stopgap for a breaking change when consuming 1.70 of cargo/rust. This ensures that _both_ the newer version of rust and the older version in the SDK are using the "sparse" protocol for pulling dependencies from crates.io

**Testing done:**

```
❯ make check-licenses
CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse cargo fetch --locked
docker run --rm \
        --network none \
        --user "1000:1000" \
        --security-opt label=disable \
        --env CARGO_HOME="/src/.cargo" \
        --volume "/home/ubuntu/workspace/bottlerocket-os/bottlerocket-update-operator/:/src" \
        --workdir "/src/" \
        "public.ecr.aws/bottlerocket/bottlerocket-sdk-x86_64:v0.32.0" \
        bash -c "CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse cargo deny --all-features check --disable-fetch licenses bans sources"
bans ok, licenses ok, sources ok

❯ cargo --version
cargo 1.70.0 (ec8a8a0ca 2023-04-25)
```

**Terms of contribution:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,13 +34,14 @@ BOTTLEROCKET_SDK_ARCH = $(UNAME_ARCH)
 
 BUILDER_IMAGE = public.ecr.aws/bottlerocket/bottlerocket-sdk-$(BOTTLEROCKET_SDK_ARCH):$(BOTTLEROCKET_SDK_VERSION)
 
+export CARGO_ENV_VARS = CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
 export CARGO_HOME = $(TOP)/.cargo
 
 image: check-licenses brupop-image
 
 # Fetches crates from upstream
 fetch:
-	cargo fetch --locked
+	$(CARGO_ENV_VARS) cargo fetch --locked
 
 # Checks allowed/denied upstream licenses against the deny.toml
 check-licenses: fetch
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -34,13 +34,14 @@ BOTTLEROCKET_SDK_ARCH = $(UNAME_ARCH)
 
 BUILDER_IMAGE = public.ecr.aws/bottlerocket/bottlerocket-sdk-$(BOTTLEROCKET_SDK_ARCH):$(BOTTLEROCKET_SDK_VERSION)
 
+export CARGO_ENV_VARS = CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
```
stmcginnis, line 37: Is there a reason not to just `export CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse`?

stmcginnis (approved):
This looks fine, especially as a stopgap for the current blocking issue. Question inline, but really just curiosity.

Makefile
etungsten: I think this is due to how env passing in Makefile works. Makefile tasks don't behave like a bash script.

etungsten (approved):

Makefile
stmcginnis: Ah, that makes sense!

Issue #469, opened 2023-06-18
Chart information is no up to date


evgeni-roif-gong:
Trying to deploy using the helm chart is not working.
```
➜  test-brupop git:(main) ✗ helm repo add brupop https://bottlerocket-os.github.io/bottlerocket-update-operator
Error: looks like "https://bottlerocket-os.github.io/bottlerocket-update-operator" is not a valid chart repository or cannot be reached: failed to fetch https://bottlerocket-os.github.io/bottlerocket-update-operator/index.yaml : 404 Not Found
```

In addition, going to the deploy/chart folder looking for the current chart shows the image of 1.0.0 instead of 1.1.0 like the current release.

Is chart based deployment still supported or using the manifest is preferred?

stmcginnis:
Hi @evgeni-roif-gong - helm support is something just being added now. I think the initial support was merged just in the last week or so. There has not been a release since then, but it will be updated when it is ready. Thanks!

jpmcb:
FYI - in the meantime, if you'd like to be _on the cutting edge_, you can clone this repo and install the helm chart manually from the chart templates directly.

```bash
# Get into the directory where you've cloned the repo
$ cd brupop-clone

# Install the bottlerocket shadow CRD from the templates directly
$ helm install brupop-crd deploy/charts/bottlerocket-shadow

# Create the namespace
$ kubectl create namespace brupop-bottlerocket-aws

# Install the operator from the templates directly
$ helm install brupop deploy/charts/bottlerocket-update-operator
```

takhat:
@jpmcb We tried this and still weren't able to deploy the BottleRocket operator. The pod seems to be restarting.
Fyi - we were able to install BottleRocket Shadow and BottleRocket API with above Helm commands but we are getting this below error for BottleRocket operator pod:
```
ERROR controller: controller exited
    at controller/src/main.rs:94 
```
Can you please suggest what is wrong and how to fix it? Thanks for your help!

gthao313:
Hi @takhat @evgeni-roif-gong Sorry for the late response. Currently we have already released new version [1.2.0 ](https://github.com/bottlerocket-os/bottlerocket-update-operator/releases/tag/v1.2.0).  You're able to use helm chart at this new version. Feel free to let me if you have question! thanks for using bottlerocket-update-operator!   : )

gthao313:
the new version with helm chart had been release and Chart information is up to date

takhat:
[...]

PR #473, opened 2023-06-27
Add UPDATE_WINDOW_START & UPDATE_WINDOW_STOP to controller-deploy template

andrea-defraia:
Add UPDATE_WINDOW_START & UPDATE_WINDOW_STOP to controller-deployment template

**Issue number:**
#428
#469

**Description of changes:**
Add missing environment variables to the controller deployment template

**Testing done:**
Deployed the helm chart without and with the changes. Controller works only after the changes are added.
Everything is working afterwards.

**Terms of contribution:**

### DIFF
--- a/deploy/charts/bottlerocket-update-operator/templates/controller-deployment.yaml
+++ b/deploy/charts/bottlerocket-update-operator/templates/controller-deployment.yaml
@@ -48,6 +48,7 @@ spec:
               value: "{{ .Values.max_concurrent_updates }}"
             - name: SCHEDULER_CRON_EXPRESSION
               value: "{{ .Values.scheduler_cron_expression}}"
+              # UPDATE_WINDOW_STOP & UPDATE_WINDOW_START are being deprecated
             - name: UPDATE_WINDOW_STOP
               value: "{{ .Values.update_window_stop}}"
             - name: UPDATE_WINDOW_START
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
The time window settings are being deprecated, but it's probably OK to add them. It does raise the question of the upgrade path by adding deprecated settings that will need to be removed, but that might not be too bad.

It might be useful to add a comment above these settings calling out that they really shouldn't be used with the new release.

andrea-defraia:
I've added the comment in the controller template, but are these 2 values being deprecated only here or as well in the agent?

Also, while I understand the reasoning, I think we could leave it as it is, and once they are deprecated, delete them on a new PR.

stmcginnis (approved):
Thanks!

gthao313 (approved):
Thanks for opening the PR! I'm not sure about if we should add this to the template since I remember the discussion with @cbgbt was we wanted to remove it. But I can understand currently it causes mismatch on the new yaml file and brupop 1.1.0 version.

andrea-defraia:
Hello, any update on this?
I see it's approved but not merged, and I can't understand if you'll fix current version or we must wait for next release. 
Please give an ETA on merge or new release, thanks!

> Thanks for opening the PR! I'm not sure about if we should add this to the template since I remember the discussion with @cbgbt was we wanted to remove it. But I can understand currently it causes mismatch on the new yaml file and brupop 1.1.0 version.

gthao313:
Hi @andrea-defraia sorry for late reply! we have already released new version [1.2.0](https://github.com/bottlerocket-os/bottlerocket-update-operator/releases/tag/v1.2.0) few days ago. Therefore, I think we might not need merge this PR. thanks! :)

PR #496, opened 2023-08-05
rate-limit brupop agent calls to the Bottlerocket update API

cbgbt:
**Issue number:**
#483

**Description of changes:**
**This PR is probably best reviewed commit-by-commit. It appears large, but most of the bigger changes are just simple refactors.**

There were reports in #483 of Brupop being unable to update hosts, entering the agent's crash loop. Sometimes the updates would end up succeeding, but sometimes they would become stuck.

Upon inspection of the brupop logs and systemd journal, it seemed like even under conditions where brupop succeeded its updates, it would often be stalled by the apiserver due to the updates lock being held. Brupop would apparently starve itself of this lock by refreshing updates too often and calling the API too quickly.

This change adds ratelimiting to the update-related API calls. It also refactors a few portions of the agent.

**Testing done:**
* Successfully watched brupop update a cluster of Bottlerocket nodes. The agent never once encountered an incident in which the update lock was held.

**Terms of contribution:**


agent/src/agentclient.rs
```diff
@@ -81,51 +81,51 @@ impl<T: APIServerClient> BrupopAgent<T> {
...
-                    }
-                    // Any other type of errors can not present that associated BottlerocketShadow doesn't exist, need return error
-                    _ => {
-                        return Err(e).context(
-                            agentclient_error::UnableFetchBottlerocketShadowSnafu {
-                                node_name: &self.associated_bottlerocketshadow_name.clone(),
-                            },
-                        );
-                    }
+        match bottlerocket_shadows
```
gthao313, line 92: Nice！

gthao313 (approved):

agent/src/apiclient.rs
```diff
@@ -225,10 +235,16 @@ pub(super) mod api {
...
     /// invoke `/bin apiclient` to interact with the Bottlerocket Update API.
     /// This function helps to invoke apiclient raw command.
-    pub(super) async fn invoke_apiclient(args: Vec<String>) -> Result<Output> {
+    pub(super) async fn invoke_apiclient(
+        args: Vec<String>,
+        rate_limiter: Option<&SimpleRateLimiter>,
+    ) -> Result<Output> {
         let mut attempts: i8 = 0;
         // Retry up to 5 times in case the Update API is busy; Waiting 10 seconds between each attempt.
         while attempts < MAX_ATTEMPTS {
```
ecpullen: Is there a reason not to use `Retry::spawn`?

ecpullen (approved):

agent/src/apiclient.rs
cbgbt: I suppose not. Looking at this, I also wonder if we should be retrying any error with the API (e.g. what if the apiserver is being restarted).

cbgbt:
* Added a commit to use tokio-retry in `agent::apiclient`
* Rebased on develop

ecpullen (approved):

PR #497, opened 2023-08-05
Use smaller hosts for github runners

cbgbt:
**Terms of contribution:**


stmcginnis (approved):
Seems like 8 should be large enough. Worth trying and seeing how it goes.

markusboehme (approved):

gthao313 (approved):

Issue #478, opened 2023-07-06
Missing BottlerocketShadow CRD causes excessive logging


maxres-ch:
https://github.com/bottlerocket-os/bottlerocket-update-operator/issues?q=is%3Aissue+backoff

**Image I'm using:**
v1.0.0

**Issue or Feature Request:**

We've started using the helm chart on the develop branch. We missed that the shadow chart had to be installed.  When the operator  started it rightly wrote error messages because the API didn't exist. We didn't see the error because the controller went healthy.

We saw around 53.6 million error messages in a 7 minute window and a peak of 130K messages / sec when the bottlerocket shadow CRD was not installed. 

I would have expected the controller to not be healthy --or--  there's exponential backoffs used when these kind of errors are hit (could potentially DoS the kube-api server too). Controllers I've written in the past used backoff to prevent crazy logs and hammering the API. I saw that at least one of the errors was had a static backoff of 5 secs, but there's some others that seem to have no backoff. 

- Please search for similar issues, including closed issues: 
  - https://github.com/bottlerocket-os/bottlerocket-update-operator/issues?q=is%3Aissue+is%3Aopen++exception.details%3A+BottlerocketShadowCreate
  - https://github.com/bottlerocket-os/bottlerocket-update-operator/issues?q=is%3Aissue+is%3Aopen+%22%27ApiError%3A+%22%22404+page+not+found+%22
  - https://github.com/bottlerocket-os/bottlerocket-update-operator/issues?q=is%3Aissue+backoff
- Please include details about the environment you're running in.
  - EKS, 1.27, bottlerocket managed nodes. Benchmarked settings for CIS Level 1
  - I can go through to the account manager if more details are need.
- Please include any error messages you received, with any required context.
  - This is the pattern of the messages, it looks like the output from the controller has ascii color? I think this should probably be good enough to track down the spots:
 ```
  Message
[[wildcard]*[/wildcard]:[wildcard][00-59][/wildcard]:[wildcard][00.018549-59.992987][/wildcard]Z[0m [33m WARN[0m [1;33mkube_client::client[0m[33m: [33mUnsuccessful data error parse: 404 page not found
[0m
[2;3mat[0m /src/.cargo/registry/src/github.com-1ecc6299db9ec823/kube-client-0.76.0/src/client/mod.rs:434
[2;3min[0m [wildcard]*[/wildcard]::[wildcard]*[/wildcard]::[[wildcard]*[/wildcard][0m
[...]

PR #506, opened 2023-08-08
Implement backoff for k8s watcher retries

cbgbt:
**Issue number:**
#478

**Description of changes:**
The `kube::runtime` provides us utilities for watching for events from Kubernetes `watch` APIs and updating an internal cache of state, which is much more efficient than polling the k8s API for changes.

If the brupop CRD is not installed, these watchers end up in an extremely tight retry loop to begin watching.

This PR:
* Updates `kube` to get access to some nice default backoff functionality that was added in a recent release.
* Uses aforementioned backoff functionality
* Unrelatedly, lets the `controller` sleep for a longer fixed interval if there is nothing to do (which is true if the CRD is not installed)

**Testing done:**
Deployed the changes to a cluster without the CRD installed.

Prior to this change, the agent logs were, several times per second, spammed with messages like this:
```
2023-08-08T20:23:28.561457Z  WARN kube_client::client: Unsuccessful data error parse: 404 page not found

    at /src/.cargo/registry/src/index.crates.io-6f17d22bba15001f/kube-client-0.85.0/src/client/mod.rs:445
```

After this change, these messages are clearly backed-off exponentially, with a short initial burst followed by longer waits of 15+ seconds.

The controller also emits this log every 10 seconds:
```
2023-08-08T20:24:59.897841Z  INFO controller::controller: Nothing to do: The bottlerocket-update-operator is not aware of any BottlerocketShadow objects. Is the bottlerocket-shadow CRD installed? Are nodes labelled so that the agent is deployed to them? See the project's README for more information.
    at controller/src/controller.rs:285
```

**Terms of contribution:**

### DIFF
--- a/controller/src/controller.rs
+++ b/controller/src/controller.rs
@@ -25,6 +25,9 @@ use tracing::{event, instrument, Level};
 // Defines the length time after which the controller will take actions.
 const ACTION_INTERVAL: Duration = Duration::from_secs(2);
 
+// The interval between control loop polls if no nodes are detected.
+const CANNOT_FIND_ANY_NODES_WAIT_INTERVAL: Duration = Duration::from_secs(10);
+
 // Defines environment variable name used to fetch max concurrent update number.
 const MAX_CONCURRENT_UPDATE_ENV_VAR: &str = "MAX_CONCURRENT_UPDATE";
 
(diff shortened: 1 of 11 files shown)


gthao313 (approved):

controller/src/controller.rs
```diff
@@ -278,6 +281,17 @@ impl<T: BottlerocketShadowClient> BrupopController<T> {
         // On every iteration of the event loop, we reconstruct the state of the controller and determine its
         // next actions. This is to ensure that the operator would behave consistently even if suddenly restarted.
         loop {
+            if self.all_brss().is_empty() {
+                event!(
+                    Level::INFO,
+                    "Nothing to do: The bottlerocket-update-operator is not aware of any BottlerocketShadow objects. \
```
stmcginnis, line 287: Nice!

stmcginnis (approved):

PR #511, opened 2023-08-09
controller: disable prometheus metric processor memory

### LINKED ISSUE #441: Incorrect Prometheus metric values
**Image I'm using:**
1.1.0

**Issue:**
Scraped Prometheus metrics don't seem to accurately reflect the existing state of all the nodes. It seems like the gauges store old historical data. For example, if a node is updated from Bottlerocket 1.13.0 to 1.13.2, the `brupop_hosts_version` gauge will have a value of the following, reflecting both the initial and final state. This is assuming there is only one node. A restart of the Deployment sets the gauges to the correct/expected values.

```
# HELP brupop_hosts_version Brupop host's bottlerocket version
# TYPE brupop_hosts_version gauge
brupop_hosts_version{bottlerocket_version="1.13.0",service_name="unknown_service"} 1
brupop_hosts_version{bottlerocket_version="1.13.2",service_name="unknown_service"} 1
```

Expected:

```
# HELP brupop_hosts_version Brupop host's bottlerocket version
# TYPE brupop_hosts_version gauge
brupop_hosts_version{bottlerocket_version="1.13.2",service_name="unknown_service"} 1
```

I believe the issue has been there from the initial implementation as the initial comment in the discussion for pull request 132 (https://github.com/bottlerocket-os/bottlerocket-update-operator/pull/132) seems to illustrate the same issue (5 node cluster, but the sum of the values from brupop_hosts_version is 6).

A similar issue seems to exist for the `brupop_hosts_state` gauge as well.

If it was a counter, this could make sense, but a gauge should represent the current value for the metric (for each set of labels specified).

cbgbt:
**Issue number:**
Closes #441

**Description of changes:**
Per #441 and testing, the gauges reported by the controller's `/metrics` endpoint would often report stale data, double-counting hosts as it showed the historic state alongside the current state.

This is because [`with_memory()`](https://docs.rs/opentelemetry_sdk/0.18.0/src/opentelemetry_sdk/metrics/processors/basic.rs.html#41) was enabled, which explicitly maintains previous metrics when only new subsets are reported.

This change disables metric processor memory to make it truly stateless, which is how we were computing the metrics anyways.

In this case, prometheus will still report old state until the metrics are stale. In order to attempt to clear old state faster, we also explicitly report `0` counts for prior metrics when we can. Because the controller is supposed to be stateless and can be rescheduled, this is done as a best-effort.

**Testing done:**
Watched the controller `/metrics` endpoint with and without this change.

After the change was made, the metrics endpoint always surfaced the current state of the cluster, e.g.:

```
# HELP brupop_hosts_state Brupop host's state
# TYPE brupop_hosts_state gauge
brupop_hosts_state{service_name="unknown_service",state="Idle"} 2
brupop_hosts_state{service_name="unknown_service",state="RebootedIntoUpdate"} 1
brupop_hosts_state{service_name="unknown_service",state="StagedAndPerformedUpdate"} 0
# HELP brupop_hosts_version Brupop host's bottlerocket version
# TYPE brupop_hosts_version gauge
brupop_hosts_version{bottlerocket_version="1.14.0",service_name="unknown_service"} 2
brupop_hosts_version{bottlerocket_version="1.14.2",service_name="unknown_service"} 1
```

(Note that `StagedAndPerformedUpdate` has a count of `0`, because the host that is now in `RebootedIntoUpdate` had it's prior count intentionally wiped)

**Terms of contribution:**

### DIFF
--- a/controller/Cargo.toml
+++ b/controller/Cargo.toml
@@ -29,3 +29,6 @@ snafu = "0.7"
 tokio = { version = "1", features = ["macros", "rt-multi-thread", "time"] }
 tracing = "0.1"
 validator = { version = "0.16", features = ["derive"] }
+
+[dev-dependencies]
+maplit = "1"
--- a/controller/src/controller.rs
+++ b/controller/src/controller.rs
@@ -1,6 +1,7 @@
+use crate::metrics;
+
 use super::{
-    metrics::{BrupopControllerMetrics, BrupopHostsData},
-    scheduler::BrupopCronScheduler,
+    metrics::BrupopControllerMetrics, scheduler::BrupopCronScheduler,
     statemachine::determine_next_node_spec,
 };
 use models::constants::{BRUPOP_INTERFACE_VERSION, LABEL_BRUPOP_INTERFACE_NAME};
@@ -16,7 +17,7 @@ use kube::Api;
 use kube::ResourceExt;
 use opentelemetry::global;
 use snafu::ResultExt;
-use std::collections::{BTreeMap, HashMap};
+use std::collections::BTreeMap;
 use std::env;
 use tokio::time::{sleep, Duration};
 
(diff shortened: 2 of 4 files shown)


cbgbt:
^ Fixed a typo in the git commit message, plus rebased onto develop.

ecpullen (approved):

gthao313 (approved):

PR #514, opened 2023-08-10
Automate helm chart releases

### LINKED ISSUE #492: Automate the chart release process
Currently we manually update the charts when we release new version brupop. Ideally we should integrate the release process to workflow and the github action will be triggered to release the new charts on each brupop release.

EKS team has a good case on this and we study from it https://github.com/aws/eks-charts/blob/master/scripts/publish-charts.sh

cbgbt:
**Issue number:**
Closes #492

**Description of changes:**
Our current github workflow for publishing charts would create a GitHub release for each new chart, but what we want is for the Releases page to focus on the operator, and for new charts to be available in the helm repository to reflect those.

The [eks-charts](https://github.com/aws/eks-charts) repo does something like this, so this commit uses a modified version of their workflow to automatically publish charts.

**Testing done:**
* I performed [several releases](https://github.com/cbgbt/bottlerocket-update-operator/releases) in my fork of brupop
* Noted that new charts were added to a PR against `gh-pages`: https://github.com/cbgbt/bottlerocket-update-operator/pull/4
* I added my github.io page as a helm repository and used it to install brupop, and then updated a cluster with that installation
* Sample [Actions workflow](https://github.com/cbgbt/bottlerocket-update-operator/actions/runs/5835899909)

**Limitations**
We have to manage our external dependencies on Helm and Kubeconform in the Makefile. Ideally we move these dependencies to the Bottlerocket SDK.

**Terms of contribution:**

### DIFF
--- a/scripts/install-toolchain.sh
+++ b/scripts/install-toolchain.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+PLATFORM=$(uname -s | tr '[:upper:]' '[:lower:]')
+ARCH=$(uname -m | sed 's/aarch64/arm64/' | sed 's/x86_64/amd64/')
+
+mkdir -p "${CHART_TMP_DIR}"
+
+## Install kubeconform
+mkdir -p "${CHART_TMP_DIR}/kubeconform"
+curl -sSL "https://github.com/yannh/kubeconform/releases/download/${KUBECONFORM_VERSION}/kubeconform-${PLATFORM}-${ARCH}.tar.gz" | tar xz -C "${CHART_TMP_DIR}/kubeconform"
+mv "${CHART_TMP_DIR}/kubeconform/kubeconform" "${CHART_TOOLS_DIR}/kubeconform"
+
+## Install helm v3
+mkdir -p "${CHART_TMP_DIR}/helmv3"
+curl -sSL "https://get.helm.sh/helm-${HELMV3_VERSION}-${PLATFORM}-${ARCH}.tar.gz" | tar xz -C "${CHART_TMP_DIR}/helmv3"
+mv "${CHART_TMP_DIR}/helmv3/${PLATFORM}-${ARCH}/helm" "${CHART_TOOLS_DIR}/helmv3"
+
+rm -rf "${CHART_TMP_DIR}"
--- a/scripts/lint-charts.sh
+++ b/scripts/lint-charts.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+FAILED_V3=()
+
+cd "${CHARTS_DIR}"
+for d in */; do
+    echo "Linting chart ${d} w/ helm v3"
+    helmv3 lint "${CHARTS_DIR}/${d}" || FAILED_V3+=("${d}")
+done
+
+if [[ "${#FAILED_V3[@]}" -eq 0 ]]; then
+    echo "All charts passed linting!"
+    exit 0
+else
+    echo "Helm v3:"
+    for chart in "${FAILED_V3[@]}"; do
+        printf "%40s ❌\n" "$chart"
+    done
+    exit 1
+fi
--- a/scripts/publish-charts.sh
+++ b/scripts/publish-charts.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+VERSION="$(git describe --tags --always)"
+
+if [[ "${VERSION}" =~ ^v[0-9]+(\.[0-9]+){2}$ ]]; then
+    git fetch --all
+    git config user.name "${GITHUB_ACTOR}"
+    git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+    git remote set-url origin "https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPO}"
+    git config pull.rebase false
+    git checkout gh-pages
+    git checkout -b "gh-pages-release-${VERSION}"
+    mv -n "${CHART_BUILD_DIR}"/charts/*.tgz .
+    helmv3 repo index . --url "https://${GITHUB_REPO_OWNER}.github.io/${GITHUB_REPO_NAME}"
+    git add ./*.tgz index.yaml
+    git commit -m "Publish helm charts for ${VERSION}"
+
+    git push origin "gh-pages-release-${VERSION}"
+    gh pr create -B gh-pages -H "gh-pages-release-${VERSION}" \
+        --title "Publish helm charts for ${VERSION}" \
+        --body "This publishes charts for the Bottlerocket Update Operator version ${VERSION} to the helm repository."
+else
+    echo "Not a valid semver release tag! Skip charts publish"
+    exit 1
+fi
--- a/scripts/validate-chart-versions.sh
+++ b/scripts/validate-chart-versions.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+EXIT_CODE=0
+
+# Determine the commit hash and tag of the previous release:
+#
+#     * If the repository head is at a release tag (e.g. v1.2.3-rc4), select
+#       the release before that (e.g. v1.2.3-rc3)
+#     * If the repository head has advanced from a release tag (e.g. v1.0-g0123abcd),
+#       select the release the current head is based on (e.g. v1.0)
+if $(git describe HEAD --tags | grep -Eq "^v[0-9]+(\.[0-9]+)*(-[a-z0-9]+)?$"); then
+    LAST_RELEASE_HASH=$(git rev-list --tags --max-count=1 --skip=1 --no-walk)
+else
+    TAG=$(git describe HEAD --tags | grep -Eo "^v[0-9]+(\.[0-9]+)*")
+    LAST_RELEASE_HASH=$(git rev-list -1 "${TAG}")
+fi
+LAST_RELEASE_TAG=$(git describe "${LAST_RELEASE_HASH}" --tags)
+
+cd "${CHARTS_DIR}"
+echo "📝 Checking for updated Chart versions since the last brupop release ${LAST_RELEASE_TAG}"
+for d in */; do
+    LAST_COMMIT_HASH=$(git --no-pager log --pretty=tformat:"%H" -- "${d}" | awk 'FNR <= 1')
+    if [[ -z "${LAST_COMMIT_HASH}" ]]; then
+        echo "✅ Chart ${d} has not been commited and appears to be in development"
+        continue
+    fi
+
+    ## If LAST_RELEASE_HASH does not include the chart, then it's a new chart and does not need a version increment
+    if [[ -z $(git ls-tree -d "${LAST_RELEASE_HASH}" "${d}") ]]; then
+        echo "✅ Chart ${d} is a new chart since the last release"
+        continue
+    fi
+
+    ## If LAST_RELEASE_HASH is NOT an ancestor of LAST_COMMIT_HASH then it has not been modified
+    if [[ -z $(git rev-list "${LAST_COMMIT_HASH}" | grep "${LAST_RELEASE_HASH}") || "${LAST_COMMIT_HASH}" == "${LAST_RELEASE_HASH}" ]]; then
+        echo "✅ Chart ${d} had no changes since the last eks-charts release"
+        continue
+    fi
+
+    LAST_RELEASE_CHART_VERSION=$(git --no-pager show "${LAST_RELEASE_HASH}":deploy/charts/"${d}"Chart.yaml | awk '$1 == "version:" { print $2 }' | tr -d '[:space:]')
+    LAST_COMMIT_CHART_VERSION=$(git --no-pager show "${LAST_COMMIT_HASH}":deploy/charts/"${d}"Chart.yaml | awk '$1 == "version:" { print $2 }' | tr -d '[:space:]')
+    if [[ "${LAST_RELEASE_CHART_VERSION}" == "${LAST_COMMIT_CHART_VERSION}" ]]; then
+        echo "❌ Chart ${d} has the same Chart version as the last release ${LAST_COMMIT_CHART_VERSION}"
+        EXIT_CODE=1
+    else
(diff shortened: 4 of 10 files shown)


gthao313 (approved):

scripts/install-toolchain.sh
```diff
@@ -0,0 +1,24 @@
...
+## Install kubeconform
+mkdir -p "${TMP_DIR}/kubeconform"
+curl -sSL https://github.com/yannh/kubeconform/releases/download/${KUBECONFORM_VERSION}/kubeconform-${PLATFORM}-${ARCH}.tar.gz | tar xz -C "${TMP_DIR}/kubeconform"
+mv "${TMP_DIR}/kubeconform/kubeconform" "${TOOLS_DIR}/kubeconform"
+
+## Install helm v3
+mkdir -p "${TMP_DIR}/helmv3"
+curl -sSL https://get.helm.sh/helm-${HELMV3_VERSION}-${PLATFORM}-${ARCH}.tar.gz | tar xz -C "${TMP_DIR}/helmv3"
+mv "${TMP_DIR}/helmv3/${PLATFORM}-${ARCH}/helm" "${TOOLS_DIR}/helmv3"
+rm -rf "${PLATFORM}-${ARCH}"
```
markusboehme: Nit: That one isn't created or extracted in the working directory, or am I missing it? Everything seems to happen in `${TMP_DIR}`.

scripts/validate-chart-versions.sh
```diff
@@ -0,0 +1,41 @@
...
+echo "📝 Checking for updated Chart versions since the last brupop release $LAST_RELEASE_TAG"
+for d in */; do
+    LAST_COMMIT_HASH=$(git --no-pager log --pretty=tformat:"%H" -- $d | awk 'FNR <= 1')
+    ## If LAST_RELEASE_HASH does not include the chart, then it's a new chart and does not need a version increment
+    if [[ -z $(git ls-tree -d $LAST_RELEASE_HASH $d) ]]; then
+    echo "✅ Chart $d is a new chart since the last release"
+    continue
+    fi
+    ## If LAST_RELEASE_HASH is NOT an ancestor of LAST_COMMIT_HASH then it has not been modified 
+    if [[ ! -z $LAST_COMMIT_HASH && -z $(git rev-list $LAST_COMMIT_HASH | grep $LAST_RELEASE_HASH) || $LAST_COMMIT_HASH == $LAST_RELEASE_HASH ]]; then
```
markusboehme: `LAST_COMMIT_HASH` should only be able to be empty if the script is running in some developer's local clone, right? Leaving this in for dev and testing purposes?

scripts/validate-chart-versions.sh
```diff
@@ -0,0 +1,41 @@
...
+if $(git describe HEAD --tags | grep -Eq "^v[0-9]+(\.[0-9]+)*(-[a-z0-9]+)?$"); then
+    LAST_RELEASE_HASH=$(git rev-list --tags --max-count=1 --skip=1 --no-walk)
+else 
+    TAG=$(git describe HEAD --tags | grep -Eo "^v[0-9]+(\.[0-9]+)*")
+    LAST_RELEASE_HASH=$(git rev-list -1 $TAG)
+fi
+LAST_RELEASE_TAG=$(git describe $LAST_RELEASE_HASH --tags)
+cd ${CHARTS}
+echo "📝 Checking for updated Chart versions since the last brupop release $LAST_RELEASE_TAG"
+for d in */; do
```
markusboehme: Nit: The loop body is all indented at the same level even though there's more conditionals inside.

scripts/validate-chart-versions.sh
```diff
@@ -0,0 +1,41 @@
...
+export PATH="${TOOLS_DIR}:${PATH}"
+
+EXIT_CODE=0
+if $(git describe HEAD --tags | grep -Eq "^v[0-9]+(\.[0-9]+)*(-[a-z0-9]+)?$"); then
+    LAST_RELEASE_HASH=$(git rev-list --tags --max-count=1 --skip=1 --no-walk)
+else 
+    TAG=$(git describe HEAD --tags | grep -Eo "^v[0-9]+(\.[0-9]+)*")
+    LAST_RELEASE_HASH=$(git rev-list -1 $TAG)
+fi
+LAST_RELEASE_TAG=$(git describe $LAST_RELEASE_HASH --tags)
```
markusboehme: I would find a comment on this block helpful since it took me a bit to grok. Writing down my understanding, please correct me or feel free to copy it in:

```
# Determine the commit hash and tag of the previous release:
#
#     * If the repository head is at a release tag (e.g. v1.2.3-rc4), select
#       the release before that (e.g. v1.2.3-rc3).
#     * If the repository head has advanced from a release tag (e.g. v1.0-g0123abcd),
#       select the release the current head is based on (e.g. v1.0).
```

scripts/validate-chart-versions.sh
```diff
@@ -0,0 +1,41 @@
...
+    echo "✅ Chart $d is a new chart since the last release"
+    continue
+    fi
+    ## If LAST_RELEASE_HASH is NOT an ancestor of LAST_COMMIT_HASH then it has not been modified 
+    if [[ ! -z $LAST_COMMIT_HASH && -z $(git rev-list $LAST_COMMIT_HASH | grep $LAST_RELEASE_HASH) || $LAST_COMMIT_HASH == $LAST_RELEASE_HASH ]]; then
+    echo "✅ Chart $d had no changes since the last eks-charts release"
+    continue
+    fi
+    LAST_RELEASE_CHART_VERSION=$(git --no-pager show $LAST_RELEASE_HASH:deploy/charts/"$d"Chart.yaml | grep 'version:' | xargs | cut -d' ' -f2 | tr -d '[:space:]')
+    LAST_COMMIT_CHART_VERSION=$(git --no-pager show $LAST_COMMIT_HASH:deploy/charts/"$d"Chart.yaml | grep 'version:' | xargs | cut -d' ' -f2 | tr -d '[:space:]')
```
markusboehme: Nit: Would `LAST_RELEASE_CHART_VERSION=$(git show ... | awk '$1 == "version:" { print $2 })` do the trick? I believe it to be equivalent, and also clearer on what is being done.

scripts/validate-charts.sh
```diff
@@ -0,0 +1,30 @@
...
+done
+
+if [[ "${#FAILED_V3[@]}" -eq 0 ]]; then
+    echo "All charts passed validations!"
+    exit 0
+else
+    echo "Helm v3:"
+    for chart in "${FAILED_V3[@]}"; do
+        printf "%40s ❌\n" "$chart"
+    done
```
markusboehme, line 23: Should we exit non-zero here after listing all problems?

scripts/lint-charts.sh
```diff
@@ -0,0 +1,30 @@
...
+done
+
+if [[ "${#FAILED_V3[@]}" -eq 0 ]]; then
+    echo "All charts passed linting!"
+    exit 0
+else
+    echo "Helm v3:"
+    for chart in "${FAILED_V3[@]}"; do
+        printf "%40s ❌\n" "$chart"
+    done
```
markusboehme, line 19: Should the script exit non-zero here in case of failure?

scripts/publish-charts.sh
```diff
@@ -0,0 +1,29 @@
...
+if echo "${VERSION}" | grep -Eq "^v[0-9]+(\.[0-9]+){2}$"; then
+    git fetch --all
+    git config user.name "$GITHUB_ACTOR"
+    git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+    git remote set-url origin https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPO}
+    git config pull.rebase false
+    git checkout gh-pages
+    mv -n $PACKAGE_DIR/charts/*.tgz .
+    helmv3 repo index . --url https://${GITHUB_REPO_OWNER}.github.io/${GITHUB_REPO_NAME}
+    git add .
```
markusboehme: Nit: `git add ./*.tgz` as a safety net?

scripts/install-toolchain.sh
```diff
@@ -0,0 +1,24 @@
...
+set -euo pipefail
+
+PLATFORM=$(uname -s | tr '[:upper:]' '[:lower:]')
+ARCH=$([[ $(uname -m) = "x86_64" ]] && echo 'amd64' || echo 'arm64')
+GIT_REPO_ROOT=$(git rev-parse --show-toplevel)
+BUILD_DIR="${GIT_REPO_ROOT}/chartbuild"
+TMP_DIR="${BUILD_DIR}/tmp"
+TOOLS_DIR="${BUILD_DIR}/tools"
+mkdir -p "${TOOLS_DIR}"
+export PATH="${TOOLS_DIR}:${PATH}"
```
markusboehme: Those seem to be used by all scripts. Since these being the same is not just incidental but part of the interface, it could make sense to export them from the Makefile instead.

markusboehme:
Will be cool to have this publishing step automatically triggered! :magic_wand: Happy to approve once it's clear whether the validation scripts are intentionally not exiting with a failure code.

scripts/install-toolchain.sh
cbgbt: No, you're right. I think we can just remove this line and let the `helm` unpack dir be removed when we delete `${TMP_DIR}`

scripts/validate-chart-versions.sh
cbgbt: That's right, it would require an uncommitted chart.

I think this is probably more relevant for EKS who maintain many charts, but I don't necessarily see the harm in leaving it here.

cbgbt:
Addressed feedback from @markusboehme. Thanks!

scripts/install-toolchain.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+PLATFORM=$(uname -s | tr '[:upper:]' '[:lower:]')
+ARCH=$([[ $(uname -m) = "x86_64" ]] && echo 'amd64' || echo 'arm64')
```
jpculp:
```suggestion
ARCH=$(uname -m| sed 's/aarch64/arm64/' | sed 's/x86_64/amd64/')
```
If you wanted one less sub-shell and to avoid assuming non-amd64 = arm64.

scripts/publish-charts.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+VERSION="$(git describe --tags --always)"
+
+if echo "${VERSION}" | grep -Eq "^v[0-9]+(\.[0-9]+){2}$"; then
```
jpculp:
```suggestion
if [[ "${VERSION}" =~ ^v[0-9]+(\.[0-9]+){2}$ ]]; then
```

scripts/install-toolchain.sh
```diff
@@ -0,0 +1,19 @@
...
+mkdir -p "${CHART_TMP_DIR}/kubeconform"
+curl -sSL https://github.com/yannh/kubeconform/releases/download/${KUBECONFORM_VERSION}/kubeconform-${PLATFORM}-${ARCH}.tar.gz | tar xz -C "${CHART_TMP_DIR}/kubeconform"
+mv "${CHART_TMP_DIR}/kubeconform/kubeconform" "${CHART_TOOLS_DIR}/kubeconform"
+
+## Install helm v3
+mkdir -p "${CHART_TMP_DIR}/helmv3"
+curl -sSL https://get.helm.sh/helm-${HELMV3_VERSION}-${PLATFORM}-${ARCH}.tar.gz | tar xz -C "${CHART_TMP_DIR}/helmv3"
+mv "${CHART_TMP_DIR}/helmv3/${PLATFORM}-${ARCH}/helm" "${CHART_TOOLS_DIR}/helmv3"
+
+rm -rf ${CHART_TMP_DIR}
```
jpculp:
```suggestion
rm -rf "${CHART_TMP_DIR}"
```
There are a couple places in this script and others where you might want to quote to prevent globbing/splitting.

scripts/validate-chart-versions.sh
```diff
@@ -0,0 +1,50 @@
...
+
+EXIT_CODE=0
+
+# Determine the commit hash and tag of the previous release:
+#
+#     * If the repository head is at a release tag (e.g. v1.2.3-rc4), select
+#       the release before that (e.g. v1.2.3-rc3)
+#     * If the repository head has advanced from a release tag (e.g. v1.0-g0123abcd),
+#       select the release the current head is based on (e.g. v1.0)
+if $(git describe HEAD --tags | grep -Eq "^v[0-9]+(\.[0-9]+)*(-[a-z0-9]+)?$"); then
```
jpculp, line 12: This could also be done with `[[ $() =~ ]]` if you wanted, but both ways work.

cbgbt:
I didn't see @jpculp's feedback yet, I'll work on that. The above modifies the publication step to use a pull request so that we can enable branch protection on the `gh-pages` branch.

Here's a sample PR: https://github.com/cbgbt/bottlerocket-update-operator/pull/1

cbgbt:
^ Responds to some feedback from @jpculp , and also fixes a few bugs I discovered while reading through the action runner logs.

Here's the most recent actions run:
https://github.com/cbgbt/bottlerocket-update-operator/actions/runs/5835899909

With this PR being created:
https://github.com/cbgbt/bottlerocket-update-operator/pull/4

scripts/publish-charts.sh
cbgbt: I also explicitly added the chart index.

scripts/validate-chart-versions.sh
cbgbt: I reworked the function to make this case more clear and exit early.

scripts/install-toolchain.sh
cbgbt: Should all be fixed!

jpculp (approved):

scripts/validate-chart-versions.sh
markusboehme: Oh, yes, you mentioned it being based on the EKS scripts, but I actually reviewed from scratch and didn't look at the originals. No harm in leaving it in at all, just been wondering to check my understanding.

markusboehme (approved):
:roller_skate:

PR #517, opened 2023-08-10
Prepare for v1.3.0 Release

cbgbt:
**Terms of contribution:**


stmcginnis (approved):
Looks like everything is covered.

cbgbt:
I realized I still needed to update the YAML version of the manifest.

stmcginnis (approved):
Good catch!

Just noting a thought - I wonder if there is any way we can script this. There are several spots to update now, and it seems more likely that we could miss one. Something to ponder for the future.

cbgbt:
Agreed, this is too many things to update for each release. I was going to add CI to perform checks on push that these files are synced with eachother, but that is a problem because folks looking at the Brupop repo for tips on configuration may end up finding unreleased features.

README.md
```diff
@@ -85,7 +85,7 @@ You can use a values file when installing brupop with helm (via the `--values /
 namespace: "brupop-bottlerocket-aws"
 
 # The image to use for brupop
-image: "public.ecr.aws/bottlerocket/bottlerocket-update-operator:v1.2.0"
+image: "public.ecr.aws/bottlerocket/bottlerocket-update-operator:v1.3.0"
```
gthao313, line 88: Nit, I used to have seceprate PR on this change and shipped it after we had new version container out. It definitely not a blocker. : )

gthao313 (approved):

PR #524, opened 2023-08-16
Improve cluster-wide update performance, as well as various small fixes

cbgbt:
**Description of changes:**
Addresses various problems found while testing larger clusters for the 1.3.0 release:

* In rare cases, we could still starve the brupop agent of access to the udpate API. This adds exponential backoff instead of fixed interval retries.
* `structopt` is unmaintained and has a dependency on `atty`, which has [a published vulnerability](https://github.com/advisories/GHSA-g98v-hv3f-hcfr), so we moved to `argh`
* The integration tests can time out too early, so the timeouts were bumped.
* The rate limits for Brupop's `apiserver` and Bottlerocket's `apiclient` were too restrictive, and caused the update process to take several minutes longer to complete. This adds some of the telemetry used to figure out ideal values, as well as the changes to go faster.
* Bottlerocket nodes check the TUF repo for updates with `refresh-updates` *far* too often. It was not a reasonable default, and with rate-limiting implemented we end up always out of burst tokens. This refrains from calling `refresh-updates` not but every 2 minutes.

**Testing done:**
* Ran the `integ` tool with a cluster of size 12, monitoring the logs

**Terms of contribution:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -248,6 +248,7 @@ dependencies = [
  "kube",
  "lazy_static",
  "models",
+ "nonzero_ext",
  "semver",
  "serde",
  "serde_json",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -326,15 +326,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
- "winapi",
-]
-
-[[package]]
 name = "apiserver"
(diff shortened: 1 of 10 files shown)


cbgbt:
^ Fixed a bug where the agent would overwrite its "target" version with it' current version unless it happened to hit its 2-minute update check time interval.

etungsten (approved):

Cargo.lock
```diff
@@ -973,21 +984,6 @@ dependencies = [
  "winapi",
 ]
 
-[[package]]
-name = "clap"
```
jpculp, line 977: Nice! You'll probably want to update the **deny.toml** to no longer skip the `clap` tree _(and `structopt-derive`!)_

cbgbt:
^ Addresses feedback from @jpculp

jpculp (approved):

Issue #538, opened 2023-09-05
It is unclear whether the operator works


jRilkef:
**Image I'm using:**
328549459982.dkr.ecr.eu-central-1.amazonaws.com/bottlerocket-update-operator:v1.3.0 (via Helm)

**Issue or Feature Request:**
I can't seem to understand whether the operator functions as intended. See the logs:
<details>
<summary>agent logs</summary>

```
  2023-09-05T09:45:29.558438Z  INFO agent::agentclient: Checking for Bottlerocket updates.
    at agent/src/agentclient.rs:209
    in agent::agentclient::shadow_status_with_refreshed_system_matadata with shadow_error_info: ShadowErrorInfo { crash_count: 0, state_transition_failure_timestamp: None }
[...]

PR #539, opened 2023-09-05
agent: clarify log messages about update lock

cbgbt:
When the update lock is held, we now emit DEBUG-level messages unless we exhaust all retries, in which case an ERROR-level message is emitted.

As a note for reviewers, `#[tracing::instrument(err)]` means that if a function returns a `Result` type, `tracing` will automatically write `ERROR`-level log messages if an error is returned. This guarantees that we get the ERROR-level log message if we exhaust our retries.

**Issue number:**
#538

**Terms of contribution:**


stmcginnis (approved):
lgtm

webern (approved):

PR #559, opened 2023-10-07
feat: add servicemonitor for prometheus

### LINKED ISSUE #558: Add Prometheus ServiceMonitor to the operator helm chart
**Feature Request:**
Add an optional Prometheus serviceMonitor/podMonitor to the helm chart to support automatic Prometheus metric collection in the operator helm chart

dani-CO-CN:
**Issue number:**
Closes #558

**Description of changes:**
Add a named port for the controller-service and add an optional prometheus servicemonitor

**Terms of contribution:**


deploy/charts/bottlerocket-update-operator/values.yaml
```diff
@@ -111,3 +111,7 @@ logging:
     tracing_filter: "info"
   apiserver:
     tracing_filter: "info"
+
+prometheus:
+  serviceMonitor:
+    enabled: false
```
cbgbt: Do you mind fixing the missing newline at EOF?

deploy/charts/bottlerocket-update-operator/values.yaml
```diff
@@ -111,3 +111,7 @@ logging:
     tracing_filter: "info"
   apiserver:
     tracing_filter: "info"
+
+prometheus:
+  serviceMonitor:
```
cbgbt: It's probably a good idea to put this under `prometheus.controller`. This PR has made me realize that we expose metrics for `apiserver` but then don't bubble up the endpoint into our helm chart. I opened https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/564 to track that.

cbgbt (commented):
Thanks for this contribution!

Do you think it makes sense to place the `prometheus.io/scrape` annotations behind a configuration flag, defaulted to enabled? You don't have to implement this, I've just worked with more charts since we wrote this one and have noticed that in a few that I've used.

deploy/charts/bottlerocket-update-operator/values.yaml
dani-CO-CN: Added missing newline

deploy/charts/bottlerocket-update-operator/values.yaml
dani-CO-CN: added an additional hierarchy level `controller`

dani-CO-CN:
> Thanks for this contribution!
> 
> Do you think it makes sense to place the `prometheus.io/scrape` annotations behind a configuration flag, defaulted to enabled? You don't have to implement this, I've just worked with more charts since we wrote this one and have noticed that in a few that I've used.

We could also move the annotations to the values.yaml and then to a toYaml on the Service, this would allow overwriting.

```yaml
prometheus:
  controller:
    service:
      annotations: 
        prometheus.io/port: "8080"
        prometheus.io/scrape: "true"
``` 

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    {{- toYaml .Values.prometheus.controller.service.annotations | nindent 4 }}
  labels:
...
```

cbgbt (approved):
LGTM, thanks! 🚀 

> We could also move the annotations to the values.yaml and then to a toYaml on the Service, this would allow overwriting.

Good idea, I do like that better. I cut https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/566 to track that.

gthao313 (approved):

gthao313:
@dani-CO-CN LGTM, would you mind solving the conflicts? Otherwise, I can help on doing it. thanks!

dani-CO-CN:
@gthao313 conflicts are resolved! :rocket:

PR #560, opened 2023-10-07
feat: allow resource overwrites

### LINKED ISSUE #557: Support resource requests and limits in Helm chart
We would like to control the resource requests and limits for all brupop components using Helm values file. Please add support of it. Thank you.

dani-CO-CN:
**Issue number:**

Closes: #557

**Description of changes:**
Add a new key in values.yaml `resources`, move currently hard coded resources of agent-daemonset to the resources key and add new requests/limits for api-server and controller. I went with the resources for controller & apiserver with defaults from the ebs-csi-driver, feel free to remove the limits or adjust to some sane defaults.

**Testing done:**

**Terms of contribution:**

### DIFF
--- a/deploy/charts/bottlerocket-update-operator/templates/agent-daemonset.yaml
+++ b/deploy/charts/bottlerocket-update-operator/templates/agent-daemonset.yaml
@@ -62,12 +62,7 @@ spec:
           image: {{ .Values.image }}
           name: brupop
           resources:
-            limits:
-              cpu: 10m
-              memory: 50Mi
-            requests:
-              cpu: 5m
-              memory: 8Mi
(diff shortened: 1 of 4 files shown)


cbgbt (approved):
Thanks so much for contributing this! We based our agent limits on some research done in https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/240#issuecomment-1293969588.

While that research suggests we could tighten the memory constraints on the controller and apiserver, those components both keep the list of managed nodes in memory and thus have memory constraints that scale with the size of the cluster. I think the requests and limits here are high enough to fit that dataset comfortably, in any case.

gthao313 (approved):

gthao313:
The change looks great to us! Thank you @dani-CO-CN

Issue #486, opened 2023-07-13
brupop API server TLS cert is untrusted


jackgill:
As I mentioned in #478, the brupop API server on one of my EKS clusters apparently has an untrusted TLS cert:
```
> kubectl -n brupop-bottlerocket-aws get bottlerocketshadows
Error from server: conversion webhook for brupop.bottlerocket.aws/v1, Kind=BottlerocketShadow failed: Post "https://brupop-apiserver.brupop-bottlerocket-aws.svc:443/crdconvert?timeout=30s": x509: certificate signed by unknown authority
```
I installed brupop using the 1.1.0 manifest file and it is working fine on several other EKS clusters deployed using the same method.
**Image I'm using:**
1.1.0
**Issue or Feature Request:**
Looking at the PKI for brupop I see a self-signed issuer cert, but I'm not clear on how this cert is supposed to be trusted. Any advice on how to troubleshoot this issue would be appreciated.

jackgill:
I found some additional information: when describing the API server pods, I noticed this event:
`MountVolume.SetUp failed for volume "bottlerocket-tls-keys" : secret "brupop-apiserver-certificate" not found`
I did however see the secret so I suspect a race condition where the secret did not exist when the pod was first created, perhaps due to `cert-manager` running slow or something. What's confusing is that I recreated all the pods, the event did not recur, and yet I am still seeing the same error when trying to list `bottlerocketshadows`.

gthao313:
@jackgill Can you share with us how you install brupops? like installing cert-manager first and then install brupop? I'm trying to reproduce this issue.

Usually we need cert-manager running on the EKS cluster first and then install brupop after few minus when we confirm cert-manager is running.

gthao313:
@jackgill I just reproduce the same error `MountVolume.SetUp failed for volume "bottlerocket-tls-keys" : secret "brupop-apiserver-certificate" not found` by installing brupop immediately after running `cert-manager` installation. I think maybe `"https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=10s"` the cert-manager endpoint just not ready.

But this maybe not same as your situation.

jackgill:
@gthao313 cert-manager was installed via helm long before brupop was deployed on the cluster. Usually we use terraform to apply the brupop manifests but I've noticed that it creates the resources out of order, which I thought might be causing this problem. So I tried installing brupop with `kubectl apply -f` but I still saw this issue.

[...]

PR #595, opened 2024-02-01
Rename root-certificate to brupop-selfsigned-ca

alekhrycaiko:
**Issue number:**

https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/486

**Description of changes:**

This attempts to address #486. 

Changes root certificate referenced based on current findings of the issue.

**Testing done:**

Ran test suite.

Our own cluster already relies on `brupop-selfsigned-ca`. But I have not ran this specific PR against the cluster and could use an additional verification/eye there. 

**Terms of contribution:**

### DIFF
--- a/bottlerocket-update-operator.yaml
+++ b/bottlerocket-update-operator.yaml
@@ -196,6 +196,15 @@ spec:
     subresources:
       status: {}
 ---
+# Source: bottlerocket-update-operator/templates/controller-priority-class.yaml
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: brupop-controller-high-priority
+  namespace: brupop-bottlerocket-aws
+preemptionPolicy: Never
+value: 1000000
+---
 # Source: bottlerocket-update-operator/templates/agent-service-account.yaml
 apiVersion: v1
 kind: ServiceAccount
@@ -470,6 +479,7 @@ spec:
   ports:
     - port: 80
       targetPort: 8080
+      name: main
   selector:
     brupop.bottlerocket.aws/component: brupop-controller
 ---
@@ -653,6 +663,12 @@ spec:
               port: 8443
               scheme: HTTPS
             initialDelaySeconds: 5
+          resources:
+            limits:
+              memory: 256Mi
+            requests:
+              cpu: 10m
+              memory: 40Mi
           volumeMounts:
             - mountPath: /etc/brupop-tls-keys
               name: bottlerocket-tls-keys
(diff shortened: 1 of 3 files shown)


yeazelm:
Thanks @alekhrycaiko for cutting this PR! I'm just getting up to speed on this issue so I'll need to look at this a bit more to confirm it will work. I'll dig in and get back to you! Sorry for the delay in response!

bottlerocket-update-operator.yaml
```diff
@@ -196,6 +196,15 @@ spec:
...
 ---
+# Source: bottlerocket-update-operator/templates/controller-priority-class.yaml
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: brupop-controller-high-priority
+  namespace: brupop-bottlerocket-aws
+preemptionPolicy: Never
+value: 1000000
+---
```
yeazelm, line 207: When running `make manifest` this ends up at the end rather than here. We might just move this to the end of the file to avoid it creating a dirty tree when running it, functionally it works the same.

deploy/tests/snapshots/insta_tests__generated_crds.snap.new
```diff
@@ -0,0 +1,193 @@
+---
```
yeazelm, line 1: Was this file autogenerated? I'm not sure where it came from?

models/src/constants.rs
```diff
@@ -39,7 +39,7 @@ pub const PUBLIC_KEY_NAME: &str = "tls.crt";
 pub const PRIVATE_KEY_NAME: &str = "tls.key";
 pub const TLS_KEY_MOUNT_PATH: &str = "/etc/brupop-tls-keys";
 // Certificate object name
-pub const ROOT_CERTIFICATE_NAME: &str = "root-certificate";
+pub const ROOT_CERTIFICATE_NAME: &str = "brupop-selfsigned-ca";
```
yeazelm, line 42: I think this is probably safe but I'm digging into how we use cert-manager under the hood to ensure we aren't going to break something else down the line. Do you happen to have links to what led you to this change?

alekhrycaiko: I do not unfortunately. I can just attest from what Jack mentions [here](https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/486#issuecomment-1894696759) that this resolved an issue on our end.

v0lumehi: Since it might help you: I just fixed this bug and set the value to cert-manager.io/inject-ca-from: brupop-bottlerocket-aws/brupop-apiserver-certificate in the CRD

PR #729, opened 2025-02-25
Publish helm charts for v1.5.0

ytsssun:
This publishes charts for the Bottlerocket Update Operator version v1.5.0 to the helm repository.


bottlerocket-shadow-1.0.0.tgz
ytsssun, line 1: This change is not expected, updating the PR

PR #730, opened 2025-02-25
Publish helm charts for v1.5.0

ytsssun:
This publishes charts for the Bottlerocket Update Operator version v1.5.0 to the helm repository.


vigh-m (approved):

cbgbt (approved):

ytsssun:
FWIW, we noticed that the create time is updated for each entries in the index.yaml. 

This is a known issue in helm - https://github.com/helm/helm/issues/7363#issuecomment-572639519

For now we will take it as it is. EKS is using the exact same script https://github.com/aws/eks-charts/blob/master/scripts/publish-charts.sh and they are seeing the same behavior. 

We can improve it by introducing some workaround like https://github.com/kubernetes-sigs/cloud-provider-azure/pull/1640/files#diff-a83721951d109a6c640d5106465c50392bdb5723cf8db075a0f03a66e02e5831R22-R26.

Issue #774, opened 2025-08-05
Index out of bounds in brupop-agent in agent/src/apiclient.rs:237:89


Broders:
**Image I'm using: 1.6.0**
**Bottlerocket OS I'm using: bottlerocket-aws-k8s-1.32-x86_64-v1.39.0-0968c061**

**Issue:**

Hello, the brupop-agent crashes with a panic in agent/src/apiclient.rs:237:89 due to an index out of bounds: the len is 1 but the index is 1 error. 

> thread 'main' panicked at agent/src/apiclient.rs:237:89:
index out of bounds: the len is 1 but the index is 1

I couldn´t figure out why this error occured, it worked for weeks until now. Bottlerocket is new to me.

if I enable DEBUG via values.yml I get these before the container dies:

> 2025-08-07T08:43:40.282815Z DEBUG agent::apiclient::api: Invoking apiclient: ["raw", "-u", "/os"]
at agent/src/apiclient.rs:267
in agent::apiclient::api::invoke_apiclient with args: ["raw", "-u", "/os"]
in agent::apiclient::api::get_os_info
in agent::agentclient::shadow_status_with_refreshed_system_matadata with shadow_error_info: ShadowErrorInfo { crash_count: 0, state_transition_failure_timestamp: None }
in agent::agentclient::initialize_metadata_shadow
in agent::agentclient::initialize_shadow_if_not_initialized
in agent::agentclient::run
2025-08-07T08:43:40.284426Z DEBUG hyper::proto::h1::io: parsed 7 headers
at /src/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hyper-0.14.32/src/proto/h1/io.rs:211
2025-08-07T08:43:40.284439Z DEBUG hyper::proto::h1::conn: incoming body is chunked encoding
at /src/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hyper-0.14.32/src/proto/h1/conn.rs:224
thread 'main' panicked at agent/src/apiclient.rs:237:89:
index out of bounds: the len is 1 but the index is 1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

I aktivated rust_backtrace
> stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic_bounds_check
   3: agent::apiclient::api::extract_status_code_from_error
[...]

PR #781, opened 2025-09-02
fix: improve status code extraction from API client error messages

maherthomsi:
**Issue number:**
#774 

**Description of changes:**
Fix a bug where an index out-of-bounds error would be returned when "Status" is not in  the apiclient response error message.  The previous string-based approach using "Status" as a delimiter could fail if the delimiter was not present or formatted differently.  

This change replaces string manipulation with proper HTTP status code parsing using the http crate's StatusCode type. The extraction now safely handles missing delimiters and invalid status codes by returning the StatusCode instead of raw string slices.

**Testing done:**
Ran ```make build```

Validated status code parsing function locally.

Created unit tests for status code parsing function

**Terms of contribution:**

### DIFF
--- a/agent/src/apiclient.rs
+++ b/agent/src/apiclient.rs
@@ -115,7 +115,7 @@ pub(super) mod api {
     use nonzero_ext::nonzero;
     use semver::Version;
     use serde::Deserialize;
-    use snafu::ResultExt;
+    use snafu::{OptionExt, ResultExt};
     use std::process::{Command, Output};
     use tokio::time::Duration;
     use tokio_retry::{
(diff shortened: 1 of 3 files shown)


mgsharm (approved):

agent/src/apiclient.rs
```diff
@@ -232,12 +233,15 @@ pub(super) mod api {
     /// Extract error statuscode from stderr string
     /// Error Example:
     /// "Failed POST request to '/actions/refresh-updates': Status 423 when POSTing /actions/refresh-updates: Update lock held\n"
-    fn extract_status_code_from_error(error: &str) -> &str {
-        let error_content_split_by_status: Vec<&str> = error.split("Status").collect();
-        let error_content_split_by_whitespace: Vec<&str> = error_content_split_by_status[1]
+    fn extract_status_code_from_error(error: &str) -> Option<StatusCode> {
```
ytsssun: I wonder if there is a more direct way of getting that status code such as directly use the unix socket and invoke the APIServer using an http client. 

Having to parse the status code from the response feels a bit odd to me..

agent/src/apiclient.rs
```diff
@@ -232,12 +233,15 @@ pub(super) mod api {
...
+        error
+            .split("Status ")
+            .nth(1)?
```
ytsssun: We should create a error enum variant like StatusCodeParsingError in https://github.com/bottlerocket-os/bottlerocket-update-operator/blob/c0d2f62a781b5a60e2af8c497ba9eed46cb09e3e/agent/src/agentclient.rs#L679 and propagate that error with .context(). You can refer to examples within this file. Same thing for next()? below. We want to be intentional about the error we throw when bail early.

ytsssun (commented):
Thanks for this fix, can we try to add unit tests for the index out of bound issue? 

Also were you able to reproduce the original issue? If yes, it would also be a good idea to run the "integration test" to confirm the fix.

agent/src/apiclient.rs
maherthomsi: We shouldn't directly use the unix socket.  The intended architectural approach is to utilize the existing apiclient binary rather than creating a custom HTTP client to communicate directly with the Bottlerocket API.  If we do want this functionality, the change would take place in the apiclient rather than within Brupop, and would be tied to the newer Bottlerocket releases.  Since the API client would return the response rather than just the code, we would need to parse the response for the code.

agent/src/apiclient.rs
maherthomsi: Agreed.  I'll add this change in my next commit.

maherthomsi:
I was not able to reproduce the original issue since there was a custom build involved, but I was able to include unit tests to mock the scenario experienced in the original issue to validate that the fix works.

ytsssun (approved):

maherthomsi:
Commit message changes

yeazelm (approved):

PR #802, opened 2026-02-17
Update Rust dependencies

jpculp:
**Description of changes:**

Updated various Rust dependencies including:

- `opentelemetry` from **0.29** to **0.31**.
- `kube` from **0.88** to **2**.
- `tokio` from **1.46** to **1.47 LTS**.
- `bottlerocket-sdk` from **0.65.0** to **0.70.0**.

Also removed unused `webpki`.

**Testing done:**

- [x] Updated k8s 1.29 cluster. _(oldest supported)_
- [x] Updated k8s 1.35 cluster. _(latest)_

**Terms of contribution:**

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,14 +51,14 @@ opentelemetry-prometheus = "0.31"
 prometheus = "0.14"
 
 # k8s-openapi must match the version required by kube and enable a k8s version feature
-k8s-openapi = { version = "0.21", default-features = false, features = ["v1_24"] }
-kube = { version = "0.88", default-features = false, features = [ "derive", "runtime", "rustls-tls" ] }
+k8s-openapi = { version = "0.26", default-features = false, features = ["v1_30"] }
+kube = { version = "2", default-features = false, features = [ "derive", "runtime", "rustls-tls" ] }
 
 regex = "1"
 reqwest = { version = "0.12", default-features = false, features =  [ "json", "rustls-tls" ] }
 rustls = { version = "0.23", default-features = false, features = ["aws_lc_rs", "logging", "std", "tls12"] }
 rustls-pemfile = { version = "2" }
-schemars = "0.8"
+schemars = "1"
 semver = "1"
 serde = "1"
 serde_json = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,12 +42,12 @@ nonzero_ext = "0.3"
 # these package versions should be moved in lockstep to match upstream
 actix-web = { version = "4.9", features = ["rustls-0_23"] }
 tracing-actix-web = "0.7"
-opentelemetry-instrumentation-actix-web = { version = "0.22", features = ["metrics"] }
+opentelemetry-instrumentation-actix-web = { version = "0.23", features = ["metrics"] }
 
 # opentelemetry dependencies
-opentelemetry = { version = "0.29", features = ["metrics"]}
-opentelemetry_sdk = { version = "0.29", features = ["rt-tokio-current-thread", "metrics"]}
-opentelemetry-prometheus = "0.29"
+opentelemetry = { version = "0.31", features = ["metrics"]}
+opentelemetry_sdk = { version = "0.31", features = ["rt-tokio-current-thread", "metrics"]}
+opentelemetry-prometheus = "0.31"
 prometheus = "0.14"
 
 # k8s-openapi must match the version required by kube and enable a k8s version feature
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,4 +72,3 @@ tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["registry", "env-filter", "json"] }
 uuid = { version = "1", default-features = false, features = ["serde", "v4"] }
 validator = { version = "0.20", features = ["derive"] }
-webpki = { version = "0.22", features = ["std"] }
--- a/apiserver/Cargo.toml
+++ b/apiserver/Cargo.toml
@@ -21,7 +21,6 @@ awc = { workspace = true }
 opentelemetry-instrumentation-actix-web = { workspace = true }
 rustls = { workspace = true }
 rustls-pemfile = { workspace = true }
-webpki = { workspace = true }
 opentelemetry = { workspace = true }
 opentelemetry_sdk = { workspace = true }
 opentelemetry-prometheus = { workspace = true }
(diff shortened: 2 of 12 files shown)


Cargo.toml
```diff
@@ -42,23 +42,23 @@ nonzero_ext = "0.3"
...
-opentelemetry-prometheus = "0.29"
+opentelemetry = { version = "0.31", features = ["metrics"]}
+opentelemetry_sdk = { version = "0.31", features = ["rt-tokio-current-thread", "metrics"]}
+opentelemetry-prometheus = "0.31"
 prometheus = "0.14"
 
 # k8s-openapi must match the version required by kube and enable a k8s version feature
-k8s-openapi = { version = "0.21", default-features = false, features = ["v1_24"] }
-kube = { version = "0.88", default-features = false, features = [ "derive", "runtime", "rustls-tls" ] }
+k8s-openapi = { version = "0.26", default-features = false, features = ["v1_30"] }
```
jpculp, line 54: Although this is set to `v1_30` (the oldest supported by `k8s-openapi@0.26`) I didn't have any issues updating a k8s 1.29 cluster.

jpculp:
Addressed clippy warning.

jpculp:
Updated CRDs.

jpculp:
Bump `actix-web`.

Cargo.toml
```diff
@@ -42,12 +42,12 @@ nonzero_ext = "0.3"
 # these package versions should be moved in lockstep to match upstream
 actix-web = { version = "4.9", features = ["rustls-0_23"] }
 tracing-actix-web = "0.7"
-opentelemetry-instrumentation-actix-web = { version = "0.22", features = ["metrics"] }
+opentelemetry-instrumentation-actix-web = { version = "0.23", features = ["metrics"] }
```
cbgbt, line 46: May be worth querying the metric endpoint to see that they are properly exposed.

apiserver/src/auth/authorizor.rs
```diff
@@ -275,7 +275,11 @@ pub(crate) mod test {
     ) -> K8STokenAuthorizor<MockTokenReviewer> {
         let mut pod_store = reflector::store::Writer::<Pod>::default();
         let pod_reader = pod_store.as_reader();
-        pod_store.apply_watcher_event(&Event::Restarted(pods));
+        pod_store.apply_watcher_event(&Event::Init);
+        for pod in pods {
+            pod_store.apply_watcher_event(&Event::InitApply(pod));
+        }
+        pod_store.apply_watcher_event(&Event::InitDone);
```
cbgbt, line 282: Do you know what the context on this change is?

Cargo.toml
cbgbt: Kubernetes has a +/- 1 support story for the API, so 1.30 should work for 1.29.

We use very stable elements of the k8s API - theoretically very new versions of k8s could make changes that we don't support, but it is unlikely.

apiserver/Cargo.toml
```diff
@@ -21,7 +21,6 @@ awc = { workspace = true }
 opentelemetry-instrumentation-actix-web = { workspace = true }
 rustls = { workspace = true }
 rustls-pemfile = { workspace = true }
-webpki = { workspace = true }
```
cbgbt, line 24: What prompts this removal? Is it just marked as unused?

jpculp: Just unused. We could leave it but it adds an unnecessary dependency and we have to keep `clarify.toml` and `deny.toml` up to date to accommodate it.

apiserver/src/auth/authorizor.rs
jpculp: There was an API change in the `kube-runtime` reflector store. Basically `Event::Restarted` was broken up into three phases.

Cargo.toml
jpculp: It looks like prometheus metrics are being properly exposed.

apiserver/src/main.rs
```diff
@@ -153,7 +153,7 @@ pub mod apiserver_error {
 
         #[snafu(display("Error creating prometheus registry: '{}'", source))]
         PrometheusRegsitry {
```
ytsssun, line 155: Nit: we have some pre-existing typo...
```suggestion
        PrometheusRegistry {
```

controller/src/main.rs
```diff
@@ -190,7 +190,7 @@ pub mod controller_error {
 
         #[snafu(display("Error creating prometheus registry: '{}'", source))]
         PrometheusRegsitry {
```
ytsssun, line 192: Nit: we have pre-existing typo
```suggestion
        PrometheusRegistry {
```

cbgbt (approved):

Cargo.toml
```diff
@@ -42,23 +42,23 @@ nonzero_ext = "0.3"
...
+opentelemetry = { version = "0.31", features = ["metrics"]}
+opentelemetry_sdk = { version = "0.31", features = ["rt-tokio-current-thread", "metrics"]}
+opentelemetry-prometheus = "0.31"
 prometheus = "0.14"
 
 # k8s-openapi must match the version required by kube and enable a k8s version feature
-k8s-openapi = { version = "0.21", default-features = false, features = ["v1_24"] }
-kube = { version = "0.88", default-features = false, features = [ "derive", "runtime", "rustls-tls" ] }
+k8s-openapi = { version = "0.26", default-features = false, features = ["v1_30"] }
+kube = { version = "2", default-features = false, features = [ "derive", "runtime", "rustls-tls" ] }
```
ytsssun, line 55: `kube` crate has a 3.0 now. Any concern if we further bump to that?

apiserver/src/main.rs
jpculp: Since this is an already big change I'd like to save this for a separate PR.

Cargo.toml
jpculp: We want to maintain compatibility with `v1_30`. Jumping to `kube@3.0` only guarantees API compatibility with a minimum of `v1_31`.

Cargo.toml
ytsssun: The https://kubernetes.io/releases/version-skew-policy/ says slightly different - the controller/client should not be newer than the api-server. So technically there is no guaranteed compatibility between controllers under version `1.30` and api-server under `1.29`. 

Is it possible to pin it to 1_29 until we drop the 1.29 variants?

jpculp: I don't think that's necessary since we've confirmed the ability to update on a 1.29 cluster. Extended support does drop soon, but I don't think we should hold these updates back when we've validated the operator functionality.

ytsssun (approved):

mgsharm (approved):

